PR #46, opened 2024-09-09
Feat ratelimiter

matoszz:
Adds a new generic rate limiter based on cloudflare's design

### DIFF
--- a/pkg/middleware/ratelimiter/examples/http_middleware.go/http_middleware.go
+++ b/pkg/middleware/ratelimiter/examples/http_middleware.go/http_middleware.go
@@ -1,0 +1,90 @@
+package main
+
+import (
+	"fmt"
+	"html"
+	"log"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/theopenlane/core/pkg/middleware/ratelimiter"
+)
+
+// GetRemoteIP returns the remote IP address of the request
+func GetRemoteIP(ipLookups []string, forwardedForIndexFromBehind int, r *http.Request) string {
+	realIP := r.Header.Get("X-Real-IP")
+	forwardedFor := r.Header.Get("X-Forwarded-For")
+
+	for _, lookup := range ipLookups {
+		if lookup == "RemoteAddr" {
+			// 1. Cover the basic use cases for both ipv4 and ipv6
+			ip, _, err := net.SplitHostPort(r.RemoteAddr)
+			if err != nil {
+				// 2. Upon error, just return the remote addr.
+				return r.RemoteAddr
+			}
+			return ip
+		}
+		if lookup == "X-Forwarded-For" && forwardedFor != "" {
+			// X-Forwarded-For is potentially a list of addresses separated with ","
+			parts := strings.Split(forwardedFor, ",")
+			for i, p := range parts {
+				parts[i] = strings.TrimSpace(p)
+			}
+
+			partIndex := len(parts) - 1 - forwardedForIndexFromBehind
+			if partIndex < 0 {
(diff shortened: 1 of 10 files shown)


pkg/middleware/ratelimiter/examples/http_middleware.go/http_middleware.go
```diff
@@ -0,0 +1,90 @@
+package main
```
golanglemonade, line 1: Can you added `pkg/middleware/ratelimiter/examples/http_middleware.go/` to the sonar exclude list so it's not included in the checks?

pkg/middleware/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// inspired by Cloudflare's approach: https://blog.cloudflare.com/counting-things-a-lot-of-different-things/
+
+// LimitStore is the interface that represents limiter internal data store
+type LimitStore interface {
+	// Inc increments current window limit counter for key
+	Inc(key string, window time.Time) error
+	// Get gets value of previous window counter and current window counter for key
+	Get(key string, previousWindow, currentWindow time.Time) (prevValue int64, currValue int64, err error)
```
golanglemonade, line 14: I see `Size` in the in memory store, should that  (`Size() int`) be included in the interface?

pkg/middleware/ratelimiter/map_limit_store.go
```diff
@@ -0,0 +1,87 @@
...
+	val        int64
+	lastUpdate time.Time
+}
+
+// NewMapLimitStore creates new in-memory data store for internal limiter data
+func NewMapLimitStore(expirationTime time.Duration, flushInterval time.Duration) (m *MapLimitStore) {
+	m = &MapLimitStore{
+		data:           make(map[string]limitValue),
+		expirationTime: expirationTime,
+	}
```
golanglemonade, line 30: This should return the interface type, that will ensure the type fully implements the interface:
```suggestion
// NewMapLimitStore creates new in-memory data store for internal limiter data
func NewMapLimitStore(expirationTime time.Duration, flushInterval time.Duration) LimitStore {
	m := &MapLimitStore{
		data:           make(map[string]limitValue),
		expirationTime: expirationTime,
	}
```

pkg/middleware/ratelimiter/map_limit_store.go
```diff
@@ -0,0 +1,87 @@
...
+// Get gets value of previous window counter and current window counter
+func (m *MapLimitStore) Get(key string, previousWindow, currentWindow time.Time) (prevValue int64, currValue int64, err error) {
+	m.mutex.RLock()
+
+	defer m.mutex.RUnlock()
+
+	prevValue = m.data[mapKey(key, previousWindow)].val
+	currValue = m.data[mapKey(key, currentWindow)].val
+
+	return prevValue, currValue, nil
```
golanglemonade: nit: you defined these returns in the function signature, you can just do this instead:
```suggestion
	return 
```

pkg/middleware/ratelimiter/ratelimiter.go
matoszz: `size` is a function that returns `return len(m.data)` so it shouldn't be a part of the interface

pkg/middleware/ratelimiter/map_limit_store.go
matoszz: `MapLimitStore` is an option in-memory implementation where others could be used; returning LimitStore doesn't make sense in this instance

golanglemonade (approved):

PR #49, opened 2024-09-09
add apple merchant file association

matoszz:
required to host this file to accept apple pay


internal/httpserve/route/static.go
```diff
@@ -124,3 +124,27 @@ func registerFaviconHandler(router *Router) (err error) {
 
 	return nil
 }
+
+//go:embed applemerchant
+var applemerchant embed.FS
+
+// registerAppleMerchant serves up the text output of the applemerchant file
+func registerAppleMerchant(router *Router) (err error) {
```
golanglemonade: This needs to be added the [routeHandlers](https://github.com/theopenlane/core/blob/main/internal/httpserve/route/router.go#L116)

PR #121, opened 2024-10-07
Feat objectupload

matoszz:


internal/httpserve/route/upload.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"net/http"
+
+	echo "github.com/theopenlane/echox"
+)
+
+var (
+	// uploadKeys are the keys that can be used to upload files in a multipart form
+	uploadKeys = []string{"uploadFile"}
```
matoszz: @golanglemonade shouldn't this use the config parameter vs. a static var?

pkg/models/models.go
```diff
@@ -714,3 +714,38 @@ var ExampleAccountRolesOrganizationReply = AccountRolesOrganizationReply{
 	Roles:          []string{"can_view", "can_edit", "audit_log_viewer"},
 	OrganizationID: "01J4HMNDSZCCQBTY93BF9CBF5D",
 }
+
+// =========
+// FILES
+// =========
+
+type UploadFilesRequest struct {
+	UploadFile []byte `form:"uploadFile"`
```
matoszz: should this be `byte` or `multipart.File`

internal/httpserve/route/upload.go
golanglemonade: ah yep, missed this when I switched it to the config; this isn't actually used

pkg/objects/context.go
```diff
@@ -0,0 +1,70 @@
+package objects
+
+import (
+	"context"
+)
+
+// FileContextKey is the context key for the files
+var FileContextKey = &ContextKey{"files"}
```
matoszz, line 11: may want to put a comment in here (or the readme) indicating the difference between the input key provided by the user and the context key we use

pkg/objects/binder.go
```diff
@@ -0,0 +1,96 @@
...
+	echo "github.com/theopenlane/echox"
+)
+
+var (
+	typeMultipartFileHeader      = reflect.TypeOf((*multipart.FileHeader)(nil))
+	typeMultipartSliceFileHeader = reflect.TypeOf(([]*multipart.FileHeader)(nil))
+)
+
+// BindFunc is a function that binds the request body to the struct pointer
+type BindFunc func(echo.Context, interface{}) error
```
matoszz: @golanglemonade did you end up using this binder anywhere?

golanglemonade: No, want me to pull it out @matoszz ?

pkg/models/models.go
golanglemonade: I think `multipart.FileHeader`, updating

pkg/objects/binder.go
golanglemonade: removing

PR #145, opened 2024-10-21
feat-object storage enhancements

matoszz:
- Updates the NewS3FromConfig function to be more robust and derive settings entirely from `Opts` vs. upstream AWS config parameters
- checks that the configured default bucket is in the slice of buckets returned available to the configured account
- adds functional option parameters to the Disk storage type
- Refactors the manager upload function in s3 storage to actually use the uploader and downloader manager
- refactors the downloader function to actually work and return the bytes of the file as well as the io.ReaderAt so the caller can decide what to do with the file


pkg/objects/io.go
```diff
@@ -101,25 +95,24 @@ func ReaderToSeeker(r io.Reader) (io.ReadSeeker, error) {
...
-	// #nosec: G501
-	hash := md5.New() //nolint:gosec  // MD5 is used for checksums, not for hashing passwords
+	hash := sha256.New()
 	if _, err := io.Copy(hash, data); err != nil {
 		return "", fmt.Errorf("could not read file: %w", err)
 	}
 
 	if _, err := data.Seek(0, io.SeekStart); err != nil { // seek back to beginning of file
-		return "", fmt.Errorf("%s: %w", seekError, err)
+		return "", fmt.Errorf("%s: %w", ErrSeekError, err)
```
golanglemonade: `%s` is a string, you either don't need to define the error or you should do `%w: %w`

pkg/objects/io.go
```diff
@@ -101,25 +95,24 @@ func ReaderToSeeker(r io.Reader) (io.ReadSeeker, error) {
...
 	}
 
-	return base64.StdEncoding.EncodeToString(hash.Sum(nil)), nil
+	return fmt.Sprintf("%x", hash.Sum(nil)), nil
```
golanglemonade: The tests for this function are failing due to the md5 -> sha256 change and need to be updated.

pkg/objects/storage/disk.go
```diff
@@ -62,12 +68,20 @@ func (d *Disk) ManagerUpload(ctx context.Context, files [][]byte) error {
 
 // Download is used to download a file from the storage backend
 // TODO: Implement this method
```
golanglemonade:
```suggestion
```

pkg/objects/storage/s3.go
```diff
@@ -94,16 +92,32 @@ func (s *S3Store) Exists(ctx context.Context, key string) (bool, error) {
 	return true, nil
 }
 
+// Exists checks if an object exists in S3
```
golanglemonade:
```suggestion
// HeadObj checks if an object exists in S3
```

golanglemonade (approved):

PR #215, opened 2024-11-16
feat: finish programs, policy & procedures permissions, add tests

golanglemonade:
- New FGA model with added structure of permissions including `editor`, `viewer`, and `blocked` permissions based on a group. I have a few more to add (e.g. more configurability into what permissions a org member can be given, right now they cannot create policy/procedures but if they are assigned to a group they can edit existing ones. We might want to also allow a group to be assigned to give create access for these objects). The auditor role also needs to be fleshed out. 
- Policies and procedures have their access entirely controlled by groups. By default - all users in the organization will have view access, edit access is granted via a group; but you can also remove view access by blocking a groupo
- Programs can have access assigned via a membership to the program (e.g. the same as concept as org-memberships) but also specific edit/view access can be granted via groups instead of membership. This should allow flexibility in the permissions. 
- Removes the subscription/feature from the fga model, this isn't used and I need to go do this fully anyways. 
- Adds the CLI for programs, policies, and procedures
- Adds tests for programs, policies, and procedures
- Adds an _example_test.go [template](https://github.com/theopenlane/core/pull/215/files#diff-68a5eb2a0bc4b317276fd6a7da841d7a90a4c55aba57aca4bfe5df81bafb19e1) that can be used as a starting point for adding CRUD tests for all graph resolvers
- fixes the FK error to return the edge object vs the main object

Example request: 

```graphql
mutation CreateProcedure($input: CreateProcedureInput!) {
  createProcedure(input: $input) {
    procedure {
      background
      id
      name
      editors {
        id
        name
      }
    }
  }
}
```

```json
{
  "input": {
    "name": "meow meow"
    "editorIDs": ["01JD85KTFAEK6J31GJZQHXDEEN"]
  }
}
```

result:
```json
{
  "data": {
    "createProcedure": {
      "procedure": {
        "background": "",
        "id": "01JD85VMT5A8SEVWQ5QGRWGDFS",
        "name": "meow meow",
        "editors": [
          {
            "id": "01JD85KTFAEK6J31GJZQHXDEEN",
            "name": "editor"
          }
        ]
      }
    }
  },
  "extensions": {
    "auth": {
      "authentication_type": "api_token",
      "authorized_organization": "01JD85KM7AVH8G409X424EVZ77"
    },
    "server_latency": "35.849834ms",
    "trace_id": "ufzfqChAErjsOjfOxyNBCQkXkkpcWSyp"
  }
}
```


matoszz (approved):
nice work :)

PR #221, opened 2024-11-18
Document adding openlane.fga

itsmed:
`task run-dev` failed with

```
Error: open fga/model/openlane.fga: no such file or directory
exit status 1
task: Failed to run task "run-dev": exit status 1
```

I fixed this by running `cp config/config-dev.example.yaml .config/.config.yaml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,12 @@ See the [README](/config/README.md) in the `config` directory.
 1. Update the configuration with whatever respective settings you desire; the
    defaults inside should allow you to run the server without a problem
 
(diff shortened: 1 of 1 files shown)


matoszz:
thanks for updating this @itsmed ! Appreciate the contribution. If you're interested in also changing where the default is generated from, you could update https://github.com/theopenlane/iam/blob/main/fgax/fga.go#L35C126-L35C134 to have the path be `fga/model/model.fga` or vice versa.

matoszz (approved):

PR #224, opened 2024-11-19
Feat add initial entitlements

matoszz:
- adds an initial (and incomplete) entitlements package with helper functions and client 
- adds configuration parameters for Stripe and respective settings
- adds stripe client to serverOpts and Handlers
- creates an event pool wrapper and stripe client addition to the db client initialization alongside global mutation hooks for event pool submission
- adds a checkout handler which today populates a field on the front end so that if a customer has an active subscription they cannot use the checkout link without logging in


config/config.go
```diff
@@ -68,6 +69,8 @@ type Config struct {
 	Ratelimit ratelimit.Config `json:"ratelimit" koanf:"ratelimit"`
 	// ObjectStorage contains the configuration for the object storage backend
 	ObjectStorage objects.Config `json:"objectStorage" koanf:"objectStorage"`
+	// Subscription contains the configuration for the entitlements service
```
golanglemonade:
```suggestion
	// Entitlements contains the configuration for the entitlements service
```

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+
+// RegisterGlobalHooks registers global hooks for the ent client with the initialized event pool
+func RegisterGlobalHooks(client *entgen.Client, pool *soiree.EventPool) {
+	client.Use(EmitEventHook(pool))
+}
+
+// InitEventPool initializes an event pool with a client and a error handler
+func InitEventPool(client interface{}) *soiree.EventPool {
+	return soiree.NewEventPool(soiree.WithPool(soiree.NewPondPool(soiree.WithMaxWorkers(100), soiree.WithName("ent_event_pool"))), soiree.WithErrorHandler(func(event soiree.Event, err error) error { // nolint: mnd
+		log.Printf("Error encountered during event '%s': %v, with payload: %v", event.Topic(), err, event.Payload())
```
golanglemonade, line 83: This should probably be an error log vs. `Printf` which is a debug log?

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+		if err != nil {
+			log.Err(err).Msg("Failed to retrieve Stripe customer by ID")
+			return err
+		}
+
+		billingEmail, exists := props["billing_email"]
+		if exists && billingEmail != "" {
+			email := billingEmail.(string)
+			if stripecustomer.Email != email {
+				_, err := event.Client().(*entgen.Client).EntitlementManager.Client.Customers.Update(stripes, &stripe.CustomerParams{
```
golanglemonade: this `Update` call is happening three times in this function, it would be more efficient to do this in a single update vs. individually per field

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+					Address: &stripe.AddressParams{
+						Line1: &address,
+					},
+				})
+				if err != nil {
+					log.Err(err).Msg("failed to update Stripe customer address")
+					return err
+				}
+			}
+		}
```
golanglemonade, line 140: I think you need a return here, otherwise I think it will attempt to create another record if the `billing_email` is in the props?

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+	for _, event := range []string{"OrganizationSetting.OpUpdate", "OrganizationSetting.OpUpdateOne"} {
+		_, err = pool.On(event, handleCustomerCreate)
+		if err != nil {
+			log.Error().Err(ErrFailedToRegisterListener)
+		}
+	}
+}
+
+// handleCustomerCreate handles the creation of a customer in Stripe
+func handleCustomerCreate(event soiree.Event) error {
```
golanglemonade, line 105: this function is pretty complex, I would suggest splitting out the "update stripe record" vs. "create new stripe record" into their own functions that are called from here

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+
+			log.Info().Msgf("Stripe subscription with ID exists %s", subs.ID)
+
+			checkout, err := event.Client().(*entgen.Client).EntitlementManager.CreateBillingPortalUpdateSession(subs.ID, customer.ID)
+			if err != nil {
+				log.Err(err).Msg("failed to create billing portal update session")
+				return err
+			}
+
+			log.Warn().Msgf("Created billing portal update session with URL %s", checkout.URL)
```
golanglemonade: why is this a warn?

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+
+			log.Info().Msgf("Created Stripe customer with ID: %s", customer.ID)
+
+			if err := updateOrganizationSettingWithCustomerID(event.Context(), orgsettingID.(string), customer.ID, event.Client()); err != nil {
+				log.Err(err).Msg("Failed to update OrganizationSetting with Stripe customer ID")
+				return err
+			}
+
+			log.Info().Msgf("Updated OrganizationSetting with Stripe customer ID: %s", customer.ID)
```
golanglemonade: Most of these `Info` levels should probably be `Debug` because they shouldn't be needed in most cases because they are success

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+
+		log.Info().Msgf("Stripe subscription with ID exists %s", subs.ID)
+
+		checkout, err := event.Client().(*entgen.Client).EntitlementManager.CreateBillingPortalUpdateSession(subs.ID, i.Customer().ID)
+		if err != nil {
+			log.Err(err).Msg("failed to create billing portal update session")
+			return err
+		}
+
+		log.Warn().Msgf("Created billing portal update session with URL %s", checkout.URL)
```
golanglemonade: Debug? not warn?

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+
+		log.Info().Msgf("Updated OrganizationSetting with Stripe customer ID: %s", i.Customer().ID)
+	}
+
+	return nil
+}
+
+// updateOrganizationSettingWithCustomerID updates an OrganizationSetting with a Stripe customer ID
+func updateOrganizationSettingWithCustomerID(ctx context.Context, orgID, customerID string, client interface{}) error {
+	if _, err := client.(*entgen.Client).OrganizationSetting.UpdateOneID(orgID).SetStripeID(customerID).Save(ctx); err != nil {
```
golanglemonade: `orgID` is wrong to update the organizationSetting by ID, you either need the Org Setting ID, or do a `Update()...Where()...`

internal/ent/hooks/organization.go
```diff
@@ -307,6 +307,35 @@ func checkAndUpdateDefaultOrg(ctx context.Context, userID string, oldOrgID strin
...
+		if err != nil {
+			return "", err
+		}
+
+		return organizationSetting.ID, nil
+	}
+
 	input := generated.CreateOrganizationSettingInput{}
 
 	organizationSetting, err := m.Client().OrganizationSetting.Create().SetInput(input).Save(ctx)
```
golanglemonade: you can simply this and do the call to `Create` once. This will also allow it to be more easily updated for other defaults later 

```suggestion
	input := generated.CreateOrganizationSettingInput{}
	
	personalOrg, _ := m.PersonalOrg()
	if !personalOrg {
		userID, err := auth.GetUserIDFromContext(ctx)
		if err != nil {
			log.Error().Err(err).Msg("unable to get user id from context")

			return "", err
		}

		user, err := m.Client().User.Get(ctx, userID)
		if err != nil {
			log.Error().Err(err).Msg("unable to fetch user from database")
		}

		billingContact := user.FirstName + " " + user.LastName

		input.BillingEmail = &user.Email,
		input.BillingContact = &billingContact
	}

	organizationSetting, err := m.Client().OrganizationSetting.Create().SetInput(input).Save(ctx)
```

internal/httpserve/handlers/checkout.go
```diff
@@ -0,0 +1,76 @@
...
+	reqCtx := ctx.Request().Context()
+
+	orgID, err := auth.GetOrganizationIDFromContext(reqCtx)
+	if err != nil {
+		log.Error().Err(err).Msg("unable to get organization id from context")
+
+		return h.BadRequest(ctx, err)
+	}
+
+	log.Warn().Msg("obtained organization ID from context")
```
golanglemonade:
```suggestion
	log.Debug().Msg("obtained organization ID from context")
```

internal/httpserve/handlers/checkout.go
```diff
@@ -0,0 +1,76 @@
...
+
+	settings, err := h.getOrgSettingByOrgID(reqCtx, orgID)
+	if err != nil {
+		log.Error().Err(err).Msg("unable to get organization settings by org id")
+
+		return h.BadRequest(ctx, err)
+	}
+
+	log.Warn().Msg("obtained organization settings by org ID")
```
golanglemonade: Same on on these `Warns` below...
```suggestion
	log.Debug().Msg("obtained organization settings by org ID")
```

internal/httpserve/handlers/checkout.go
```diff
@@ -0,0 +1,76 @@
...
+
+	result, err := h.Entitlements.Client.CustomerSessions.New(params)
+	if err != nil {
+		log.Error().Err(err).Msg("unable to create stripe checkout session")
+
+		return h.BadRequest(ctx, err)
+	}
+
+	log.Warn().Msg("created stripe checkout session")
+	log.Warn().Msgf("sending back client secret %s", result.ClientSecret)
```
golanglemonade: not sure we want to log the client secret?

internal/httpserve/handlers/ent.go
```diff
@@ -501,3 +500,86 @@ func (h *Handler) getOrgByID(ctx context.Context, id string) (*ent.Organization,
 
 	return org, nil
 }
+
+// getOrgSettingByOrgID returns the organization settings from an organization ID and context
+func (h *Handler) getOrgSettingByOrgID(ctx context.Context, orgID string) (*ent.OrganizationSetting, error) {
+	orgGetCtx := privacy.DecisionContext(ctx, privacy.Allow)
```
golanglemonade: Why do we need to override and add `allow` here? This is called from an authorized handler, right?

If this is something we need a comment should be added about why.

internal/httpserve/handlers/ent.go
```diff
@@ -501,3 +500,86 @@ func (h *Handler) getOrgByID(ctx context.Context, id string) (*ent.Organization,
...
+	if err != nil {
+		log.Error().Err(err).Msg("error retrieving organization setting")
+
+		return nil, err
+	}
+
+	return setting, nil
+}
+
+func (h *Handler) fetchOrCreateStripe(context context.Context, orgsetting *ent.OrganizationSetting) (*stripe.Customer, error) {
```
golanglemonade: this function doesn't really belong in this file, this file is reusable queries to the database, but this is specific to the `checkout` handler (I think?) so it should probably live in there instead.

internal/httpserve/route/checkout.go
```diff
@@ -0,0 +1,57 @@
...
+		Method:      method,
+		Path:        path,
+		Middlewares: authMW,
+		Handler: func(c echo.Context) error {
+			return router.Handler.CheckoutSessionHandler(c)
+		},
+	}
+
+	//	uploadOperation := router.Handler.BindUploadBander()
+
```
golanglemonade:
```suggestion
```

internal/httpserve/route/checkout.go
```diff
@@ -0,0 +1,57 @@
...
+		Method:      method,
+		Path:        path,
+		Middlewares: authMW,
+		Handler: func(c echo.Context) error {
+			return router.Handler.CheckoutSuccessHandler(c)
+		},
+	}
+
+	//	uploadOperation := router.Handler.BindUploadBander()
+
```
golanglemonade:
```suggestion
```

pkg/entitlements/config.go
```diff
@@ -0,0 +1,13 @@
+package entitlements
+
+type Config struct {
+	Enabled bool `json:"enabled" koanf:"enabled" default:"false"`
```
golanglemonade, line 5: missing comment on `Enabled`

internal/ent/hooks/event.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	billingEmail, exists := props["billing_email"]
+	if exists && billingEmail != "" {
+		email := billingEmail.(string)
+
+		i := event.Client().(*entgen.Client).EntitlementManager.Client.Customers.List(&stripe.CustomerListParams{Email: &email})
+
+		if !i.Next() {
+			customer, err := event.Client().(*entgen.Client).EntitlementManager.Client.Customers.New(&stripe.CustomerParams{Email: &email})
```
golanglemonade: I see a `CreateCustomer` in the new `entitlements` package - should that be used here instead?

pkg/entitlements/subscriptions.go
```diff
@@ -0,0 +1,155 @@
...
+	subscription, err := sc.Client.Subscriptions.New(params)
+	if err != nil {
+		return nil, err
+	}
+
+	return subscription, nil
+}
+
+// ListStripeSubscriptions lists stripe subscriptions by customer
+func (sc *StripeClient) ListOrCreateStripeSubscriptions(customerID string) (*Subscription, error) {
```
golanglemonade: We should be consistent on the name, most of these functions don't say `Stripe` except this one. This should follow that 
```suggestion
func (sc *StripeClient) ListOrCreateSubscriptions(customerID string) (*Subscription, error) {
```

pkg/events/soiree/errors.go
```diff
@@ -19,4 +19,6 @@ var (
 	ErrEmitterClosed = errors.New("soiree is closed")
 	// ErrEmitterAlreadyClosed is returned when the soiree is already closed
 	ErrEmitterAlreadyClosed = errors.New("soiree is already closed")
+	// ErrPayloadisNotEvent is returned when the payload is not an event
+	ErrPayloadisNotEvent = errors.New("payload is not an event")
```
golanglemonade: not sure what this breaks, but `Is` should be capitalized (or omitted)
```suggestion
	ErrPayloadIsNotEvent = errors.New("payload is not an event")
```

pkg/events/soiree/event.go
```diff
@@ -96,11 +114,58 @@ type Properties map[string]interface{}
...
+	if p == nil {
+		p = NewProperties()
+	}
+
 	p[name] = value
+
 	return p
 }
+
+// Get a property from the Properties map
```
golanglemonade, line 131:
```suggestion
// GetKey from the Properties map
```

pkg/events/soiree/options.go
```diff
@@ -59,3 +59,9 @@ func WithErrChanBufferSize(size int) EventPoolOption {
 		m.SetErrChanBufferSize(size)
 	}
 }
+
+func WithClient(client interface{}) EventPoolOption {
```
golanglemonade, line 64: missing comment

internal/ent/hooks/event.go
matoszz: so i can see the colorz in the logz :)

internal/ent/hooks/event.go
matoszz: the orgsettingID is what's passed, the var name is just wrong

internal/ent/hooks/event.go
matoszz: this is the error handler func pattern so it's printing the error in return within a function

internal/ent/hooks/event.go
matoszz: do you have a suggestion on how to do that given i need to check to see if each field exists?

internal/ent/hooks/event.go
matoszz: I'm not sure I can do that as easily as you're suggesting (I've tried to simplify it as much as possible already) without creating more complex data mapping structures. Each of the `Update` commands are using different inputs based on whether or not there is a match because we don't want to stomp on a field that should not be updated (e.g. if a mutation does not contain the field we don't want to overwrite what's already in stripe)

internal/ent/hooks/event.go
matoszz: this is a psuedo reconciliation because if billing_email is present in the mutation, but the stripe ID does NOT match, we want to update the existing stripe record by ID with the email inside of the mutation. Because this is triggered when we first create the email in the settings and then have to handle updates to the email on an ongoing basis (and based on whether or not there is a stripe ID with an email that matches) we have to evaluate it both times

golanglemonade (approved):

PR #247, opened 2024-11-25
feat: risks and their permissions, program-members cli

golanglemonade:
- Adds program owned permissions (and updates the model) for `Risks` 
- Risks can be associated to program(s)
- Users with access to the program, inherit the same access to the risks
- Users in the organizations *do not* inherit access to the risk, they must have access to the program(s) or be assigned access via a group
- Adds the cli for risks, program members
- A program is now created as part of `task cli:user:all`:
- removes the auto retries in bk for tests, this was added previously for a flakey test that has since been addressed

```
task cli:user:all               
...
task: [cli:program:create] go run main.go program create -n "mitb program" -d "program for mitb"
  ID                          NAME          DESCRIPTION       STATUS       AUDITORREADY  AUDITORWRITECOMMENTS  AUDITORREADCOMMENTS  STARTDATE  ENDDATE  
```

Create a risk with no program
```
go run cmd/cli/main.go risk create -n risky-no-program                   
  ID                          NAME              PROGRAM(S)  RISKTYPE  BUSINESSCOSTS  IMPACT    LIKELIHOOD  MITIGATION  SATISFIES  
  01JDJ9YM51DZ776WR6HMTX78S9  risky-no-program                                       MODERATE  LIKELY                            
```

Create a risk with one program: 
```
 go run cmd/cli/main.go risk create -n risky-with-program -p 01JDJ9Y6ZF25PW9XP3T5ERQVEV
  ID                          NAME                PROGRAM(S)    RISKTYPE  BUSINESSCOSTS  IMPACT    LIKELIHOOD  MITIGATION  SATISFIES  
  01JDJA6HJDK6YPN1JDE1N9DMS8  risky-with-program  mitb program                           MODERATE  LIKELY 
```

Create (or update) a risk with multiple programs:
```
 go run cmd/cli/main.go risk create -n risky-with-programs-again --programs=01JDJ9Y6ZF25PW9XP3T5ERQVEV,01JDJA7WKY3KKD0W5T4D9R8SKV  
  ID                          NAME                       PROGRAM(S)          RISKTYPE  BUSINESSCOSTS  IMPACT    LIKELIHOOD  MITIGATION  SATISFIES  
  01JDJAMD4FGD8N4VYYKQCTXN3Q  risky-with-programs-again  mitb program,mitb2                           MODERATE  LIKELY    
```

Users that have access to the organization (e.g. the user created by `task user:all:another` can access and update the risk if they are added as a program admin: 

```
go run cmd/cli/main.go program-members create -p 01JDJA7WKY3KKD0W5T4D9R8SKV -u 01JDJAMMS2V2WXKXWZNAPCD4KB -r ADMIN
```

This user can now see two of the 4 risks created because they have access to one of the pgorams: 
```
(⎈ |default:default)➜  core git:(feat-risks) ✗ go run cmd/cli/main.go program get                                                                                
  ID                          NAME   DESCRIPTION  STATUS       AUDITORREADY  AUDITORWRITECOMMENTS  AUDITORREADCOMMENTS  STARTDATE  ENDDATE  
  01JDJA7WKY3KKD0W5T4D9R8SKV  mitb2               NOT_STARTED  false         false                 false                <nil>      <nil>    

(⎈ |default:default)➜  core git:(feat-risks) ✗ go run cmd/cli/main.go risk get    
  ID                          NAME                       PROGRAM(S)  RISKTYPE  BUSINESSCOSTS  IMPACT    LIKELIHOOD  MITIGATION  SATISFIES  
  01JDJAB2EDNDAXZFMGYQ722EQ6  risky-with-programs        mitb2                                MODERATE  LIKELY                             
  01JDJAMD4FGD8N4VYYKQCTXN3Q  risky-with-programs-again  mitb2                                MODERATE  LIKELY   
```


fga/model/model.fga
```diff
@@ -104,19 +104,25 @@ type control
...
     define can_edit: [user] or can_edit from parent
     define can_delete: [user] or can_delete from parent
-    define parent: [program]
+    define parent: [user, service, program]
+type risk
+  relations
+    define can_view: [user] or can_view from parent
+    define can_edit: [user] or can_edit from parent
+    define can_delete: [user] or can_delete from parent
+    define parent: [user, service, program]
```
matoszz, line 137: this should also include organization

matoszz (dismissed):

fga/model/model.fga
golanglemonade: We are not directly inheriting the `organization` permissions (meaning having view access to the organization, does not automatically give you access to the risk). Instead I've added groups to allow for additional access and removed the requirement for a risk to have have to be associated to a program (and can be associated to 1:many programs instead of just 1). 

This means: 
- When a risk is created, if there are no programs or groups associated, only the user who created the risk can view/edit the risk
- If a risk is associated to program(s) permissions to the risk are inherited from the program, so if a user has edit access to the program, they will have edit access to the risk associated (same for view permissions)
- Risks can be associated with groups as editors, viewers, or blocked which allows for additional control over the access

golanglemonade: We are not directly inheriting the `organization` permissions (meaning having view access to the organization, does not automatically give you access to the risk). Adding the `organization` here as a `parent` would give this access. Instead I've added groups to allow for additional access and removed the requirement for a risk to have have to be associated to a program (and can be associated to 1:many programs instead of just 1). 

This means: 
- When a risk is created, if there are no programs or groups associated, only the user who created the risk can view/edit the risk
- If a risk is associated to program(s) permissions to the risk are inherited from the program, so if a user has edit access to the program, they will have edit access to the risk associated (same for view permissions)
- Risks can be associated with groups as editors, viewers, or blocked which allows for additional control over the access

However, I did add the org owned mixin, which requires an organization ID to be associated with the risk, and the user to have edit access to the organization to create the risk. It will also mean that when an organization is deleted, the risks will be cascade deleted.

matoszz (approved):

PR #264, opened 2024-12-03
feat: permissions refactor, added create permissions

golanglemonade:
- Updates the fga model to allow for additional object creation permissions based on group membership, tests added to each create to test for access
- Updates the ent template to include a basic policy for new schemas
- Upgrades iam dependency (which cleans up a lot of unnecessary code in the auth_checks.go generated file and adds some helper wrappers)
- Adds a new policy package (`internal/ent/privacy/policy/base.go`) to make writing the policies in the schemas cleaner using `With` options
- Should be a 1:1 replacement in all the schemas of the old policies -> new policies - only additional create policies were added to schemas
- `mixins` in the schema package were all renamed to be prefixed with `mixin_` to make them easier to find (they need to live in the `schema` package because they have dependencies on the package itself
- Two mixins added/updated: 
    - `internal/ent/schema/mixin_createacess.go` : adds the to/from fields for org -> group access and adds the hooks for adding tuples to fga. This currently contains a manual list of objects types, I want to follow-up later to see if these can be pulled based on annotations in the schema instead
    - `internal/ent/schema/mixin_grouppermissions.go`: updates to add the reverse edges to the group schema so we don't have to manually add three fields for every editor/viewer/blocked group
 - Updates testing suite to output error logs when using the `-v` flag; removed the `-v` flag from the task command that runs in CI.


fga/model/model.fga
```diff
@@ -37,18 +39,52 @@ type organization
...
+    define can_create_procedure: can_edit or procedure_creator
+
+    define program_creator: [group#member]
+    define can_create_program: can_edit or program_creator
+
+    define risk_creator: [group#member]
+    define can_create_risk: can_edit or risk_creator
+
+    define template_creator: [group#member]
+    define can_create_template: can_edit or template_creator
```
matoszz, line 78: it looks like you hit all of the main "compliance" related objects but should we also add the other existing objects that don't yet have these types of permission structures? (e.g. webhooks, hush / secrets, entity, etc.?)

internal/ent/privacy/policy/base.go
```diff
@@ -0,0 +1,112 @@
...
+
+// WithPostPolicy overrides the post-policy to be executed.
+func WithPostPolicy(policy privacy.Policy) PolicyOption {
+	return func(policies *policies) {
+		policies.post = policy
+	}
+}
+
+// NewPolicy creates a privacy policy.
+func NewPolicy(opts ...PolicyOption) ent.Policy {
```
matoszz, line 82: sooo much easier to use <3

pkg/middleware/auth/auth.go
```diff
@@ -265,6 +265,10 @@ func isValidAPIToken(ctx context.Context, dbClient ent.Client, token string) (*a
 
 // getSubjectName returns the subject name for the user
 func getSubjectName(user *ent.User) string {
+	if user == nil {
+		return ""
```
matoszz, line 269: instead of returning empty wouldn't we want to progress down the same function setting subjectName to display name, etc.?

internal/ent/privacy/policy/base.go
```diff
@@ -0,0 +1,112 @@
...
+		opt(&policies)
+	}
+
+	return privacy.Policy{
+		Query:    policies.queryPolicy(),
+		Mutation: policies.mutationPolicy(),
+	}
+}
+
+func (policies policies) queryPolicy() privacy.QueryPolicy {
```
matoszz: just a nit but both of these being the same word makes it kind of confusing to read

internal/ent/privacy/policy/checks.go
```diff
@@ -0,0 +1,62 @@
...
+func CheckOrgWriteAccess() privacy.MutationRule {
+	return privacy.MutationRuleFunc(func(ctx context.Context, q ent.Mutation) error {
+		return rule.CheckOrgAccess(ctx, fgax.CanEdit)
+	})
+}
+
+// DenyQueryIfNotAuthenticated denies a query if the user is not authenticated
+func DenyQueryIfNotAuthenticated() privacy.QueryRule {
+	return privacy.QueryRuleFunc(func(ctx context.Context, q ent.Query) error {
+		if _, err := auth.GetAuthenticatedUserContext(ctx); err != nil {
```
matoszz: don't you need to return `AuthenticatedUser` and check if nil ? afaik from the function in the iam pacakge, the error is surfaced only if the `AuthenticatedUser` couldn't be fetched from the context but could still return nil?

fga/model/model.fga
golanglemonade: yes - these will get added when I go back through and add additional permissions (those objects dont even have basic permissions yet and didnt want to make this PR even larger. Going to do a final sweep of all schemas next to add the rest of the permissions

pkg/middleware/auth/auth.go
golanglemonade: if the `user` is `nil` we wouldn't have any information, we do fall back to display name below. This wasn't an actual issue we hit but I thought it was/had nil pointer errors I was trying to fix.

internal/ent/privacy/policy/checks.go
golanglemonade: I thought it would result the `ErrNoAuthUser` but adding a nil check to be sure

matoszz (approved):

PR #280, opened 2024-12-06
subscriptions updates and event hook refactor

matoszz:
- clean up of first implementation config / functional options
- adds a few helpers for working with ent mutations / hooks
- context wrapper + event hook refactor to only emit events on successful transaction commits
- update of entitlements library for better required field checks (e.g. don't create a stripe customer without a billing email)
- overhaul of existing organizationsettings event listener
- convenience wrappers in ent/hooks for "eventer"


internal/ent/hooks/helpers.go
```diff
@@ -4,11 +4,58 @@ import (
...
+			if err != nil {
+				return v, err
+			}
+
+			entvalue, ok := v.(T)
+
+			if ok {
+				err2 := hook(ctx, entvalue)
+				if err2 != nil {
+					log.Debug().Ctx(ctx).Err(err2).Msg("post mutation hook error")
```
golanglemonade, line 54: should this return the error ?

PR #298, opened 2024-12-09
fix deprecated functions from latest gqlen

golanglemonade:


pkg/testutils/client.go
```diff
@@ -106,11 +107,17 @@ func createAuthConfig(c *ent.Client) *auth.AuthOptions {
...
+	srv := handler.New(
 		graphapi.NewExecutableSchema(
 			graphapi.Config{Resolvers: graphapi.NewResolver(c, u)},
 		))
 
+	// add all the transports to the server
+	srv.AddTransport(transport.Options{})
+	srv.AddTransport(transport.GET{})
+	srv.AddTransport(transport.POST{})
+	srv.AddTransport(transport.MultipartForm{})
```
matoszz, line 119: don't we want to set this?

golanglemonade: These were the defaults from the deprecated function: https://github.com/99designs/gqlgen/blob/master/graphql/handler/server.go#L55-L58 

In the prod resolver, we already overrode them. For tests just putting it back to what it was

matoszz (approved):

PR #355, opened 2024-12-31
Feat stripe webhook receiver

matoszz:
- initial implementation of a route + handler which receives stripe webhooks, performs signature verification, and matches event by type to perform event unmarshaling
- current implementation creates internal event ID references and performs look ups to prevent double processing
- no actual update logic is being performed by the handler - future pull requests will add additional logic as necessary but this will allow is to begin receiving and logging webhooks
- other misc changes: pre-commit update, removal of unnecessary white space / line breaks, initialize stripe client with webhook secret in serverOpts

### DIFF
--- a/internal/httpserve/handlers/errors.go
+++ b/internal/httpserve/handlers/errors.go
@@ -65,8 +65,6 @@ var (
 	ErrNoBillingEmail = errors.New("no billing email found")
 	// ErrPersonalOrgsNoBilling is returned when the org ID looked up is a personal org
 	ErrPersonalOrgsNoBilling = errors.New("personal orgs do not have billing")
-	// ErrEventNotFound is returned when the event is not found
-	ErrEventNotFound = errors.New("generated: event not found")
 )
 
 var (
--- a/internal/httpserve/handlers/errors.go
+++ b/internal/httpserve/handlers/errors.go
@@ -17,77 +17,56 @@ import (
 var (
 	// ErrBadRequest is returned when the request cannot be processed
 	ErrBadRequest = errors.New("invalid request")
-
 	// ErrProcessingRequest is returned when the request cannot be processed
 	ErrProcessingRequest = errors.New("error processing request, please try again")
-
 	// ErrMissingRequiredFields is returned when the login request has an empty username or password
 	ErrMissingRequiredFields = errors.New("invalid request, missing username and/or password")
-
 	// ErrInvalidInput is returned when the input is invalid
 	ErrInvalidInput = errors.New("invalid input")
-
 	// ErrNotFound is returned when the requested object is not found
 	ErrNotFound = errors.New("object not found in the database")
-
 	// ErrMissingField is returned when a field is missing duh
 	ErrMissingField = errors.New("missing required field")
-
 	// ErrInvalidCredentials is returned when the password is invalid or missing
 	ErrInvalidCredentials = errors.New("credentials are missing or invalid")
-
 	// ErrUnverifiedUser is returned when email_verified on the user is false
 	ErrUnverifiedUser = errors.New("user is not verified")
-
 	// ErrUnableToVerifyEmail is returned when user's email is not able to be verified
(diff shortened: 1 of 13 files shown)


internal/httpserve/handlers/errors.go
```diff
@@ -17,77 +17,56 @@ import (
...
-
 	// ErrAlreadySwitchedIntoOrg is returned when a user attempts to switch into an org they are currently authenticated in
 	ErrAlreadySwitchedIntoOrg = errors.New("user already switched into organization")
-
 	// ErrNoBillingEmail is returned when the user has no billing email
 	ErrNoBillingEmail = errors.New("no billing email found")
 	// ErrPersonalOrgsNoBilling is returned when the org ID looked up is a personal org
 	ErrPersonalOrgsNoBilling = errors.New("personal orgs do not have billing")
+	// ErrEventNotFound is returned when the event is not found
+	ErrEventNotFound = errors.New("generated: event not found")
```
golanglemonade: This doesn't look used

internal/httpserve/handlers/webhook.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return ctx.String(http.StatusServiceUnavailable, fmt.Errorf("problem with request. Error: %w", err).Error())
+	}
+
+	event, err := webhook.ConstructEvent(payload, req.Header.Get("Stripe-Signature"), h.Entitlements.Config.StripeWebhookSecret)
+	if err != nil {
+		return ctx.String(http.StatusBadRequest, fmt.Errorf("error verifying webhook signature. Error: %w", err).Error())
+	}
+
+	exists, err := h.checkForEventID(ctx.Request().Context(), event.ID)
```
golanglemonade: nit: you already define `ctx.Request()` as `req` so we should use that
```suggestion
	exists, err := h.checkForEventID(req.Context(), event.ID)
```

internal/httpserve/handlers/webhook.go
```diff
@@ -0,0 +1,89 @@
...
+			// TODO unmarshall event data into internal event
+		}
+
+		meowevent, err := h.createEvent(ctx.Request().Context(), input)
+		if err != nil {
+			return h.InternalServerError(ctx, err)
+		}
+
+		log.Warn().Msgf("Internal event: %v", meowevent)
+	}
```
golanglemonade: this if isn't needed; if the `exists == true` you have a return, so everything would fall into this if 

Also don't think you want these warning logs? 

```suggestion
        if err := h.Entitlements.HandleEvent(req.Context(), &event); err != nil {
	        return h.InternalServerError(ctx, err)
        }
        
        input := ent.CreateEventInput{
	        EventID:   &event.ID,
	        EventType: "stripe",
	        // TODO unmarshall event data into internal event
        }
        
        meowevent, err := h.createEvent(ctx.Request().Context(), input)
        if err != nil {
	        return h.InternalServerError(ctx, err)
        }        
```

internal/httpserve/route/webhook.go
```diff
@@ -0,0 +1,30 @@
+package route
+
+import (
+	"net/http"
+
+	echo "github.com/theopenlane/echox"
+)
+
+// registerWebhookHandler registers a webhook endpoint handler behind the /stripe/ path for handling inbound event receivers from stripe
+func registerWebhookHandler(router *Router) (err error) {
```
golanglemonade, line 10: This is just for stripe to send us updates, right? Do we need to do any restrictions to this endpoint so its not exposed generally?

golanglemonade (commented):
just a few minor comments, otherwise LGTM! 🚀

internal/httpserve/route/webhook.go
matoszz: I'll add IP restrictions at the LB level in addition to the signature verification

golanglemonade (approved):

PR #373, opened 2025-01-10
add featurename field in addition to lookup key

matoszz:
- adds additional schema field to hold both the feature name and feature lookup key
- adds the lookup to fetch the feature name in the existing loop
- updates the organizationcustomer model to include both feature name and lookup keys

### DIFF
--- a/internal/ent/schema/organizationsetting.go
+++ b/internal/ent/schema/organizationsetting.go
@@ -66,6 +66,9 @@ func (OrganizationSetting) Fields() []ent.Field {
 		field.String("organization_id").
 			Comment("the ID of the organization the settings belong to").
 			Optional(),
+		field.Bool("email_notifications_enabled").
+			Comment("should we send email notifications").
+			Default(false),
 	}
 }
 
--- a/internal/ent/schema/organizationsetting.go
+++ b/internal/ent/schema/organizationsetting.go
@@ -66,9 +66,9 @@ func (OrganizationSetting) Fields() []ent.Field {
 		field.String("organization_id").
 			Comment("the ID of the organization the settings belong to").
 			Optional(),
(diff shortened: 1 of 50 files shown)


internal/ent/schema/organizationsetting.go
```diff
@@ -66,6 +66,9 @@ func (OrganizationSetting) Fields() []ent.Field {
 		field.String("organization_id").
 			Comment("the ID of the organization the settings belong to").
 			Optional(),
+		field.Bool("email_notifications_enabled").
+			Comment("should we send email notifications").
+			Default(false),
```
golanglemonade: should we default this to true? Seems like we'd want them on unless someone turns them off explicitly? Also I assume this is specific to billing, because its the only email address on the organization settings, but we should specify that

internal/ent/schema/orgsubscription.go
```diff
@@ -50,6 +50,9 @@ func (OrgSubscription) Fields() []ent.Field {
 		field.JSON("features", []string{}).
 			Comment("the features associated with the subscription").
 			Optional(),
+		field.JSON("feature_lookup_keys", []string{}).
+			Comment("the features associated with the subscription").
```
golanglemonade:
```suggestion
			Comment("the feature lookup keys associated with the subscription").
```

internal/ent/schema/organizationsetting.go
matoszz: I personally prefer not have to go through and disable notifications but to turn them on if i want them ¯\_(ツ)_/¯

golanglemonade: works for me

golanglemonade (dismissed):

golanglemonade (approved):

PR #400, opened 2025-01-22
fix up openapi spec, add security requirements, add webauthn

matoszz:
- removes unused models
- adds bind function for webauthn
- adds functions for security requirements
- updates tags

### DIFF
--- a/internal/httpserve/handlers/acccountroles.go
+++ b/internal/httpserve/handlers/acccountroles.go
@@ -78,17 +78,13 @@ func (h *Handler) BindAccountRoles() *openapi3.Operation {
 	roles.Description = "List roles a subject has in relation to an object"
 	roles.Tags = []string{"account"}
 	roles.OperationID = "AccountRoles"
-	roles.Security = &openapi3.SecurityRequirements{
-		openapi3.SecurityRequirement{
-			"bearer": []string{},
-		},
-	}
+	roles.Security = AllSecurityRequirements()
 
 	h.AddRequestBody("AccountRolesRequest", models.ExampleAccountRolesRequest, roles)
 	h.AddResponse("AccountRolesReply", "success", models.ExampleAccountRolesReply, roles, http.StatusOK)
 	roles.AddResponse(http.StatusInternalServerError, internalServerError())
 	roles.AddResponse(http.StatusBadRequest, badRequest())
-	roles.AddResponse(http.StatusUnauthorized, unauthorized())
+	roles.AddResponse(http.StatusBadRequest, invalidInput())
 
 	return roles
 }
--- a/internal/httpserve/handlers/acccountrolesorganization.go
+++ b/internal/httpserve/handlers/acccountrolesorganization.go
@@ -99,15 +99,11 @@ func (h *Handler) BindAccountRolesOrganization() *openapi3.Operation {
 	orgRoles.Description = "List roles a subject has in relation to the authenticated organization"
 	orgRoles.Tags = []string{"account"}
 	orgRoles.OperationID = "AccountRolesOrganization"
-	orgRoles.Security = &openapi3.SecurityRequirements{
-		openapi3.SecurityRequirement{
-			"bearer": []string{},
-		},
-	}
+	orgRoles.Security = AllSecurityRequirements()
 
 	orgRoles.AddResponse(http.StatusInternalServerError, internalServerError())
 	orgRoles.AddResponse(http.StatusBadRequest, badRequest())
-	orgRoles.AddResponse(http.StatusUnauthorized, unauthorized())
+	orgRoles.AddResponse(http.StatusBadRequest, invalidInput())
 
 	return orgRoles
 }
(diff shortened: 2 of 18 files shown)


internal/httpserve/handlers/acccountroles.go
```diff
@@ -78,17 +78,13 @@ func (h *Handler) BindAccountRoles() *openapi3.Operation {
...
-			"bearer": []string{},
-		},
-	}
+	roles.Security = AllSecurityRequirements()
 
 	h.AddRequestBody("AccountRolesRequest", models.ExampleAccountRolesRequest, roles)
 	h.AddResponse("AccountRolesReply", "success", models.ExampleAccountRolesReply, roles, http.StatusOK)
 	roles.AddResponse(http.StatusInternalServerError, internalServerError())
 	roles.AddResponse(http.StatusBadRequest, badRequest())
-	roles.AddResponse(http.StatusUnauthorized, unauthorized())
```
golanglemonade, line 91: This endpoint requires authorization, why are you removing the authorized response?

internal/httpserve/handlers/acccountrolesorganization.go
```diff
@@ -99,15 +99,11 @@ func (h *Handler) BindAccountRolesOrganization() *openapi3.Operation {
...
-	orgRoles.Security = &openapi3.SecurityRequirements{
-		openapi3.SecurityRequirement{
-			"bearer": []string{},
-		},
-	}
+	orgRoles.Security = AllSecurityRequirements()
 
 	orgRoles.AddResponse(http.StatusInternalServerError, internalServerError())
 	orgRoles.AddResponse(http.StatusBadRequest, badRequest())
-	orgRoles.AddResponse(http.StatusUnauthorized, unauthorized())
```
golanglemonade, line 110: same here

internal/httpserve/handlers/forgotpassword.go
```diff
@@ -97,14 +97,15 @@ func (h *Handler) storeAndSendPasswordResetToken(ctx context.Context, user *User
 func (h *Handler) BindForgotPassword() *openapi3.Operation {
 	forgotPassword := openapi3.NewOperation()
 	forgotPassword.Description = "ForgotPassword is a service for users to request a password reset email. The email address must be provided in the POST request and the user must exist in the database. This endpoint always returns 200 regardless of whether the user exists or not to avoid leaking information about users in the database"
-	forgotPassword.Tags = []string{"authentication"}
```
golanglemonade, line 100: these are tags to allow the ability to group endpoints, I don't think they should be so specific they all end up not overlapping at all? Thoughts?

for context, I had added these when looking at doc features where you could group by tags (I ended up not using it though)

internal/httpserve/handlers/openapi.go
```diff
@@ -86,3 +100,60 @@ func (h *Handler) AddResponse(name string, description string, body interface{},
...
+}
+
+// apiKeySecurity is used to add a apiKey security definition to the OpenAPI schema
+func APIKeySecurity() *openapi3.SecurityRequirements {
+	return &openapi3.SecurityRequirements{
+		openapi3.SecurityRequirement{
+			"apiKey": []string{},
+		},
+	}
+}
```
golanglemonade, line 138: We always use the `Authorization: Bearer <>`  and not an `apiKey` header, I had even previously removed this when I was fixing the spec to make it work because it results in invalid options in the docs:

https://github.com/theopenlane/core/pull/342/files#diff-535cccf0dfdd646bce151a378c6cd6b156f1592bfc93cff1d72bb97ebea17b04L96

internal/httpserve/handlers/verifysubscribe.go
```diff
@@ -160,7 +160,7 @@ func (h *Handler) BindVerifySubscriberHandler() *openapi3.Operation {
 	h.AddResponse("VerifySubscriptionReply", "success", models.ExampleVerifySubscriptionResponse, verify, http.StatusOK)
 	verify.AddResponse(http.StatusInternalServerError, internalServerError())
 	verify.AddResponse(http.StatusBadRequest, badRequest())
-	verify.AddResponse(http.StatusCreated, created())
```
golanglemonade, line 163: Created should not be removed: https://github.com/theopenlane/core/blob/76a0f8f692f291041861f8b966d90ed545a1602f/internal/httpserve/handlers/verifysubscribe.go#L65

internal/httpserve/handlers/acccountroles.go
matoszz: these are return responses, and we dont return this response in any of our code paths. It's got nothing to do with whether or not the endpoint is authenticated or not.

internal/httpserve/handlers/acccountrolesorganization.go
matoszz: same answer ^

internal/httpserve/handlers/forgotpassword.go
matoszz: endpoints should be grouped by the functional purpose, generally. The existing tags were super generic (e.g. more than half were categorized under "authentication" when the forgot password request is really not directly authentication related but account management). From my experience the tags are pretty granular, to the point of the tag being unique to the endpoint itself

internal/httpserve/handlers/openapi.go
matoszz: There is no header specified within the scheme; if we allow both JWT authentication and API key authentication we need to specify both regardless of the placement within the http header

golanglemonade (approved):

PR #457, opened 2025-02-15
feat: allow organizations to be restricted to a email domains for members

golanglemonade:
- Removes the use of all the `entfga.CheckReadAccess[*generated.ControlQuery](),` checks on the policies; these were skipped many times and interceptors are in place to filter queries based on org owned + object owned settings. Most policies were changed to `privacy.AlwaysAllowRule()` instead. I plan to update the default from `AlwaysDenyRule` to `AlwaysAllowRule` once all default policies are in place (we have a few schemas with no policies and this prevents them from being used today

- Allows organizations to set `allowed_domains` and only users with that email domain can be invited + login to the organization
- Uses an fga condition to enforce the domain check:
```
condition email_domains_allowed(email_domain: string, allowed_domains: list<string>) {
  allowed_domains == [] || email_domain == "" || email_domain in allowed_domains
}

...

# organization policies to restrict access
define access: [organization#member with email_domains_allowed]
define can_view: [service, user with time_based_grant] or (member and access) or owner or can_edit or can_view from parent
define can_edit: [service] or (admin and access)  or owner or can_edit from parent
```
In human words, this changes the `can_view` check (which is what we use to check read access (and not the roles (member/admin/owner)) to require `access` to also be true for members - which uses the condition. This also changes the `can_edit` with the same restrictions.  The only role that gets around the domain check is the `owner` because this would cause issues - I can incorporate that if we think we should. 

The condition is skipped when there are no allowed_domains (all domains are allowed), when email_domain in the check is empty (allows for list requests). 

- Invites + Directly adding org members where the email domains don't match will return an error
```
go run cmd/cli/main.go org-members create -u 01JM61N70XH11MQVBTMXSEP2MS
Error: {"networkErrors":null,"graphqlErrors":[{"message":"email domain not allowed in organization","path":["createOrgMembership"]}]}
exit status 1
```
- Logging in, where you can no longer access your default org because of a domain restriction, will update the default org to the user's personal org
- Switching to an org where you are no longer allowed because of a domain restriction will return an error
```
 go run cmd/cli/main.go switch -t01JMCWZ8XC4YR7KATB3TYAS37D
Error: unable to authenticate (status 401): you are not authorized to perform this action
```


fga/model/model.fga
```diff
@@ -31,23 +31,26 @@ type organization
   relations
     # main roles
     define admin: [user] or admin from parent
-    define member: [user] or owner or admin or member from parent
+    define member:  [user] or owner or admin or member from parent
```
matoszz: intentionally adding 2 spaces?
```suggestion
    define member: [user] or owner or admin or member from parent
```

internal/ent/hooks/organizationsetting.go
```diff
@@ -0,0 +1,205 @@
...
+	return hook.If(func(next ent.Mutator) ent.Mutator {
+		return ent.MutateFunc(func(ctx context.Context, m ent.Mutation) (ent.Value, error) {
+			retVal, err := next.Mutate(ctx, m)
+			if err != nil {
+				return nil, err
+			}
+
+			orgID := ""
+
+			allowedDomains := []string{}
```
matoszz, line 33: any reason this couldn't be `make([]string, 0, len(setting.AllowedEmailDomains)` ?

internal/ent/hooks/organizationsetting.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+
+			if err := updateOrgConditionalTuples(ctx, m, orgID, allowedDomains); err != nil {
+				return nil, err
+			}
+
+			return retVal, nil
+		})
+	},
+		hook.HasOp(ent.OpCreate),
```
matoszz, line 82: does this actually need 2 fully separate hooks or could you simply add a conditional check to see if the `AllowedEmailDomains` is set or is being set?

internal/ent/hooks/organizationsetting.go
```diff
@@ -0,0 +1,205 @@
...
+//
+//	name: email_domain_allowed
+//	context:
+//	  allowed_domains: []
+func updateOrgConditionalTuples(ctx context.Context, m ent.Mutation, orgID string, allowedEmailDomains []string) error {
+	tk := fgax.TupleRequest{
+		ObjectID:         orgID,
+		ObjectType:       generated.TypeOrganization,
+		SubjectType:      generated.TypeOrganization,
+		SubjectID:        orgID,
```
matoszz, line 162: both objectID and subjectID are `orgID`?

internal/ent/hooks/organizationsetting.go
```diff
@@ -0,0 +1,205 @@
...
+// to the organization based on the email domain
+func HookOrganizationCreatePolicy() ent.Hook {
+	return hook.If(func(next ent.Mutator) ent.Mutator {
+		return ent.MutateFunc(func(ctx context.Context, m ent.Mutation) (ent.Value, error) {
+			retVal, err := next.Mutate(ctx, m)
+			if err != nil {
+				return nil, err
+			}
+
+			orgID := ""
```
matoszz, line 31: why does this need to be set as empty instead of setting it as the return of `getOrgIDFromSettingMutation`? (e.g. `orgID, err := getOrgIDFromSettingMutation(ctx, m, retVal)`

internal/ent/interceptors/organizationsetting.go
```diff
@@ -17,17 +18,26 @@ func InterceptorOrganizationSetting() ent.Interceptor {
...
 			return nil
 		}
 
-		orgID, err := auth.GetOrganizationIDFromContext(ctx)
+		orgIDs, err := auth.GetOrganizationIDsFromContext(ctx)
 		if err != nil {
 			return err
 		}
 
+		if len(orgIDs) == 0 {
```
matoszz: this feels a little strange... why do we need 2 functions for `GetOrganizationIDs` and `GetOrganizationID`? It seems like we should update the IAM package `AuthenticatedUser` to only have a `[]string` field with 1 or many organization IDs instead of duplicate checks?

internal/ent/interceptors/user.go
```diff
@@ -110,8 +111,9 @@ func filterUsingFGA(ctx context.Context, q *generated.UserQuery) error {
 
 	for _, orgID := range orgIDs {
 		req := fgax.ListRequest{
-			ObjectID:   orgID,
-			ObjectType: "organization",
+			ObjectID:         orgID,
+			ObjectType:       generated.TypeOrganization,
+			ConditionContext: utils.NewOrganizationContextKey(""), // use an empty domain context on list
```
matoszz, line 116: what is `NewOrganizationContextKey` doing?

internal/ent/privacy/rule/organization.go
```diff
@@ -14,11 +15,53 @@ import (
...
+		return err
+	}
+
+	if len(requestedOrgs) == 0 {
+		// return nil if no organizations were found
+		// to allow the next check to run
+		return nil
+	}
+
+	for _, org := range requestedOrgs {
```
matoszz, line 49: wouldn't it be simplier to make the check `if len(requestedOrgs) != 0 { for _, org := .....`

internal/ent/privacy/utils/contextchecks.go
```diff
@@ -0,0 +1,34 @@
+package utils
+
+import "strings"
+
+// Context keys for authorization checks
+const (
+	emailDomainContextKey = "email_domain"
```
matoszz, line 8: any reason not to use `contextx` for this over string literals?

internal/ent/schema/control.go
```diff
@@ -123,11 +124,16 @@ func (Control) Annotations() []schema.Annotation {
 	}
 }
 
+// Interceptors of the Controls
+func (Control) Interceptors() []ent.Interceptor {
```
matoszz: why include this if it's just empty?

internal/ent/schema/documentdata.go
```diff
@@ -65,15 +69,20 @@ func (DocumentData) Annotations() []schema.Annotation {
 		entgql.QueryField(),
 		entgql.RelayConnection(),
 		entgql.Mutations(entgql.MutationCreate(), (entgql.MutationUpdate())),
-		entfga.OrganizationInheritedChecks(), // TODO(sfunk): update to template checks instead of org checks
+		entfga.SelfAccessChecks(),
 	}
 }
 
+// Interceptors of the DocumentData
+func (DocumentData) Interceptors() []ent.Interceptor {
```
matoszz: same question with this one

internal/ent/schema/integration.go
```diff
@@ -73,11 +74,16 @@ func (Integration) Mixin() []ent.Mixin {
 	}
 }
 
+// Interceptors of the Integration
+func (Integration) Interceptors() []ent.Interceptor {
```
matoszz: I won't put the same comment on every one of these but not sure why we need to include the interceptor if it's just empty; we don't include hooks, other types in various schemas if there is no actual hook that's called

internal/ent/schema/organizationsetting.go
```diff
@@ -69,6 +70,9 @@ func (OrganizationSetting) Fields() []ent.Field {
 		field.Bool("billing_notifications_enabled").
 			Comment("should we send email notifications related to billing").
 			Default(true),
+		field.Strings("allowed_email_domains").
```
matoszz, line 73: there should probably be a `Validate` function on this field type given domains follow pretty standard rulesets on formatting

internal/ent/privacy/utils/contextchecks.go
golanglemonade: its a string set in fga; not a golang context.

internal/ent/hooks/organizationsetting.go
golanglemonade: On create; we call it no matter if its set or not; we need to create the conditional tuple always even if no domains are set. And this can happen on either a Org or Org Setting mutation; it seemed pretty confusing to combine it all into one. The Update is only on org settings; you'll see in the schemas one of these is on both organization and organization setting; the other is just on org setting

internal/ent/hooks/organizationsetting.go
golanglemonade: Yes; it's a relation to itself

internal/ent/hooks/organizationsetting.go
golanglemonade: yes; its not always a setting; so we need to check the type first of the mutation.

internal/ent/hooks/organizationsetting.go
golanglemonade: because we get the org id from either the settings or organization mutation after we determine the type; but then use the variable outside the scope of the switch statement

internal/ent/interceptors/organizationsetting.go
golanglemonade: thats pretty out of scope of this PR; I'll think about the impcliations of that; but we use the single organization to verify requests; but a Personal Access Token is authorized for multiple organization(s) vs. api token + JWTs which are a single organization; in most cases we use the single org; but for filters on the org level with use the list

internal/ent/privacy/rule/organization.go
golanglemonade: if the length is 0; the for loop is skipped.

internal/ent/hooks/organizationsetting.go
golanglemonade: adding comments around this

internal/ent/interceptors/organizationsetting.go
golanglemonade: good call, but because I don't actually need this. I forgot I always set orgIDs, but orgID (singular) is not in the case of personal access token. updating 

ref: https://github.com/theopenlane/core/blob/main/pkg/middleware/auth/auth.go#L181

internal/ent/interceptors/user.go
golanglemonade: its a helper function that is creating this map needed for the FGA check context; it takes in either a domain or an email and creates the map. 

```
map[string]any{
		"email_domain": "theopenlane.io",
	}
```

in this one instance; when we are doing a list users; we bypass this check because its not relevant; but in other places its adding that context to the check. 

If you look at the fga tests, you'll see the checks now have the context of the email domain of the user we are checking the access of:

```
    check:
      - user: user:ulid-of-owner
        object: organization:openlane
        context:
          email_domain: "example.com"
```

internal/ent/schema/organizationsetting.go
golanglemonade: I even already have a function...used in this same schema :-D

PR #465, opened 2025-02-19
Feature: initial domain scanning package / subpackages for asset identification and classification

matoszz:
The `sleuth` package is a comprehensive suite of tools designed for DNS and certificate analysis. It leverages various open-source projects to provide detailed reports on DNS records, subdomain enumeration, and certificate transparency logs. The package is modular, with subpackages dedicated to specific tasks such as DNS queries (`dnsx`) or (`tech`) to identify technology on http payloads / websites.

This PR was limited in scope to the creation of the subpackages that would perform the "jobs" we'd launch. The follow on PR's will include:
- riverqueue jobs
- updates to the ent schema fields for entities and/or a new schema for "systems" (or similar)
- updates to the organization creation mutations to insert jobs
- probably a lot more stuff but you get the general idea

#### Subpackages and their functionalities:

1. **dnsx**: 
   - Provides DNS query capabilities for various record types including A, AAAA, MX, TXT, NS, CNAME, and more.
   - Supports querying for DMARC and DKIM records.
   - Includes functionality for checking CDN providers.
   - Example output:
     ```go
     // Example output for DNSRecordsReport
     report := DNSRecordsReport{
         Domain: "example.com",
         DNSRecords: &DNSRecords{
             A: []*DNSRecord{
                 {Name: "example.com", TTL: 300, Type: "A", Value: "93.184.216.34"},
             },
             // ...other record types...
         },
         Errors: []string{},
     }
     ```

2. **certx**:
   - Queries crt.sh for all certificates associated with a given domain.
   - Provides detailed information about each certificate including issuer, common name, and validity period.
   - Example output:
     ```go
     // Example output for CertsReport
     report := CertsReport{
         Domain: "example.com",
         Certificates: []CertificateRecord{
             {IssuerName: "Let's Encrypt", CommonName: "example.com", NotBefore: "2023-01-01", NotAfter: "2023-12-31"},
         },
         Errors: []string{},
     }
     ```

#### TODOs and Follow-up Items:
- Implement the `crt.sh` subpackage for certificate reporting or determine if it's needed
- Enhance error handling and logging across the package
- Add more granular control over DNS query types and options
- Add tests
- Create "report" outputs that can be loaded into our ent schemas
- Add metrics / statistic functions
- Add guards / methods for robust rate limit / back off to avoid requests getting flagged as malicious


pkg/sleuth/dnsx/dnsx.go
```diff
@@ -0,0 +1,257 @@
...
+	miekgdns "github.com/miekg/dns"
+	"github.com/projectdiscovery/cdncheck"
+	pddnsx "github.com/projectdiscovery/dnsx/libs/dnsx"
+	"github.com/rs/zerolog/log"
+)
+
+// DNSX is a wrapper around the dnsx library for client initialization and functional options settings
+type DNSX struct {
+	// Resolver is the dnsx resolver
+	Client *pddnsx.DNSX
```
golanglemonade:
```suggestion
	// Client is the dnsx resolver
	Client *pddnsx.DNSX
```

pkg/sleuth/dnsx/dnsx.go
```diff
@@ -0,0 +1,257 @@
...
+	dkimRecords := DNSRecords{}
+
+	for _, selector := range selectors {
+		dkimRecordForSelector, err := d.getDNSRecords(selector+"._domainkey."+domain, []uint16{miekgdns.TypeTXT})
+
+		if err != nil {
+			errors = append(errors, err.Error())
+		}
+
+		dkimRecords.Txt = append(dkimRecords.Txt, dkimRecordForSelector.Txt...)
```
golanglemonade, line 107: since you are continuing on even in the case of an error, is there a chance `dkimRecordForSelector` is nil and that will fail? Should this be an else? e.g. 

```
if err != nil {
	errors = append(errors, err.Error())
} else {
	dkimRecords.Txt = append(dkimRecords.Txt, dkimRecordForSelector.Txt...)
}
```

pkg/sleuth/dnsx/dnsx.go
```diff
@@ -0,0 +1,257 @@
...
+
+	return report, nil
+}
+
+// LookupCDN checks if the given domain is a CDN and returns the CDN name and type
+func (d *DNSX) LookupCDN(domain string) (value, cdnType string, err error) {
+	if net.ParseIP(domain) == nil {
+		ips, err := d.Lookup(domain)
+		if err != nil {
+			return "", "", err
```
golanglemonade: nit: since these are named returns you can just do a return on all these that are `return "", "", err` or `value, cdnType, nil`  (otherwise don't name your return vars)

```suggestion
			return
```

pkg/sleuth/dnsx/dnsx.go
```diff
@@ -0,0 +1,257 @@
...
+			for i, record := range records {
+				val, cdnType, err := d.LookupCDN(record)
+				if err != nil {
+					log.Error().Msgf("error looking up CDN for %s: %v", record, err)
+					continue
+				}
+
+				cdn := fmt.Sprintf("%s (%s)", val, cdnType)
+
+				if cdn != "" {
```
golanglemonade: cdn will never be "", if both values are empty it will be `" ()"`. I think you want to check `val` and `cdnType` before creating the `cdn` var

pkg/sleuth/dnsx/options.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// WithQuestionTypes sets the question types for the DNSX client
+func WithQuestionTypes(types []uint16) Option {
+	return func(o *Options) {
+		o.QuestionTypes = types
+	}
+}
+
+// WithTrace sets the trace option for
```
golanglemonade: for... ? :)

pkg/sleuth/models.go
```diff
@@ -0,0 +1,83 @@
+package sleuth
+
+// DNSRecord represents a DNS record with its associated properties
+type DNSRecord struct {
+	Name  string `json:"name" url:"name"`
```
golanglemonade: whats the `url` tag for?

pkg/sleuth/models.go
```diff
@@ -0,0 +1,83 @@
+package sleuth
```
golanglemonade: at a quick glance, it looks like these are the same models as what is in `pkg/sleuth/dnsx/models.go`?

golanglemonade (commented):
LGTM, just a few minor comments/questions 🚀

golanglemonade (approved):

golanglemonade (approved):

PR #472, opened 2025-02-20
feat: task schema update, notes schema

golanglemonade:
- Removes priority field from tasks
- Adds category field as a string to tasks - debated using an enum, but think it might be better to keep it free form; but have the UI give choices + other field to direct users.  This will be things like "evidence upload", "risk review", etc. 
- Allows assigner to be empty, creates tuples when its set. The assigner no longer needs to match the user who created the take. User who created the task will get `parent` tuple - giving them full CRUD ability to the task.
- Updates details type to `Text` from `JSON` - it's just a text field, no need for JSON
- Adds notes authz policies as parent owned, tested with tasks. This will be expanded as more objects get edges to notes. Notes will be:
    - view + edit + delete by user who created the note
    - viewable by users with inherited permissions (e.g. if linked to a task, if I can see the task, I can see the note but I cannot edit) - slightly different from other parent inherited permissions where you would inherit edit permissions from the parent
    - view + edit by org owner 
- Adds comments (using notes) to tasks schema
- Updates the org owner_id to be Optional (e.g. not required on GraphQL request) but NotEmpty - to allow the field to be omitted on request. This affects all object owned schemas with `WithOrganizationOwner: true`
- moves all fga tests tuples to their own files. This makes it easier to add the tuples to a local fga playground and see the interaction between the tuples and the assertions

Add Comment to a task: 

```
go run cmd/cli/main.go task update -i 01JMMQ6A3AS63E6SEJ6RSRWH22 -c "hellllo"
...
go run cmd/cli/main.go task get  -z json
{
  "tasks": {
    "edges": [
      {
        "node": {
          "assignee": {
            "firstName": "matt",
            "id": "01JMMQ5N17210T635QFR6B21CX",
            "lastName": "anderson"
          },
          "assigner": {
            "firstName": "matt",
            "id": "01JMMQ5RYXWR3YSV86XB21WFJG",
            "lastName": "anderson"
          },
          "category": "",
          "comments": [
            {
              "createdAt": "2025-02-21T10:00:08.285371-07:00",
              "createdBy": "01JMMQ5RYXWR3YSV86XB21WFJG",
              "displayID": "NTE-RHCM75",
              "id": "01JMMQ7HPXZCRW0PR16PFVR9BP",
              "text": "hellllo",
              "updatedAt": "2025-02-21T10:00:08.285152-07:00",
              "updatedBy": "01JMMQ5RYXWR3YSV86XB21WFJG"
            }
          ],
          "createdAt": "2025-02-21T09:59:27.722282-07:00",
          "createdBy": "01JMMQ5RYXWR3YSV86XB21WFJG",
          "description": "",
          "details": "",
          "displayID": "TSK-TXWGQP",
          "id": "01JMMQ6A3AS63E6SEJ6RSRWH22",
          "status": "OPEN",
          "title": "Review access to systems - CC5.2",
          "updatedAt": "2025-02-21T10:08:25.435797-07:00",
          "updatedBy": "01JMMQ5RYXWR3YSV86XB21WFJG"
        }
      }
    ]
  }

```


internal/ent/schema/mixin_objectowned.go
```diff
@@ -95,7 +98,9 @@ func (o ObjectOwnedMixin) Fields() []ent.Field {
 		fields = append(fields,
 			field.String("owner_id").
 				Comment("the ID of the organization owner of the object").
-				NotEmpty())
+				Immutable(). // Immutable because it is set on creation and never changes
```
matoszz, line 101: won't this be problematic when we create functionality for organization ownership transfers or other object ownership trasfers? like if someone was previously the owner of the org and left the organization someone else should be able to take over?

internal/ent/schema/task.go
```diff
@@ -84,12 +84,15 @@ func (Task) Edges() []ent.Edge {
 		edge.From("assigner", User.Type).
 			Ref("assigner_tasks").
 			Field("assigner_id").
+			Immutable().
```
matoszz: won't this make it so we can't change the assignment?

internal/ent/schema/mixin_objectowned.go
golanglemonade: This is the organization ID, which should never change, not the user ID of the owner

internal/ent/schema/task.go
golanglemonade: No, this will make it so the assigner can't change, the user who made the task. Not the user who is assigned the task

golanglemonade: Updated to allow the assigner to be changed, empty on creation

matoszz (approved):

PR #492, opened 2025-02-27
schema changes

golanglemonade:
# Schema Updates
- Updates/Adds the following schemas:
  - Standards
     - currently limited to only allow creation by system admins; will expand this in a follow-up when I add tests for org-owned standards
  - Controls
    - common `fields` list [shared](https://github.com/theopenlane/core/pull/492/files#diff-570b0e54d8ca79acb055ad30da0782156c3910afb370a36731df184d768f053a) with `subcontrols` with a few additional fields added directly to each schema file
    - can be attached to a program, standard, or just org owned
  - Sub-controls
     - same as controls, requires a parent control ID (changes to many to a single parent control id)
  - Control Implementation
     - edge for controls with implementation details
  - Mapped Controls
     - edge for controls that will hold mappings of controls between standards and can be expanded to have additional details on the mappings
   -  Control Objectives
      - can be linked to 0:many controls 
          
# Standards

- Adds standards to FGA Model
- Tested creation and retrieval by created a script (will be added to another repo) to seed system owned standards from our supported frameworks

```
{
  "data": {
    "standard": {
      "id": "01JNHNSFXFPP4M1ACTDAGGNJ2D",
      "link": "https://csrc.nist.gov/publications/detail/sp/800-53/rev-5/final",
      "name": "NIST 800-53 Rev 5: Security and Privacy Controls for Information Systems and Organizations",
      "ownerID": "",
      "shortName": "NIST 800-53",
      "systemOwned": true,
      "isPublic": true,
      "tags": [],
      "version": "5.1.1",
      "governingBody": "National Institute of Standards and Technology (NIST)",
      "freeToUse": true,
      "framework": "NIST-SP-800-53-r5",
      "domains": [
        "Access Control",
        "Awareness and Training",
        "Audit and Accountability",
        "Assessment, Authorization, and Monitoring",
        "Configuration Management",
        "Contingency Planning",
        "Identification and Authentication",
        "Incident Response",
        "Maintenance",
        "Media Protection",
        "Physical and Environmental Protection",
        "Planning",
        "Program Management",
        "Personnel Security",
        "Personally Identifiable Information Processing and Transparency",
        "Risk Assessment",
        "System and Services Acquisition",
        "System and Communications Protection",
        "System and Information Integrity",
        "Supply Chain Risk Management"
      ],
[...]

PR #501, opened 2025-03-05
feat: standards + controls implementation; system-admins

golanglemonade:
Depends on https://github.com/theopenlane/iam/pull/178 before merge

# Schema Updates
- Updates/Adds the following schemas:
  - Standards
     - currently limited to only allow creation by system admins; will expand this in a follow-up when I add tests for org-owned standards
  - Controls
    - common `fields` list [shared](https://github.com/theopenlane/core/pull/492/files#diff-570b0e54d8ca79acb055ad30da0782156c3910afb370a36731df184d768f053a) with `subcontrols` with a few additional fields added directly to each schema file
    - can be attached to a program, standard, or just org owned
  - Sub-controls
     - same as controls, requires a parent control ID (changes to many to a single parent control id)
  - Control Implementation
     - edge for controls with implementation details
  - Mapped Controls
     - edge for controls that will hold mappings of controls between standards and can be expanded to have additional details on the mappings
   -  Control Objectives
      - can be linked to 0:many controls 
          
# Standards

- Adds standards to FGA Model
- Tested creation and retrieval by created a script (will be added to another repo) to seed system owned standards from our supported frameworks

<details>
<summary>Get Standard With Controls</summary>

```
{
  "data": {
    "standard": {
      "id": "01JNHNSFXFPP4M1ACTDAGGNJ2D",
      "link": "https://csrc.nist.gov/publications/detail/sp/800-53/rev-5/final",
      "name": "NIST 800-53 Rev 5: Security and Privacy Controls for Information Systems and Organizations",
      "ownerID": "",
      "shortName": "NIST 800-53",
      "systemOwned": true,
      "isPublic": true,
      "tags": [],
      "version": "5.1.1",
      "governingBody": "National Institute of Standards and Technology (NIST)",
      "freeToUse": true,
      "framework": "NIST-SP-800-53-r5",
      "domains": [
        "Access Control",
        "Awareness and Training",
        "Audit and Accountability",
        "Assessment, Authorization, and Monitoring",
        "Configuration Management",
        "Contingency Planning",
        "Identification and Authentication",
        "Incident Response",
        "Maintenance",
        "Media Protection",
        "Physical and Environmental Protection",
        "Planning",
        "Program Management",
        "Personnel Security",
        "Personally Identifiable Information Processing and Transparency",
        "Risk Assessment",
        "System and Services Acquisition",
        "System and Communications Protection",
        "System and Information Integrity",
        "Supply Chain Risk Management"
      ],
      "description": "This publication provides a catalog of security and privacy controls for information systems and organizations to protect organizational operations and assets, individuals, other organizations, and the Nation from a diverse set of threats and risks, including hostile attacks, human errors, natural disasters, structural failures, foreign intelligence entities, and privacy risks. The controls are flexible and customizable and implemented as part of an organization-wide process to manage risk. The controls address diverse requirements derived from mission and business needs, laws, executive orders, directives, regulations, policies, standards, and guidelines. Finally, the consolidated control catalog addresses security and privacy from a functionality perspective (i.e., the strength of functions and mechanisms provided by the controls) and from an assurance perspective (i.e., the measure of confidence in the security or privacy capability provided by the controls). Addressing functionality and assurance helps to ensure that information technology products and the systems that rely on those products are sufficiently trustworthy.",
      "controls": [
        {
          "category": "",
          "categoryID": "ac",
          "controlQuestions": [
            "Does the organization establish, maintain and disseminate cybersecurity and data protection policies, standards and procedures?",
            ...
```

</details>

# Controls 

- Cli updated to get by ref-code for controls
- Cli updated to get by ref-code of subcontrol or parent control for subcontrols

<details>
<summary>Get Control By Ref Code</summary>

```
go run cmd/cli/main.go control get -z json -r ac-3.3
{
  "controls": {
    "edges": [
      {
        "node": {
          "assessmentMethods": [
            {
              "id": "AC-03(03)-Examine",
              "method": "Access control policy\n\nmandatory access control policies\n\nprocedures addressing access enforcement\n\nsystem design documentation\n\nsystem configuration settings and associated documentation\n\nlist of subjects and objects (i.e., users and resources) requiring enforcement of mandatory access control policies\n\nsystem audit records\n\nsystem security plan\n\nother relevant documents or records",
              "type": "EXAMINE"
            },
            {
              "id": "AC-03(03)-Interview",
              "method": "Organizational personnel with access enforcement responsibilities\n\nsystem/network administrators\n\norganizational personnel with information security responsibilities\n\nsystem developers",
              "type": "INTERVIEW"
            },
            {
              "id": "AC-03(03)-Test",
              "method": "Automated mechanisms implementing mandatory access control",
              "type": "TEST"
            }
          ],
          "assessmentObjectives": [
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(a)[01]",
              "objective": "[Assignment: organization-defined mandatory access control policy] is uniformly enforced across the covered subjects within the system;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(a)[02]",
              "objective": "[Assignment: organization-defined mandatory access control policy] is uniformly enforced across the covered objects within the system;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(b)(01)",
              "objective": "[Assignment: organization-defined mandatory access control policy] and [Assignment: organization-defined mandatory access control policy] specifying that a subject that has been granted access to information is constrained from passing the information to unauthorized subjects or objects are enforced;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(b)(02)",
              "objective": "[Assignment: organization-defined mandatory access control policy] and [Assignment: organization-defined mandatory access control policy] specifying that a subject that has been granted access to information is constrained from granting its privileges to other subjects are enforced;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(b)(03)",
              "objective": "[Assignment: organization-defined mandatory access control policy] and [Assignment: organization-defined mandatory access control policy] specifying that a subject that has been granted access to information is constrained from changing one of more security attributes (specified by the policy) on subjects, objects, the system, or system components are enforced;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(b)(04)",
              "objective": "[Assignment: organization-defined mandatory access control policy] and [Assignment: organization-defined mandatory access control policy] specifying that a subject that has been granted access to information is constrained from choosing the security attributes and attribute values (specified by the policy) to be associated with newly created or modified objects are enforced;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(b)(05)",
              "objective": "[Assignment: organization-defined mandatory access control policy] and [Assignment: organization-defined mandatory access control policy] specifying that a subject that has been granted access to information is constrained from changing the rules governing access control are enforced;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)(c)",
              "objective": "[Assignment: organization-defined mandatory access control policy] and [Assignment: organization-defined mandatory access control policy] specifying that [Assignment: organization-defined subjects] may explicitly be granted [Assignment: organization-defined privileges] such that they are not limited by any defined subset (or all) of the above constraints are enforced."
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)[01]",
              "objective": "[Assignment: organization-defined mandatory access control policy] is enforced over the set of covered subjects specified in the policy;"
            },
            {
              "class": "sp800-53a",
              "id": "AC-03(03)[02]",
              "objective": "[Assignment: organization-defined mandatory access control policy] is enforced over the set of covered objects specified in the policy;"
            }
          ],
          "category": "",
          "categoryID": "ac",
          "controlType": "PREVENTATIVE",
          "createdAt": "2025-03-04T20:17:49.973538-07:00",
          "createdBy": "01JNJ4X2H1YVJDY5TCAGJEXAXN",
          "description": "",
          "displayID": "CTL-UAGQIE",
          "id": "01JNJ4YFRN1DMBCYA3EE6PQGJC",
          "implementationGuidance": [
            {
              "guidance": [
                "Mandatory access control is a type of nondiscretionary access control. Mandatory access control policies constrain what actions subjects can take with information obtained from objects for which they have already been granted access. This prevents the subjects from passing the information to unauthorized subjects and objects. Mandatory access control policies constrain actions that subjects can take with respect to the propagation of access control privileges; that is, a subject with a privilege cannot pass that privilege to other subjects. The policy is uniformly enforced over all subjects and objects to which the system has control. Otherwise, the access control policy can be circumvented. This enforcement is provided by an implementation that meets the reference monitor concept as described in [AC-25](#ac-25) . The policy is bounded by the system (i.e., once the information is passed outside of the control of the system, additional means may be required to ensure that the constraints on the information remain in effect).\n\nThe trusted subjects described above are granted privileges consistent with the concept of least privilege (see [AC-6](#ac-6) ). Trusted subjects are only given the minimum privileges necessary for satisfying organizational mission/business needs relative to the above policy. The control is most applicable when there is a mandate that establishes a policy regarding access to controlled unclassified information or classified information and some users of the system are not authorized access to all such information resident in the system. Mandatory access control can operate in conjunction with discretionary access control as described in [AC-3(4)](#ac-3.4) . A subject constrained in its operation by mandatory access control policies can still operate under the less rigorous constraints of AC-3(4), but mandatory access control policies take precedence over the less rigorous constraints of AC-3(4). For example, while a mandatory access control policy imposes a constraint that prevents a subject from passing information to another subject operating at a different impact or classification level, AC-3(4) permits the subject to pass the information to any other subject with the same impact or classification level as the subject. Examples of mandatory access control policies include the Bell-LaPadula policy to protect confidentiality of information and the Biba policy to protect the integrity of information."
              ],
              "referenceId": "ac-3.3"
            }
          ],
          "ownerID": "",
          "refCode": "ac-3.3",
          "source": "FRAMEWORK",
          "standard": {
            "governingBody": "National Institute of Standards and Technology (NIST)",
            "id": "01JNJ4YBB9EESG2WY4P5E5MF1S",
            "name": "NIST 800-53 Rev 5: Security and Privacy Controls for Information Systems and Organizations",
            "shortName": "NIST 800-53"
          },
          "standardID": "01JNJ4YBB9EESG2WY4P5E5MF1S",
          "status": "",
          "subcategory": "",
          "tags": [
            "NIST 800-53",
            "openlane",
            "system-owned"
          ],
          "updatedAt": "2025-03-04T20:17:49.973527-07:00",
          "updatedBy": "01JNJ4X2H1YVJDY5TCAGJEXAXN"
        }
      }
    ]
  }
} 
```

</details>

# FGA - System Admin  
Adds a `system_admin` role in fga; this is required to created `systemOwned` standards. There is also a new corresponding task command that can be used to add the tuple to FGA:
  
```
task fga:add:admin USER_ID=<ULID_OF_USER> FGA_API_TOKEN=<token>
```

In dev, you can use the following to create a new user that is a system admin:

```
task cli:user:all:admin
```

- Adds a new `AllowMutationIfSystemAdmin` privacy rule, which is not currently in use
- Adds a `CheckIsSystemAdmin` function, used by the Standard hook

# Performance

- Adds a default complexity limit of `100` which can be configured using in the config
```
server:
    complexityLimit:
```
- Adds some eager loaded edges to standards; follow-up will be to add these (hopefully generated) across a lot of our resolvers to help with performance
- Fixes a bug with removal of parent tuple that was trying to remove the added field id

# Other Updates

- adds the `_self` tuple for api tokens; already exists for users; allows for the correct retrieval of objects created and retrieved with an api token
- removes unused custom types package

# Follow-ups
- Eager Loaded edges when pulling large amount of data
- Update interceptors for `listObjects` with parents to pull the _parent_ instead of all the child ids; we end up with 1000's of ids on things like controls when doing it on the child
- Tests for standards
- Standards policy for org-owned (vs. system-owned)
- Control Implementation + Control Mapping Permissions, since these are unused setting to deny all requests for now
- Add tests for control implementation and control mapping resolvers
- Ensure system-owned controls have a parent standard (add hook + test)
- Ensure controls associated with a system-owned parent standard cannot be edited (add test)
- Control Owner + Delegate Permissions
- Cli for standards
- auto reversion standards (and apply to other schemas like policies + procedures)


fga/Taskfile.yaml
```diff
@@ -1,5 +1,14 @@
 version: '3'
 
+vars:
+  # defaults to the dev token, to override pass `task <task_name> FGA_API_TOKEN=your_token`
+  FGA_API_TOKEN: '{{ .FGA_API_TOKEN | default "QKwHEmWX99RnFh28eSRJ3GWlfb2FQkL7toh1GJpzch1mMkVeMg" }}'
```
matoszz, line 5: is this an actual token? shouldn't be committed if so

internal/ent/hooks/authzmutationhelpers.go
```diff
@@ -260,7 +277,12 @@ func getAddedParentIDsFromEntMutation(ctx context.Context, m ent.Mutation, paren
 // if it is not set, it will return an empty string
 // this function does not ensure that the mutation was successful, it only extracts the id
 func getRemovedParentIDsFromEntMutation(ctx context.Context, m ent.Mutation, parentField string) ([]string, error) {
-	if v, ok := m.Field(parentField); ok {
+	if ok := m.FieldCleared(parentField); ok && m.Op().Is(ent.OpUpdateOne) {
```
matoszz: should this be both `OpUpdateOne` and `OpUpdate` ?

internal/ent/hooks/standard.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"github.com/theopenlane/entx"
+	"github.com/theopenlane/iam/fgax"
+
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/hook"
+)
+
+// HookStandardPublicAccess adds tuples for publicly available standards
+func HookStandardPublicAccess() ent.Hook {
```
matoszz: this function is a little confusing... I assumed there would be a check somewhere if the standard was flagged as public vs. private but it looks like it's attempting a create and delete tuple operation without that check?

matoszz: I'm still a little confused exactly how this is intended to work but I do see the check I was asking about called as a part of `AddOrDeleteStandardTuple`

internal/ent/privacy/rule/program.go
```diff
@@ -87,28 +84,38 @@ type ProgramParentMutation interface {
...
-func getParentIDFromEntMutation[T ent.Mutation](m generated.Mutation, parentType string) ([]string, error) {
+func getParentIDFromEntMutation[T ent.Mutation](m generated.Mutation, parentType string) []string {
 	switch parentType {
 	case ProgramParent:
 		return getProgramIDFromEntMutation[ProgramParentMutation](m)
 	case ControlParent:
-		return getControlIDFromEntMutation[ControlParentMutation](m)
+		id := getControlIDFromEntMutation[ControlParentMutation](m)
+		if id != "" {
+			return []string{id}
```
matoszz, line 114: why `[]string{}` vs. `string`?

internal/ent/schema/control.go
```diff
@@ -24,74 +26,80 @@ type Control struct {
...
+		edge.To("risks", Risk.Type),
+		edge.To("action_plans", ActionPlan.Type),
+
+		// policies and procedures are used to implement the control
+		edge.To("procedures", Procedure.Type),
+		edge.To("internal_policies", InternalPolicy.Type),
+
+		// owner is the user who is responsible for the control
+		edge.To("control_owner", User.Type).
+			Unique().
```
matoszz, line 86: It's likely we'll want to allow for the possibility of more than 1 owner (we could add this in the form of a "secondary owner" or similar but just thinking about things like... pending approvals and being able to set an "out of office" or time period tasks could be reassigned if not completed, etc.

internal/ent/schema/control.go
```diff
@@ -24,74 +26,80 @@ type Control struct {
...
 		edge.From("evidence", Evidence.Type).
 			Ref("controls"),
+
+		edge.To("implementation", ControlImplementation.Type).
+			Unique().
+			Annotations(entx.CascadeAnnotationField("Control")). // cascade delete the implementation when the control is deleted
+			Comment("the implementation of the control"),
+
+		edge.To("mapped_controls", MappedControl.Type).
+			Unique().
```
matoszz: should this be unique?

internal/ent/schema/control.go
```diff
@@ -24,74 +26,80 @@ type Control struct {
...
-		edge.To("tasks", Task.Type),
 		edge.From("programs", Program.Type).
 			Ref("controls"),
+
+		// evidence can be associated with the control
 		edge.From("evidence", Evidence.Type).
 			Ref("controls"),
+
+		edge.To("implementation", ControlImplementation.Type).
+			Unique().
```
matoszz: should this be unique?

internal/ent/schema/control.go
```diff
@@ -24,74 +26,80 @@ type Control struct {
...
 // Edges of the Control
 func (Control) Edges() []ent.Edge {
 	return []ent.Edge{
-		edge.To("procedures", Procedure.Type),
-		edge.To("subcontrols", Subcontrol.Type),
-		edge.To("control_objectives", ControlObjective.Type),
+		// parents of the control (standard, program)
 		edge.From("standard", Standard.Type).
+			Field("standard_id").
+			Unique().
```
matoszz, line 51: if a control could be re-used across standards we wouldn't want to make this a unique relationship right?

internal/ent/schema/mappedcontrol.go
```diff
@@ -0,0 +1,108 @@
...
+	ent.Schema
+}
+
+// Fields of the MappedControl
+func (MappedControl) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("control_id").
+			Comment("the id of the control being mapped").
+			Unique().
+			Immutable().
```
matoszz: we would not want to make this unique or immutable I think; we already assign an `ID` field to the table for uniqueness, and control ID's can change from the standards themselves (or they are collapsed together, or 2 existing ones in a new revision of a standard may get entirely new codes in the one they are collapsed into). Either that, or we need to be sure we can easily / elegantly retain versions of the controls and then create a new standard with a new revision and new control sets, it will just be confusing to the client if year over year they're expecting their controls from the prior period to be re-used when creating a new program / standard

internal/ent/schema/mappedcontrol.go
```diff
@@ -0,0 +1,108 @@
...
+			NotEmpty(),
+		field.String("mapped_control_id").
+			Unique().
+			Immutable().
```
matoszz: same comment with this one, although the reason is purely the re-use of code combinations

internal/ent/schema/standard.go
```diff
@@ -22,49 +27,63 @@ func (Standard) Fields() []ent.Field {
...
 			Optional().
 			Comment("type of the standard - security, privacy, etc."),
 		field.String("version").
 			Optional().
 			Comment("version of the standard"),
-		field.Text("purpose_and_scope").
-			Optional().
-			Comment("purpose and scope"),
-		field.Text("background").
+		field.String("revision").
```
matoszz, line 77: I believe there are semver packages or other versioning functions we could use to make this not a totally user-defined input; the "version" field above will (i'm certain) likely be inconsistently defined (e.g. NIST 800-53 rev5 vs. NIST 800-53A v5.1.1) but the revision should be programmatically created (or at least automatically incremented?)

internal/ent/schema/standard.go
```diff
@@ -22,49 +27,63 @@ func (Standard) Fields() []ent.Field {
...
+			Default(false).
+			Comment("indicates if the standard is freely distributable under a trial license, only for public standards"),
+		field.Bool("system_owned").
+			Optional().
+			Default(false).
+			Comment("indicates if the standard is owned by the the openlane system"),
 		field.String("standard_type").
 			Optional().
 			Comment("type of the standard - security, privacy, etc."),
 		field.String("version").
```
matoszz, line 74: we may want to consider using a yaml file or similar config to control the standard definitions in the system (or on startup the file is read in and the database records are created or similar) otherwise I worry about how we get the initial standards seeded into the system and then change / update them on an ongoing basis in a way we have an "approval" process for

internal/ent/schema/subcontrol.go
```diff
@@ -25,76 +27,59 @@ type Subcontrol struct {
...
-			Comment("date the subcontrol implementation was verified"),
-		field.JSON("details", map[string]any{}).
-			Optional().
-			Comment("json data details of the subcontrol"),
-		field.Text("example_evidence").
-			Comment("example evidence to provide for the control").
-			Optional(),
+			Comment("the unique reference code for the control"),
+		field.String("control_id").
+			Unique().
```
matoszz, line 39: wouldn't this need to be non-unique given subcontrols can map to many identical top level controls?

fga/Taskfile.yaml
golanglemonade: its the dev token use in local environments; its in the docker-compose already. It is not a live production token

internal/ent/hooks/standard.go
golanglemonade: Correct, this function adds (or removes) the tuple to fga if the standard is system owned + public = true. The logic is complex (hence the pretty long test associated with the `AddOrDeleteStandardTuple` function). It ensures that the tuples are appropriately created on creation and updates of the standard. This gives us the flexibility to add a new standard or standard version without making it available (is_public) immediately. The interceptor is what will actually filter if the standard is public/private to return the correct ones on query.

internal/ent/privacy/rule/program.go
golanglemonade: because `getProgramIDFromEntMutation` a few lines up can return multiple ids

internal/ent/schema/control.go
golanglemonade: There is another `delegate` field that was intended to accomplish this:

```
edge.To("delegate", User.Type).
	Unique().
	Comment("temporary delegate for the control, used for temporary control ownership"),
```

internal/ent/schema/control.go
golanglemonade: Can we talk more about this; I'm unsure why a control would be used in multiple standards (yes in multiple programs, but standards?)

internal/ent/schema/standard.go
golanglemonade: I don't think YAML is the way to go; this is a really complex relationships and seeded them is not idea. I have a script that takes our JSON and creates the objects; we just need to setup the version -> approval -> set to public steps. I think something as a simple as a buildkite pipeline could accomplish this.

internal/ent/schema/subcontrol.go
golanglemonade: Can they? I changed this to be unique because I didn't see any examples a subcontrol mapping to more than 1 parent control?

internal/ent/schema/control.go
golanglemonade: Can a control have more than one implementation?

internal/ent/hooks/standard.go
golanglemonade: updating name and added a more explicit comment

internal/ent/schema/control.go
golanglemonade: good call; I dont think so; removing it for now and will ensure this is correct on follow-up when I add tests.

internal/ent/schema/control.go
golanglemonade: ^ talked in slack; they can so removing unique - again need to do tests around this

internal/ent/schema/control.go
golanglemonade: Changing to groups instead of users

internal/ent/schema/control.go
golanglemonade: I think this falls into the mapping conversations we talked about earlier and not the uniqueness here. Keeping for now as its easier to back out this way and I will keep this in mind when looking at mapping controls.

internal/ent/schema/standard.go
golanglemonade: good callout; I'm going to add this to my follow-ups if you are ok with that

matoszz (approved):

matoszz (approved):

PR #522, opened 2025-03-14
Host example CSV files

matoszz:
- leverages the functional config parameter added for the bulkgen plugin and configures the output to internal/handlers/csv
- generated CSV's included in this PR
- adds route `/example/csv` and associated handler which reads from `embed.FS` file system and serves the file if the name provided in the filename input matches
- adds basic model and validation rule to require a filename in the JSON payload

example request + response: 

```
curl -L 'http://localhost:17608/example/csv' -H 'Content-Type: application/json' -d '{"filename": "actionplan"}'
Tags,Name,Description,Status,DueDate,Priority,Source,Details,RiskIDs,ControlIDs,UserIDs,ProgramIDs
example_tags,example_name,example_description,example_status,example_duedate,example_priority,example_source,example_details,example_riskids,example_controlids,example_userids,example_programids
```


internal/httpserve/handlers/examplecsv.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/rs/zerolog/log"
+	echo "github.com/theopenlane/echox"
+
+	"github.com/theopenlane/core/pkg/models"
+)
+
+//go:embed csv/*.csv
+var examplecsv embed.FS
+
+// ForgotPassword will send an forgot password email if the provided email exists
```
golanglemonade:
```suggestion
// ExampleCSV will return an example csv file that can be used for bulk uploads of the object
```

internal/httpserve/route/static.go
```diff
@@ -148,3 +148,26 @@ func registerAppleMerchantHandler(router *Router) (err error) {
...
+	method := http.MethodPost
+	name := "ExampleCSV"
+
+	route := echo.Route{
+		Name:        name,
+		Method:      method,
+		Path:        path,
+		Middlewares: authMW,
+		Handler: func(c echo.Context) error {
+			return router.Handler.ExampleCSV(c)
```
golanglemonade, line 167: You need to set the response header here; otherwise its using the default from the `mime` middleware:

```suggestion
			c.Response().Header().Set(httpsling.HeaderContentType, "text/csv")

			return router.Handler.ExampleCSV(c)
```

golanglemonade (commented):
two comments; otherwise LGTM

golanglemonade (dismissed):

golanglemonade (approved):

PR #558, opened 2025-03-26
feat: batch check requests instead of list objects

golanglemonade:
## Performance Improvement 🎉 
Task query in the UI before changes (1-2 seconds locally, way longer in production)
image

Query time after changes (less 100ms locally)
image

## Changes
- Upgrades iam to latest version to include `BatchAccessChecks`
- Adds a `count.tmpl` which creates a `CountIDs` function for reach receiver; this will do run an `IDs()` query and count the results that come back; this allows the use of the `filter` query to happen on the count after the query is run
- uses custom pagination template to use the `CountIDs` function instead of the default `Count` for the `totalCount` field
- Adds a new `FilterQueryResults` interceptor; the current function `AddPredicateIDs` is a traverse function that does a ListObjects request to FGA, takes all the IDs the user has access to and adds them to the query predicates. This had issues with performance due to the ListObjects fga API. The new function adds no filters (we still filter by org ID first for org owned objects) and then all results are filtered using the fga `BatchCheck` after the query is run and only returns the results the user has access to.
- I've left the `AddIDPredicate` on the org interceptor and membership checks; but everything else is using the new interceptor
- Removes org/owner setting in test seed models; this is almost always overwritten by the org in the context and was causing issues/confusing tests + not needed. 
- Fixes bug with `createGroupWithMembers` when there are no members provided


matoszz (approved):
WOW! Great work @golanglemonade !

PR #568, opened 2025-03-30
Echo log wrapper, context logger, middleware

matoszz:
- adds additional `logx` package functionality which wraps echo.Logger for more robust log output
- updates relevant instances of log lines to use zerolog.Ctx(ctx) to include relevant context information within the log
- creates new echo init function to set logger and middleware prior to server opts
- includes additional log output for user ID, org ID, and query duration
- expands `Caller()` information to include full path instead of just file name
- removes request metadata from `BodyDump` middleware and instead isolates `BodyDumo` to specifically request and response body, and moves request / response metadata to the logx middleware (since we would almost always want this information whereas the request and response body should only be output while testing)

Example of new log output:

Screenshot 2025-03-29 at 9 54 25 PM


internal/httpserve/route/router.go
```diff
@@ -212,6 +206,7 @@ func RegisterRoutes(router *Router) error {
 		registerWebhookHandler,
 		register2faHandler,
 		registerExampleCSVHandler,
+		registerPPROFroutes,
```
golanglemonade, line 209: We shouldn't expose our pprof routes on the same http server/port. This could potentially expose sensitive debugging information; and could also be used as a vector for DDoS attacks due to the cpu intensiveness of some of these functions.

internal/httpserve/server/server.go
```diff
@@ -17,6 +22,38 @@ type Server struct {
...
+
+	e.Use(middleware.RequestIDWithConfig(middleware.RequestIDConfig{
+		TargetHeader: "X-Request-ID",
+	}))
+
+	e.Use(logx.LoggingMiddleware(logx.Config{
+		Logger:          logger,
+		RequestIDHeader: "X-Request-ID",
+		RequestIDKey:    "request_id",
+		//		NestKey:         "REQUEST",
```
golanglemonade:
```suggestion
```

remove commented out code

internal/httpserve/server/server.go
```diff
@@ -17,6 +22,38 @@ type Server struct {
...
+		&output,
+		logx.WithLevel(echo_log.DEBUG),
+		logx.WithTimestamp(),
+		logx.WithCaller(),
+	)
+
+	e.Logger = logger
+
+	e.Use(middleware.RequestIDWithConfig(middleware.RequestIDConfig{
+		TargetHeader: "X-Request-ID",
```
golanglemonade, line 45: I'd recommend using the constant `echo.HeaderXRequestID` here and on line 50

internal/httpserve/server/stacktrace.go
```diff
@@ -0,0 +1,51 @@
...
+func CustomHTTPErrorHandler(c echo.Context, err error) {
+	code := http.StatusInternalServerError
+	if he, ok := err.(*echo.HTTPError); ok {
+		code = he.Code
+	}
+
+	// Create a custom error response
+	errorResponse := map[string]interface{}{
+		"error": err.Error(),
+		// "stackTrace": getStackTrace(),
```
golanglemonade:
```suggestion
```

internal/httpserve/server/stacktrace.go
```diff
@@ -0,0 +1,51 @@
...
+
+	log.Error().
+		Err(err).
+		Str("query", fmt.Sprintf("%v", c.QueryParams())).
+		Str("url", c.Request().URL.String()).
+		Msg("Error")
+
+	// Send response
+	if !c.Response().Committed {
+		if c.Request().Method == "HEAD" { // Issue #608
```
golanglemonade: What is issue 608?

golanglemonade: Use the http constant:
```suggestion
		if c.Request().Method == http.MethodHead { // Issue #608
```

internal/httpserve/server/stacktrace.go
```diff
@@ -0,0 +1,51 @@
...
+			err = c.NoContent(code)
+		} else {
+			err = c.JSON(code, errorResponse)
+		}
+	}
+
+	_ = err
```
golanglemonade, line 49: What is this doing?

internal/httpserve/serveropts/option.go
```diff
@@ -238,12 +237,12 @@ func WithMiddleware() ServerOption {
...
-			middleware.LoggerWithConfig(middleware.LoggerConfig{
-				Output: log.Logger.Hook(LevelNameHook{}),
-				Format: "remote_ip=${remote_ip}, method=${method}, uri=${uri}, status=${status}, session=${header:Set-Cookie}, host=${host}, referer=${referer}, user_agent=${user_agent}, route=${route}, path=${path}",
-			}),
+			//			middleware.RequestID(), // add request id
+			//			middleware.Recover(),   // recover server from any panic/fatal error gracefully
+			//			middleware.LoggerWithConfig(middleware.LoggerConfig{
+			//				Output: log.Logger.Hook(LevelNameHook{}),
+			//				Format: "remote_ip=${remote_ip}, method=${method}, uri=${uri}, status=${status}, session=${header:Set-Cookie}, host=${host}, referer=${referer}, user_agent=${user_agent}, route=${route}, path=${path}",
+			//			}),
```
golanglemonade: Remove commented out code
```suggestion
```

Just to confirm, the `Recover` middleware is added somewhere else now?

pkg/entitlements/customers.go
```diff
@@ -215,7 +215,16 @@ func (sc *StripeClient) GetCustomerByStripeID(ctx context.Context, customerID st
...
+	if id == "" || params == nil {
+		return nil, ErrCustomerIDRequired
+	}
+
+	cust, err := sc.Client.Customers.Update(id, params)
+	if err != nil {
+		return nil, err
+	}
+
+	return cust, nil
```
golanglemonade, line 227: nothing extra is happening on error so you can simplify this:

```suggestion
	return sc.Client.Customers.Update(id, params)
```

pkg/entitlements/subscriptions.go
```diff
@@ -111,6 +114,10 @@ func (sc *StripeClient) CreateTrialSubscription(cust *stripe.Customer) (*Subscri
 		return nil, err
 	}
 
+	zerolog.Ctx(ctx).UpdateContext(func(c zerolog.Context) zerolog.Context {
+		return c.Str("customer_id", cust.ID).Str("subscription_id", subs.ID)
+	})
+	log.Debug().Str("customer_id", cust.ID).Str("subscription_id", subs.ID).Msg("Created trial subscription")
 	log.Debug().Msgf("Created trial subscription with ID: %s", subs.ID)
```
golanglemonade: Should these also use the context logger instead?

pkg/logx/middleware.go
```diff
@@ -0,0 +1,200 @@
...
+				if config.HandleError {
+					c.Error(err)
+				}
+			}
+
+			if config.AfterNextSkipper(c) {
+				return err
+			}
+
+			logEvent(c, logger, config, start, err)
```
golanglemonade, line 119: this won't be called if something above errors; should this be towards the top with a `defer` instead so we always attempt to log the event?

pkg/logx/middleware.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			if err = next(c); err != nil {
+				if config.HandleError {
+					c.Error(err)
+				}
+			}
+
+			if config.AfterNextSkipper(c) {
+				return err
```
golanglemonade: Theres no `err` here so its always nil; we should just return that instead
```suggestion
				return nil
```

pkg/logx/middleware_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/labstack/gommon/log"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+	echo "github.com/theopenlane/echox"
+
+	"github.com/theopenlane/core/pkg/logx"
+)
+
+func TestMiddleware(t *testing.T) {
+	//	t.Run("should not trigger error handler when HandleError is false", func(t *testing.T) {
```
golanglemonade: Commented out test?

pkg/logx/logger_test.go
```diff
@@ -0,0 +1,257 @@
...
+	assert.Equal(
+		t,
+		`{"key":"test","level":"-","message":"foo"}
+`,
+		b.String(),
+	)
+}
+
+func TestLogger_SetPrefix(t *testing.T) {
+	//	b := &bytes.Buffer{}
```
golanglemonade: Commented out these here too

pkg/logx/context.go
```diff
@@ -0,0 +1,29 @@
...
+	if zerologger.GetLevel() == zerolog.NoLevel {
+		return ctx // Return the original context if zerologger is uninitialized
+	}
+
+	return zerologger.WithContext(ctx)
+}
+
+// Ctx returns a logger from the provided context; if no logger is found in the context, a new one is created
+func Ctx(ctx context.Context) *zerolog.Logger {
+	if ctx == nil {
```
golanglemonade, line 23: Should we be using this over `zerolog.Ctx` in the hooks/interceptors?

golanglemonade (commented):
this is great 🎉

internal/httpserve/route/router.go
matoszz: so what's your suggestion?

internal/httpserve/server/stacktrace.go
matoszz: its making sure that sending the error inside the error handler itself does not cause further errors (any error that occurs while sending the response is ignored; ensures the function does not panic or propagate further errors)

internal/httpserve/serveropts/option.go
matoszz: correct it's being added in the ConfigureEcho function via `e.Use(middleware.Recover())`

pkg/entitlements/customers.go
matoszz: not sure what you mean... the request itself returns `(*stripe.Customer, error)` and was actively erroring (hence why I changed it to check for the error); it was previously 404'ing due to missing ID parameter

pkg/logx/middleware.go
matoszz: there isn't anything above which errors and returns?

pkg/logx/context.go
matoszz: They both do effectively the same thing; seemed easier to start by switching over the contexts before re-wrapping

pkg/entitlements/customers.go
golanglemonade: right; I see the line above that checks for the missing parameter. What I'm suggesting is that you don't actually need the `if err` check line here because you don't handle the error path vs. success path differently (e.g. you aren't logging anything if there is an error, and aren't doing anything after that function call if it was successful. It can simply be:


```
func (sc *StripeClient) UpdateCustomer(id string, params *stripe.CustomerParams) (*stripe.Customer, error) {
	if id == "" || params == nil {
		return nil, ErrCustomerIDRequired
	}

	return sc.Client.Customers.Update(id, params)
```

pkg/logx/middleware.go
golanglemonade: if the request errors it does: 

```
	if err = next(c); err != nil {
		if config.HandleError {
			c.Error(err)
		}
	}
```

pkg/logx/context.go
golanglemonade: makes sense; I wasn't sure if we could end up with nil pointer errors without this wrapper was all.

pkg/logx/middleware.go
matoszz: that just sets the error, it doesn't return

golanglemonade (approved):

PR #572, opened 2025-03-31
Add filtering options to pat and apitoken queries

adelowo:
Added `last_used_at` and `expires_at`as sorting options to queries for `personalaccesstokens` and `apiTokens`


golanglemonade (approved):
🚀

PR #584, opened 2025-04-01
Make control status enums for better DX

adelowo:
`controlstatus` has been moved to an enum.

Tests also updated and command line usage too


internal/ent/schema/controlfields.go
```diff
@@ -25,10 +25,12 @@ var controlFields = []ent.Field{
...
 		Unique().
 		Comment("external auditor id of the control, can be used to map to external audit partner mappings"),
-	field.String("status").
+	field.Enum("status").
+		GoType(enums.ControlStatus("")).
 		Optional().
+		Default(enums.ControlStatusPreparing.String()).
 		Annotations(
-			entgql.OrderField("status"),
+			entgql.OrderField("STATUS"),
```
adelowo, line 33: fyi, all other enums use upper case for the ordering so this is a consistency change.

pkg/enums/controlstatus.go
```diff
@@ -0,0 +1,74 @@
...
+
+type ControlStatus string
+
+var (
+	// ControlStatusPreparing indicates the control is being prepared
+	ControlStatusPreparing ControlStatus = "PREPARING"
+	// ControlStatusNeedsApproval indicates the control needs to be approved before it is available
+	ControlStatusNeedsApproval ControlStatus = "NEEDS APPROVAL"
+	// ControlStatusChangesRequested indicates the control was rejected and needs some changes to be approved
+	ControlStatusChangesRequested ControlStatus = "CHANGES REQUESTED"
```
golanglemonade:
```suggestion
	ControlStatusNeedsApproval ControlStatus = "NEEDS_APPROVAL"
	// ControlStatusChangesRequested indicates the control was rejected and needs some changes to be approved
	ControlStatusChangesRequested ControlStatus = "CHANGES_REQUESTED"
```

Can you allow `null` as a valid value here as well - that should be the default on creation - only when controls are within an org would we have a status

internal/ent/schema/controlfields.go
```diff
@@ -25,10 +25,12 @@ var controlFields = []ent.Field{
 		Optional().
 		Unique().
 		Comment("external auditor id of the control, can be used to map to external audit partner mappings"),
-	field.String("status").
+	field.Enum("status").
+		GoType(enums.ControlStatus("")).
 		Optional().
+		Default(enums.ControlStatusPreparing.String()).
```
golanglemonade: lets have the default empty still; this field needs to be nillable as its not used when in a public standard vs. in org owned control

golanglemonade (commented):
nice job!

pkg/enums/controlstatus.go
```diff
@@ -12,13 +12,15 @@ var (
...
+	ControlStatusNeedsApproval ControlStatus = "NEEDS_APPROVAL"
 	// ControlStatusChangesRequested indicates the control was rejected and needs some changes to be approved
-	ControlStatusChangesRequested ControlStatus = "CHANGES REQUESTED"
+	ControlStatusChangesRequested ControlStatus = "CHANGES_REQUESTED"
 	// ControlStatusApproved indicates the control is approved
 	ControlStatusApproved ControlStatus = "APPROVED"
 	// ControlStatusArchived indicates the control is now archived
 	ControlStatusArchived ControlStatus = "ARCHIVED"
+	// ControlStatusNull indicates that this control status does not exists and is also not invalid
+	ControlStatusNull ControlStatus = "NULL"
```
adelowo, line 23: gqlgen does not allow `""`

golanglemonade (dismissed):

golanglemonade (approved):

PR #590, opened 2025-04-02
update interceptor and add additional urls

matoszz:
- updates org subscription interceptor to add data into the extended resolver fields for updating payment methods and performing account cancellation (TO-DO: receive + process webhook event for cancellations and payment method added)
- adds additional functions / updated functions to entitlements to fetch these URL's
- adds additional plan + feature + tier structures / mapping in preparation for attempting to automate the tuples needed for fga feature sets
- adds additional entitlements package functions for prices, products


internal/ent/interceptors/orgsubscription.go
```diff
@@ -20,20 +19,22 @@ func InterceptorSubscriptionURL() ent.Interceptor {
...
-			fields := graphutils.CheckForRequestedField(ctx, "subscriptionURL")
+			urlFields := []string{"subscriptionURL", "managePaymentMethods", "cancellation"}
 
-			// if the SubscriptionURL field wasn't queried, return the result as is
-			if !fields {
-				return v, nil
+			for _, field := range urlFields {
+				fields := graphutils.CheckForRequestedField(ctx, field)
+
+				if !fields {
```
golanglemonade: I think your return here is too early, you only want to return if _none_ of the fields are requested (I know the common use case would be all, but we still should technically allow for it). The logic won't change if there are 1 or all three requested, so I would just have a `hasField = false` and set it to true + break out of the loop if any of them are set, and check the value outside the loop before continuing

internal/ent/interceptors/orgsubscription.go
```diff
@@ -20,20 +19,22 @@ func InterceptorSubscriptionURL() ent.Interceptor {
...
+				}
 			}
 
 			// cast to the expected output format
 			orgSubResult, ok := v.([]*generated.OrgSubscription)
 			if ok {
 				for _, orgSub := range orgSubResult {
 					if err := setSubscriptionURL(orgSub, q); err != nil {
-						zerolog.Ctx(ctx).Warn().Err(err).Msg("failed to set subscription URL")
+						log.Warn().Err(err).Msg("failed to set subscription URL")
```
golanglemonade: these (and the new log lines) should all be `zerolog.Ctx(ctx)` not `log`

pkg/entitlements/portalsession.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &BillingPortalSession{
+		PaymentMethods: billingPortalSession.URL,
+	}, nil
+}
+
+// CreateBillingPortalPaymentMethods generates a session in stripe's billing portal which allows the customer to add / update payment methods
```
golanglemonade: bad comment, should be `CancellationBillingPortalSession .... `

pkg/entitlements/portalsession.go
```diff
@@ -0,0 +1,75 @@
+package entitlements
+
+import "github.com/stripe/stripe-go/v81"
+
+// CreateBillingPortalUpdateSession generates an update session in stripe's billing portal which displays the customers current subscription tier and allows them to upgrade or downgrade
+func (sc *StripeClient) CreateBillingPortalUpdateSession(subsID, custID string) (*BillingPortalSession, error) {
```
golanglemonade, line 6: these all return `BillingPortalSession` and you are calling all three functions in the interceptor individually. Would it be helpful to have a single function you can call to get all three back as well?

pkg/entitlements/prices.go
```diff
@@ -0,0 +1,64 @@
+package entitlements
+
+import "github.com/stripe/stripe-go/v81"
+
+// Create a price for a product:
```
golanglemonade:
```suggestion
// CreatePrice for a product in stripe
```

pkg/entitlements/prices.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if err := i.Err(); err != nil {
+		return nil, err
+	}
+
+	return prices, nil
+}
+
+// GetPrices retrieves all prices from stripe which are active and maps them into a []Price struct
```
golanglemonade:
```suggestion
// GetPricesMapped retrieves all prices from stripe which are active and maps them into a []Price struct
```

pkg/entitlements/products.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return sc.Client.Products.Update(productID, params)
+}
+
+// DeleteProduct deletes a product in Stripe
+func (sc *StripeClient) DeleteProduct(productID string) (*stripe.Product, error) {
+	return sc.Client.Products.Del(productID, nil)
+}
+
+// GetProducts retrieves all products from stripe which are active
```
golanglemonade:
```suggestion
// GetAllProductPricesMapped retrieves all products from stripe which are active
```

pkg/entitlements/portalsession.go
matoszz: I can add that in another PR; they are all setting different fields within the `BillingPortalSession` struct but I can certainly collapse them for convenience

golanglemonade (approved):
🚀

golanglemonade (approved):

PR #591, opened 2025-04-02
simplify policy validation by using internal context for tests

adelowo:
Added a new internal context that can be used in the tests. I have set this up as a pre policy so it will always 
run first and do a quick pass through if it is running in tests


internal/ent/privacy/rule/internal.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+
+	"github.com/theopenlane/core/internal/ent/generated/privacy"
+)
+
+type contextKey string
+
+const (
+	internalCtx contextKey = "internalCtx"
```
golanglemonade: you can just use a type, e..g:

```
 type internalAllowContextKey struct{}
```

This should be used in conjunction with our `contextx` packge: https://github.com/theopenlane/utils/blob/main/contextx/contextx.go

so you can just do 


```
 contextx.With(ctx, &internalAllowContextKey{token})
 ```
 
 and get it back wiith
 ```
 _, allow := contextx.From[internalAllowContextKey](ctx)
```

adelowo: resolved in https://github.com/theopenlane/core/pull/591/commits/a289c9395d305083c25b677cb64d1928568e91b2

golanglemonade (approved):
more tests still need to be updated, but merging so that we can use in new tests and not have to update them later

PR #605, opened 2025-04-08
tests for the notes/comments feature

adelowo:
This also includes adding file upload capabilities for task comments.


internal/graphapi/models_test.go
```diff
@@ -1019,6 +1029,34 @@ func (s *StandardBuilder) MustNew(ctx context.Context, t *testing.T) *ent.Standa
...
+	if n.Text == "" {
+		n.Text = gofakeit.HipsterSentence(10)
+	}
+
+	mutation := n.client.db.Note.Create().
+		SetText(n.Text)
+
+	if n.OwnerID != "" {
+		mutation.SetOwnerID(n.OwnerID)
+	}
```
golanglemonade: you shouldn't ever need this - we set the ownerID via the context of the user (in tests is the seed user you use the context for)
```suggestion
```

internal/graphapi/models_test.go
```diff
@@ -255,6 +255,16 @@ type StandardBuilder struct {
 	IsPublic  bool
 }
 
+type NoteBuilder struct {
+	client *client
+
+	// Fields
+	Text    string
+	OwnerID string
```
golanglemonade:
```suggestion
```

See other comment about why

internal/graphapi/models_test.go
adelowo: Ha thank you. I was testing something I can't recall anymore and let this there. I think it was the ownership thing before I updated the pga model

matoszz (approved):

PR #610, opened 2025-04-09
Fix trigger syntax

adelowo:
This fixes and updates the trigger syntax.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -305,7 +305,7 @@ steps:
                 - NAME=${APP_NAME}
       - trigger: "core-deploy"
         depends_on: "docker-pr-build"
-        label: ":rocket: Trigger core-deploy"
+        label: ":rocket: trigger core-deploy"
         if: build.tag != null
         build:
           env:
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -303,20 +303,20 @@ steps:
               push: true
               build-args:
                 - NAME=${APP_NAME}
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yaml
```diff
@@ -303,20 +303,20 @@ steps:
...
+        build:
+          env:
+            CORE_IMAGE: "${IMAGE_TAG}"
+          meta_data:
+            org: "theopenlane"
+            pipeline: "core-deploy"
+            branch: "main"
+            CORE_IMAGE: "${BUILDKITE_TAG}"
+            env:
+              CORE_IMAGE: "${BUILDKITE_TAG}"
```
golanglemonade: do you need both the meta data and the env var?

.buildkite/pipeline.yaml
```diff
@@ -303,20 +303,20 @@ steps:
               push: true
               build-args:
                 - NAME=${APP_NAME}
-      - label: ":rocket: Trigger core-deploy"
-        key: "trigger-deploy"
+      - trigger: "core-deploy"
+        depends_on: "docker-pr-build"
+        label: ":rocket: Trigger core-deploy"
```
golanglemonade: this is very much a nit, but can you lower case `trigger` - all the other steps labels are not capitalized 
```suggestion
        label: ":rocket: trigger core-deploy"
```

.buildkite/pipeline.yaml
adelowo: actually no. just the env is fine. Will update

golanglemonade (approved):

PR #613, opened 2025-04-10
Resend email to subscriber up until 5 times

adelowo:
I think long term it makes sense to move the maxAttempts to the config but i don't want to do that until we probably have to extend this logic elsewhere

### DIFF
--- a/internal/ent/hooks/subscriber.go
+++ b/internal/ent/hooks/subscriber.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/theopenlane/core/internal/ent/generated"
 	"github.com/theopenlane/core/internal/ent/generated/hook"
+	"github.com/theopenlane/core/internal/ent/generated/subscriber"
 )
 
 // HookSubscriberCreate runs on subscriber create mutations
@@ -31,9 +32,25 @@ func HookSubscriberCreate() ent.Hook {
 				return nil, err
 			}
 
-			retValue, err := next.Mutate(ctx, m)
-			if err != nil {
-				return nil, err
+			var retValue ent.Value
+
+			existingSubscriber, err := getSubscriber(ctx, m)
+
+			if existingSubscriber != nil && err == nil {
+				zerolog.Ctx(ctx).Info().Msg("user is already a subscriber. Resending email now")
+
+				retValue, err = updateSubscriber(ctx, m, existingSubscriber)
+				if err != nil {
+					zerolog.Ctx(ctx).Error().Err(err).Msg("unable to update email subscription")
+
+					return retValue, err
+				}
+			} else {
+				// create new subscription
+				retValue, err = next.Mutate(ctx, m)
+				if err != nil {
+					return retValue, err
+				}
 			}
 
 			if err := queueSubscriberEmail(ctx, m); err != nil {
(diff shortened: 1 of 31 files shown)


internal/graphapi/subscriber_test.go
```diff
@@ -149,50 +150,56 @@ func (suite *GraphTestSuite) TestMutationCreateSubscriber() {
...
-			email:   "c.stark@example.com",
-			client:  suite.client.api,
-			ctx:     testUser1.UserCtx,
-			wantErr: false,
+			name:             "happy path, duplicate subscriber but original was unsubscribed",
+			email:            "c.stark@example.com",
+			client:           suite.client.api,
+			ctx:              testUser1.UserCtx,
+			wantErr:          false,
+			expectedAttempts: 2,
```
golanglemonade: We should reset the send attempts when a user unsubscribes, I would expect this to be 1 and not increase on re-subscriber

internal/graphapi/subscriber_test.go
```diff
@@ -149,50 +150,56 @@ func (suite *GraphTestSuite) TestMutationCreateSubscriber() {
...
 		},
 		{
-			name:    "happy path, duplicate subscriber, case insensitive",
-			email:   "c.STARK@example.com",
-			client:  suite.client.api,
-			ctx:     testUser1.UserCtx,
-			wantErr: true,
```
golanglemonade, line 180: Can we add another test in here and ensure these pass:

1. If a subscriber is active (already verified) - we should return the error that the subscriber already exists
2. If a subscribe is not active/email is not verified we should resent.

internal/ent/hooks/subscriber.go
```diff
@@ -31,9 +32,25 @@ func HookSubscriberCreate() ent.Hook {
...
 
-			retValue, err := next.Mutate(ctx, m)
-			if err != nil {
-				return nil, err
+			var retValue ent.Value
+
+			existingSubscriber, err := getSubscriber(ctx, m)
+
+			if existingSubscriber != nil && err == nil {
+				zerolog.Ctx(ctx).Info().Msg("user is already a subscriber. Resending email now")
```
golanglemonade: we should _only_ resend the email if the subscriber is not verified. otherwise we want to return the subscriber already found error

internal/graphapi/subscriber_test.go
```diff
@@ -247,6 +299,24 @@ func (suite *GraphTestSuite) TestMutationCreateSubscriber() {
...
+
+			require.NoError(t, err)
+
+			if tc.setUnsubscribed {
+				require.Zero(t, sub.SendAttempts) // reset attempts count to zero
+				return
+			}
+
+			require.NoError(t, err)
+			require.Equal(t, tc.expectedAttempts, sub.SendAttempts)
```
golanglemonade: I think this check should be up at line 289 so you are checking as part of the request, not the cleanup/changes we are doing to test other things

internal/graphapi/subscriber_test.go
```diff
@@ -247,6 +299,24 @@ func (suite *GraphTestSuite) TestMutationCreateSubscriber() {
...
+				Only(ctx)
+
+			require.NoError(t, err)
+
+			if tc.setUnsubscribed {
+				require.Zero(t, sub.SendAttempts) // reset attempts count to zero
+				return
+			}
+
+			require.NoError(t, err)
```
golanglemonade:
```suggestion
```

you already check the last error on line 311

internal/graphapi/subscriber_test.go
```diff
@@ -157,7 +158,75 @@ func (suite *GraphTestSuite) TestMutationCreateSubscriber() {
...
+			expectedAttempts: 1,
+		},
+		{
+			name:             "happy path, new subscriber using personal access token",
+			email:            "a.stark@example.com",
+			ownerID:          testUser1.OrganizationID,
+			client:           suite.client.apiWithPAT,
+			ctx:              context.Background(),
+			wantErr:          false,
+			expectedAttempts: 1,
```
golanglemonade: `expectedAttempts` is never checked in the tests themselves

internal/graphapi/subscriber_test.go
golanglemonade: actually, why is this being queried again? I think instead the `sendAttempts` should be added to the graphql query so it can be checked directly in the response

adelowo: I put it here because I don't want to have to duplicate the logic to fetch the subscriber from the db. 

As I want to verify the send attempts of both unsubscribing logic too

internal/graphapi/subscriber_test.go
adelowo: Copy/pasta :) I'd remove them in this token tests as we already test them in the other one.

golanglemonade (approved):

golanglemonade (approved):

PR #616, opened 2025-04-11
Feat prom

matoszz:
- adds new metrics middleware / package
- adds global ent hooks to emit metrics
- adds basic counter stubs for login / queue / worker that will be used later


internal/httpserve/server/server.go
```diff
@@ -145,6 +146,13 @@ func (s *Server) StartEchoServer(ctx context.Context) error {
 
 		return sc.StartTLS(s.Router.Echo, s.config.Settings.Server.TLS.CertFile, s.config.Settings.Server.TLS.CertKey)
 	}
+
+	newMetrics := metrics.New(17609)
```
golanglemonade: this should be configurable just like the main server port

internal/ent/hooks/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	return promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "ent_operation_duration_seconds",
+			Help: "Time in seconds per operation",
+		},
+		entLabels,
+	)
+}
+
+// MetricsHook inits the collectors with count total at beginning, error on mutation error and a duration after the mutation
```
golanglemonade, line 59: What about query route metrics? should we have an interceptor for that as well?

internal/httpserve/server/server.go
```diff
@@ -145,6 +146,13 @@ func (s *Server) StartEchoServer(ctx context.Context) error {
...
+	go func() {
+		if err := newMetrics.Start(ctx); err != nil {
+			log.Error().Err(err).Msg("metrics server failed to start")
```
golanglemonade, line 153: since you aren't doing anything with the error here; if the metrics server fails to start; the main server will still start - which might mean missing this error. Should this be a `log.Fatal()` instead?

internal/ent/hooks/metrics.go
matoszz: yea i just didn't write it yet ;)

internal/httpserve/server/server.go
matoszz: nah i wouldn't think so

internal/httpserve/server/server.go
matoszz: does a thing we never change really need a bunch of config wrapping?

golanglemonade: for consistency, yeah. I've personally changed these before to be able to run two servers locally at once.

golanglemonade (dismissed):

adelowo (approved):

PR #617, opened 2025-04-11
Make go:lint task much smarter

adelowo:
Now `task go:lint CHANGED=true` will run linting against all files that have changed against the main branch regardless of if they are committed or not

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -80,12 +80,27 @@ tasks:
       CHANGED: '{{.CHANGED | default "false"}}'
     cmds:
       - |
+        if [ "$(git rev-parse --abbrev-ref HEAD)" = "main" ]; then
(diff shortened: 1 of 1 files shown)


Taskfile.yaml
```diff
@@ -80,12 +80,27 @@ tasks:
       CHANGED: '{{.CHANGED | default "false"}}'
     cmds:
       - |
+        if [ "$(git rev-parse --abbrev-ref HEAD)" = "main" ]; then
+          echo "⚠️  Skipping lint: you are on the 'main' branch."
```
golanglemonade: instead of skipping completely; can we just add the warning and run the full linter?

adelowo: fixed

golanglemonade (approved):

matoszz (approved):

PR #626, opened 2025-04-14
Send welcome email on sign up

adelowo:
Welcome emails are now sent out.

This PR also fixes a bug with sending out invitations. When you invite a user with an api token, we get this 

```
11:26PM ERR internal/ent/hooks/invite.go:78 | error sending email to user error=generated: user not found org_id=01JRV552T6G0XJ97YRYP6PJWCD request_id=aFGTpkuGIFVKOShiMgXSsJHAAoNyhFaH user_id=01JRV585ZPMVGPFGTH1SZ60T9C

```

This handles the case for creating invitations by api tokens


internal/httpserve/handlers/verifyemail.go
```diff
@@ -97,9 +98,20 @@ func (h *Handler) VerifyEmail(ctx echo.Context) error {
...
+			log.Err(err).Msg("unable to get user id from context")
+
+			return h.BadRequest(ctx, err)
+		}
+
 		if err := h.setEmailConfirmed(userCtx, entUser); err != nil {
 			return h.BadRequest(ctx, err)
 		}
+
+		if err := h.sendWelcomeEmail(userCtx, user, orgID); err != nil {
```
golanglemonade: This should also be called when a user is created via the oauth register flow; in that case there is no verify email so this path isn't called. After here I believe: https://github.com/theopenlane/core/blob/main/internal/httpserve/handlers/oauth_register.go#L41

internal/ent/hooks/invite.go
```diff
@@ -256,6 +256,24 @@ func createInviteToSend(ctx context.Context, m *generated.InviteMutation) error
...
+	case auth.UserSubjectType:
+	// fallthrough
+
+	case auth.ServiceSubjectType:
+		token, err := m.Client().APIToken.Get(ctx, reqID)
+		if err != nil {
+			return err
+		}
+
+		reqID = token.CreatedBy
```
golanglemonade: hm, API tokens are owned by an organization so it not be that user that used it to created the invite (that user is possibly no longer even in the system). What if instead, if its an API token we set the Inviter Name to the same as the Organization Name?

golanglemonade (approved):

PR #636, opened 2025-04-17
Passkeys registration and login

adelowo:
I have added an extra api endpoint `v1/login/methods`, this will essentially take an email address,
then return which methods are available for the user to use.

```json

{
  "success": true,
  "methods": [
    "CREDENTIALS",
    "WEBAUTHN"
  ]
}

```

Also made a few changes to webauthn implementation by adding the RegistrationRelyingPartyName ( which is required per spec ) and some sql queries.

Usersetting hook has also been updated to delete the passkey if the user turns off passkeys from the frontend 

UI changes => https://github.com/theopenlane/openlane-ui/pull/302


internal/ent/hooks/usersettings.go
```diff
@@ -28,6 +29,21 @@ func HookUserSetting() ent.Hook {
...
+					// if webauthn is disabled, clean up the passkey we stored previously
+					if allowed, _ := m.IsWebauthnAllowed(); !allowed {
+						_, err := m.Client().Webauthn.Delete().
+							Where(webauthn.OwnerID(userID)).
+							Exec(ctx)
+						if err != nil {
+							return nil, err
+						}
+					}
+				}
```
golanglemonade, line 47: this _shouldnt_ ever happen here since its an update to user settings, but we are currently silently ignoring the error if it did. 
```suggestion
				if err != nil {
					return err
				}
				
				if userID != "" {
					// if webauthn is disabled, clean up the passkey we stored previously
					if allowed, _ := m.IsWebauthnAllowed(); !allowed {
						_, err := m.Client().Webauthn.Delete().
							Where(webauthn.OwnerID(userID)).
							Exec(ctx)
						if err != nil {
							return nil, err
						}
					}
				}
```

internal/httpserve/handlers/auth_type.go
```diff
@@ -0,0 +1,76 @@
...
+	availableAuthReq := openapi3.NewOperation()
+	availableAuthReq.Description = `AvailableAuthType is oriented towards human users who have their emails but 
+	would like to get authenticated via a self selected option of possible auth choices 
+	( currently password or passkey )`
+	availableAuthReq.Tags = []string{"authentication"}
+	availableAuthReq.OperationID = "AvailableAuthTypeHandler"
+	availableAuthReq.Security = BasicSecurity()
+
+	h.AddRequestBody("AvailableAuthTypeLoginRequest", models.ExampleAvailableAuthTypeRequest, availableAuthReq)
+	h.AddResponse("LoginReply", "success", models.ExampleAvailableAuthTypeSuccessResponse, availableAuthReq, http.StatusOK)
```
golanglemonade:
```suggestion
	h.AddResponse("AuthTypeReply", "success", models.ExampleAvailableAuthTypeSuccessResponse, availableAuthReq, http.StatusOK)
```

internal/httpserve/handlers/ent.go
```diff
@@ -183,21 +183,25 @@ func (h *Handler) getUserByEmail(ctx context.Context, email string, authProvider
...
 		Only(ctx)
 	if err != nil {
 		log.Error().Err(err).Msg("error obtaining user from id")
 
-		return nil, err
+		return nil, ctx, err
 	}
 
+	// set user in the viewer context for the rest of the request
+	ctx = setAuthenticatedContext(ctx, user)
```
golanglemonade, line 199: feels a bit weird this is being called + returned in a function that is used for wrappers around queries. This should probably be called from the handler itself instead.

golanglemonade (commented):
nice work! just a few comments

internal/ent/hooks/usersettings.go
adelowo: I think the issue here might be for when we need to create a new user ( say on sign up , from cli ), we do not have a user in the context in both instances. 

that was why but I could just log the error regardless.

I will try out that and see if it can be worked around

golanglemonade: ahhh gotcha. thats actually a good point - just add a comment about why we are ignoring it then

internal/httpserve/handlers/ent.go
```diff
@@ -183,21 +183,25 @@ func (h *Handler) getUserByEmail(ctx context.Context, email string, authProvider
...
+
 	// Add webauthn to the response
 	user.Edges.Webauthn = user.QueryWebauthn().AllX(ctx)
```
golanglemonade: I know you didn't write this, but this shouldn't be `AllX`, that will panic. We should do:
```
user.Edges.Webauthn , err = user.QueryWebauthn().All(ctx)
if err != nil {
     return err
}
```

PR #639, opened 2025-04-18
 fix date input

adelowo:
Allow for multiple custom inputs for date time when uploading csv for tasks.

I have added a new DateTime scalar type that understands both formats `2025-04-28T04:00:00Z` and `2025-04-18`.
We can extend to support more if needed

> we can reuse in other places


cmd/cli/cmd/task/create.go
```diff
@@ -60,7 +60,7 @@ func createValidation() (input openlaneclient.CreateTaskInput, err error) {
 
 	due := cmd.Config.Duration("due")
```
golanglemonade: We should change the input type to a string instead of a duration, e.g. 

```
	createCmd.Flags().String("due", 0, "time until due date of the task")
```

internal/graphapi/generate/.gqlgen.yml
```diff
@@ -108,3 +108,5 @@ models:
   VersionBump:
     model:
       - github.com/theopenlane/core/pkg/models.VersionBump
+  DateTime:
```
golanglemonade, line 111: Does this need to be added to the `gqglenc.yaml` config too?

golanglemonade (commented):
🎉

internal/graphapi/generate/.gqlgen.yml
adelowo: fixed in https://github.com/theopenlane/core/pull/639/commits/a30b6601cf3b57a8ab26e2e13ae0be8bf09b269a

golanglemonade (dismissed):

PR #645, opened 2025-04-19
stripe webhook updates; additional ent and webhook metrics

matoszz:
- changes the current webhook handler behavior from returning a code in the error range (4xx 5xx) when an event type we dont support is received to a success response code - we are discarding the event regardless, but when you return 4xx/5xx to stripe they attempt to resend every event up to 5+ times which is a lot of extra overhead on us (and results in getting emails from them, etc.)
- updates / corrects several of the field evaluations in the `syncOrgSubscriptionWithStripe` function to perform updates if the fields are _not_ alike rather than if they are (`if !slices.Equal` vs. `if slices.Equal` 🤦 )
- fixes a bug in `InterceptorSubscriptionURL` causing none of the URL's to be populated and populates different log error text to make it more obvious which individual URL may have been the culprit

### DIFF
--- a/internal/httpserve/authmanager/authmanager.go
+++ b/internal/httpserve/authmanager/authmanager.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 
 	"github.com/theopenlane/iam/auth"
@@ -101,11 +102,22 @@ func (a *Client) GenerateOauthAuthSession(ctx context.Context, w http.ResponseWr
 	return auth, nil
 }
 
-// checkActiveSubscription checks if the user has an active subscription for the organization
+var (
+	// ErrOrganizationIDRequired is the error message when the organization ID is required
+	ErrOrganizationIDRequired = errors.New("organization ID is required")
+)
+
+// checkActiveSubscription checks if the organization has an active subscription
 func (a *Client) checkActiveSubscription(ctx context.Context, orgID string) (active bool, err error) {
 	// if the entitlement manager is disabled, we can skip the check
 	if a.db.EntitlementManager == nil {
 		return true, nil
+	}
+
+	if orgID == "" {
+		log.Debug().Msg("organization ID is required to check for active subscription")
+
+		return false, ErrOrganizationIDRequired
 	}
 
 	if _, ok := contextx.From[auth.OrganizationCreationContextKey](ctx); ok {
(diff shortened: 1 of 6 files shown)


internal/httpserve/authmanager/authmanager.go
```diff
@@ -116,8 +122,10 @@ func (a *Client) checkActiveSubscription(ctx context.Context, orgID string) (act
 	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
 	allowCtx = contextx.With(allowCtx, auth.OrgSubscriptionContextKey{})
 
-	subscription, err := a.db.OrgSubscription.Query().Select("active").Where(orgsubscription.OwnerID(orgID)).Only(allowCtx)
```
golanglemonade, line 119: Any reason we need all the fields now instead of just `active`? I dont see changes below that would require this

internal/httpserve/handlers/webhook.go
```diff
@@ -327,15 +371,15 @@ func syncOrgSubscriptionWithStripe(ctx context.Context, subscription *stripe.Sub
...
 	}
 
-	if stripeOrgSubscription.ExpiresAt != nil && orgSubscription.ExpiresAt != stripeOrgSubscription.ExpiresAt {
-		mutation.SetExpiresAt(*stripeOrgSubscription.ExpiresAt)
+	if stripeOrgSubscription.ExpiresAt != nil && orgSubscription.TrialExpiresAt != stripeOrgSubscription.TrialExpiresAt {
+		mutation.SetTrialExpiresAt(*stripeOrgSubscription.ExpiresAt)
 
 		changed = true
 
 		log.Debug().Str("subscription_id", orgSubscription.ID).Str("expires_at", stripeOrgSubscription.ExpiresAt.String()).Msg("subscription expiration changed")
```
golanglemonade: Should this also be `TrialExpiresAt`? 

Also - for my information - why is this being changed?

matoszz: There are 2 fields on the orgsubscription schema, the `TrialExpiresAt` and `ExpiresAt` - the `ExpiresAt` is only set / should be set when the subscription is cancelled and there's a prorated number of days remaining within the billing cycle; the `TrialExpiresAt` has a default set on creation, and the trial could also be extended by us manually (in which case we'd push this date out but keep the `active` flag). In the future our subscription check / evaluation logic will need to get more complex to support a la carte feature-based billing where each feature might have a trial period (or maybe not, not 100% sure) so I wanted to be sure we were syncing the dates in the correct fields given 1 of the expiresAt is specific to cancellations, the other is specific to trials

golanglemonade: thanks thats helpful! - should the comment be updated to match the check though?

matoszz: and yes the fields should match lol - pushed an update

golanglemonade (approved):

golanglemonade (approved):

PR #667, opened 2025-04-28
Allow for users to set up multiple passkeys

adelowo:
- add webauthn resolvers for frontend
- exposed `webauthns` edge to frontend so they can list devices that are set up already
- add passkey tests
- expose aaguid model so we can display the device that generated the passkey on the frontend
- resolvers disabled and only the delete resolver is valid

```
query User($userId: ID!) {
  user(id: $userId) {
    webauthns {
      edges {
        node {
          id
        }
      }
    }
  }
}


  query GetPasskeys {
    webauthns {
      edges {
        node {
          backupEligible
          backupState
          id
          aaguid
        }
      }
    }
  }

mutation DeleteWebauthn($deleteWebauthnId: ID!) {
  deleteWebauthn(id: $deleteWebauthnId) {
    deletedID
  }
}

```


internal/ent/hooks/webauthn.go
```diff
@@ -0,0 +1,54 @@
...
+				if err != nil {
+					zerolog.Ctx(ctx).Error().Err(err).Msg("could not disable webauthn from user")
+
+					return nil, err
+				}
+			}
+
+			return next.Mutate(ctx, m)
+		})
+	}, ent.OpDeleteOne|ent.OpDelete)
```
golanglemonade, line 53: Does this work? We do soft deletes and usually when the hook is hit its an Update/UpdateOne operation, not a delete anymore

adelowo: Passkeys are not soft deleted actually. They are hard deleted at the moment

internal/ent/hooks/webauthn.go
```diff
@@ -0,0 +1,54 @@
...
+
+			passkey, err := m.Client().Webauthn.Get(ctx, deletedID)
+			if err != nil {
+				zerolog.Ctx(ctx).Error().Err(err).Msg("could not fetch webauthn to delete")
+
+				return nil, err
+			}
+
+			count, err := m.Client().Webauthn.Query().
+				Where(webauthn.OwnerID(passkey.OwnerID)).
```
golanglemonade, line 31: we can just grab the userID from the auth context using `GetSubjectIDFromContext` and not pull the deleted, to then do another query of count

internal/ent/hooks/webauthn.go
```diff
@@ -0,0 +1,54 @@
...
+				zerolog.Ctx(ctx).Error().Err(err).Msg("could not get count of webauthns")
+
+				return nil, err
+			}
+
+			// 1 since this tx is not complete yet
+			if count == 1 {
+				_, err := m.Client().UserSetting.Update().Where(usersetting.UserID(passkey.OwnerID)).
+					SetIsWebauthnAllowed(false).
+					Save(ctx)
```
golanglemonade, line 43: use `Exec` instead of `Save` if you don't need the result

internal/ent/hooks/webauthn.go
golanglemonade: You actually might not even need this at all; webauthn is userOwned, so we should have an interceptor that filters on user id already:

https://github.com/theopenlane/core/blob/main/internal/ent/schema/webauthn.go#L115

internal/ent/hooks/webauthn.go
```diff
@@ -0,0 +1,54 @@
...
+				if err != nil {
+					zerolog.Ctx(ctx).Error().Err(err).Msg("could not disable webauthn from user")
+
+					return nil, err
+				}
+			}
+
+			return next.Mutate(ctx, m)
```
golanglemonade, line 51: Instead of doing `count == 1`; and we run these queries _after_ calling `next.Mutate(ctx,  m)` to make sure the passkey is delete before this is called?

internal/ent/schema/webauthn.go
```diff
@@ -81,11 +119,25 @@ func (w Webauthn) Mixin() []ent.Mixin {
 // Annotations of the Webauthn
 func (Webauthn) Annotations() []schema.Annotation {
 	return []schema.Annotation{
-		entgql.Skip(entgql.SkipAll),
-		entx.SchemaGenSkip(true),
-		entx.QueryGenSkip(true),
+		entgql.RelayConnection(),
+		entx.SchemaSearchable(true),
```
golanglemonade, line 123: I dont think we need this schema searchable

internal/graphapi/query/webauthn.graphql
```diff
@@ -0,0 +1,125 @@
...
+      backupState
+      createdAt
+      createdBy
+      id
+      tags
+      updatedAt
+      updatedBy
+    }
+  }
+}
```
golanglemonade, line 30: Let's remove the bulk endpoints, we dont need to allow bulk creation of passkeys

internal/ent/hooks/webauthn.go
adelowo: It's in a transaction so I supposed it should be fine but I can move it around

internal/ent/hooks/webauthn.go
adelowo: Will make this update. Lesser queries. Thank!!

internal/ent/hooks/webauthn.go
golanglemonade: Ohhh nevermind; webauthn doesn't have the soft delete mixin - this can be ignored.

internal/graphapi/query/webauthn.graphql
```diff
@@ -0,0 +1,125 @@
...
```
golanglemonade, line 125: Do we need an update resolver?

adelowo: They are all disabled actually in the implementation.

The only resolver that works is the deletion

pkg/models/aaguid.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/google/uuid"
+)
+
+var (
+	ErrUnsupportedDataType = errors.New("unsupported aaguid format")
+)
+
+// AAGUID is a custom type representing an authenticator attestation uuid.
+type AAGUID []byte
```
golanglemonade, line 17: is this specific to `aaguid` or can it be generically a `uuid`?

internal/ent/hooks/webauthn.go
```diff
@@ -0,0 +1,54 @@
...
+	"entgo.io/ent"
+	"github.com/rs/zerolog"
+
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/hook"
+	"github.com/theopenlane/core/internal/ent/generated/usersetting"
+	"github.com/theopenlane/core/internal/ent/generated/webauthn"
+)
+
+func HookWebauthDelete() ent.Hook {
```
golanglemonade, line 15: missing comment for exported function

internal/ent/schema/webauthn.go
```diff
@@ -81,11 +119,25 @@ func (w Webauthn) Mixin() []ent.Mixin {
...
+		entx.FieldAdminSearchable(true),
+
+		entgql.QueryField(),
+		entgql.Mutations(entgql.MutationCreate(), entgql.MutationUpdate()),
```
golanglemonade, line 127: don't think we need update resolver:

```
entgql.Mutations(entgql.MutationCreate()),
`

pkg/models/aaguid.go
adelowo: specific to the aaguuid. I didn't want to change the db structure especially as we already have passkeys being used in cloud environment

internal/graphapi/query/webauthn.graphql
golanglemonade: yep; so we should delete it - they all get autogenerated but if you remove it from here and rerun the resolvers wont get generated

golanglemonade: If you remove anything in. here that you have as just returning ErrPermissionDenied in the resolver (I always miss that file because it gets hidden by github by default) we wont have to have the resolvers at all

internal/ent/schema/webauthn.go
golanglemonade: If the create resolver is also hidden; you can actually remove this whole line

PR #670, opened 2025-04-28
prevent same device from multiple registration

adelowo:
- Fixed all reviews from #667 
- Prevent same device getting registered multiple times


internal/ent/hooks/webauthn.go
```diff
@@ -6,29 +6,26 @@ import (
...
+
 	"github.com/theopenlane/core/internal/ent/generated"
 	"github.com/theopenlane/core/internal/ent/generated/hook"
 	"github.com/theopenlane/core/internal/ent/generated/usersetting"
-	"github.com/theopenlane/core/internal/ent/generated/webauthn"
 )
 
+// HookWebauthDelete runs on passkey delete mutations to ensure
+// that we update the user's settings if needed
 func HookWebauthDelete() ent.Hook {
```
golanglemonade:
```suggestion
// HookWebauthnDelete runs on passkey delete mutations to ensure
// that we update the user's settings if needed
func HookWebauthnDelete() ent.Hook {
```

internal/ent/hooks/webauthn.go
```diff
@@ -6,29 +6,26 @@ import (
...
 	return hook.On(func(next ent.Mutator) ent.Mutator {
 		return hook.WebauthnFunc(func(ctx context.Context, m *generated.WebauthnMutation) (generated.Value, error) {
-			deletedID, ok := m.ID()
```
golanglemonade: You should be able to:
1. remove `userID` at all; the user owned mixin has an [interceptor](https://github.com/theopenlane/core/blob/main/internal/ent/schema/mixin_userowned.go#L220) + [hook](https://github.com/theopenlane/core/blob/main/internal/ent/schema/mixin_userowned.go#L183) that filters all requests on user id:
2. Run this after the deletion mutation:

Example code:
```go
                        retVal, err := next.Mutate(ctx, m)
			if err != nil {
				return nil, err
			}

			// get the count of webauthns for the user
			count, err := m.Client().Webauthn.Query().
				Count(ctx)
			if err != nil {
				zerolog.Ctx(ctx).Error().Err(err).Msg("could not get count of webauthns")

				return nil, err
			}

                        // if the count is 0, we need to disable webauthn for the user
			if count == 0 {
				if err = m.Client().UserSetting.Update().
					SetIsWebauthnAllowed(false).
					Exec(ctx); err != nil {
					zerolog.Ctx(ctx).Error().Err(err).Msg("could not disable webauthn from user")

					return nil, err
				}
			}

			return retVal, nil
```

internal/graphapi/webauthn_test.go
```diff
@@ -9,64 +9,6 @@ import (
 	"github.com/theopenlane/core/pkg/openlaneclient"
 )
 
-func (suite *GraphTestSuite) TestQueryPasskeys() {
```
golanglemonade: I dont think you want to delete this test - you still are doing a list query for webauthns, you want `GetAllWebauthns` instead

internal/graphapi/webauthn_test.go
```diff
@@ -120,10 +62,6 @@ func (suite *GraphTestSuite) TestMutationDeletePasskeys() {
 
 			require.NoError(t, err)
 			require.NotNil(t, resp)
-
-			passkeys, err := tc.client.GetWebauthns(tc.ctx, nil, nil, &openlaneclient.WebauthnWhereInput{})
-			require.NoError(t, err)
-			require.Len(t, passkeys.Webauthns.Edges, tc.expectedCount)
 		})
```
golanglemonade: Same here - shouldn't be deleted, should be switched to `GetAllWebauthns`  e.g. 

```
passkeys, err := tc.client.GetAllWebauthns(tc.ctx)
```

golanglemonade (commented):
nice work!

internal/graphapi/webauthn_test.go
adelowo: added back 💯

golanglemonade (approved):

PR #676, opened 2025-04-29
Handle org deletion and subscription cancellation / deletion scenarios

matoszz:
- upgrades stripe API version and sdk version to v82
- changes `mock.CallRaw` to accommodate new version and removes deprecated fields
- updates ent soiree event pool to include the stats collector
- adds new event listener for org delete mutations and associated handler which will deactivate any still active subscriptions in stripe based on a delete operation triggered within Openlane
- updates data sync parameters for stripe webhooks to properly change records in openlane when subscription cancellation or deletions occur
- minor version updates


pkg/entitlements/subscriptions.go
```diff
@@ -227,3 +233,54 @@ func (sc *StripeClient) MapStripeSubscription(subs *stripe.Subscription) *Subscr
...
+	case stripe.SubscriptionStatusCanceled,
+		stripe.SubscriptionStatusIncompleteExpired,
+		stripe.SubscriptionStatusUnpaid,
+		stripe.SubscriptionStatusPaused:
+		return false
+	default:
+		// If an unknown status is encountered, default to inactive for safety
+		return false
+	}
+}
```
adelowo: We can probably simplify this. The linter can be very annoying a lot of times

adelowo (dismissed):

internal/ent/hooks/event.go
```diff
@@ -220,6 +228,38 @@ func RegisterListeners(e *Eventer) error {
...
+	return nil
+}
+
+// handleOrganizationDelete handles the deletion of an organization and deletes the customer in Stripe
+func handleOrganizationDelete(event soiree.Event) error {
+	client := event.Client().(*entgen.Client)
+	entMgr := client.EntitlementManager
+
+	if entMgr == nil {
+		log.Debug().Msg("EntitlementManager not found on client, skipping customer deletion")
```
golanglemonade: Use the context logger, e.g. 
```suggestion
		zerolog.Ctx(event.Context()).Debug().Msg("EntitlementManager not found on client, skipping customer deletion")
```

internal/ent/hooks/event.go
```diff
@@ -220,6 +228,38 @@ func RegisterListeners(e *Eventer) error {
...
+
+		return nil
+	}
+
+	if err := entMgr.FindAndDeactivateCustomerSubscription(event.Context(), lo.ValueOr(event.Properties(), "ID", "").(string)); err != nil {
+		log.Err(err).Msg("Failed to deactivate customer subscription")
```
golanglemonade: Same here:, also missing the level here

```suggestion
		zerolog.Ctx(event.Context()).Error().Err(err).Msg("Failed to deactivate customer subscription")
```

internal/ent/hooks/organization.go
```diff
@@ -205,6 +207,15 @@ func HookOrganizationDelete() ent.Hook {
 				return v, err
 			}
 
+			// if this is a soft delete, update the org subscription to inactive
+			if entx.CheckIsSoftDelete(ctx) {
+				if err := updateOrgSubscriptionOnDelete(ctx, m); err != nil {
+					return v, err
+				}
+
+				return v, nil
```
golanglemonade: we don't want to return here; this skips all the other checks later to update the user auth session
```suggestion
```

internal/ent/hooks/organization.go
```diff
@@ -205,6 +207,15 @@ func HookOrganizationDelete() ent.Hook {
 				return v, err
 			}
 
+			// if this is a soft delete, update the org subscription to inactive
+			if entx.CheckIsSoftDelete(ctx) {
```
golanglemonade: I dont think we need this if here; line 196 already checks if its a delete op:

```
if !isDeleteOp(ctx, m) {
```

internal/ent/hooks/organization.go
```diff
@@ -395,6 +406,20 @@ func updateUserDefaultOrgOnDelete(ctx context.Context, m *generated.Organization
...
 }
 
+// updateOrgSubscriptionOnDelete updates the org subscription to inactive and sets the status to canceled
+func updateOrgSubscriptionOnDelete(ctx context.Context, m *generated.OrganizationMutation) error {
+	deletedOrgID, ok := m.ID()
+	if !ok {
+		return nil
+	}
+
+	if err := m.Client().OrgSubscription.Update().Where(orgsubscription.OwnerID(deletedOrgID)).SetActive(false).SetStripeSubscriptionStatus("canceled").SetExpiresAt(time.Now()).Exec(ctx); err != nil {
```
golanglemonade: Can you use the stripe constants, particularrly canceled (cancelled) cause that is a common mispelled word - and we are generally using the constants:

```suggestion
	if err := m.Client().OrgSubscription.Update().Where(orgsubscription.OwnerID(deletedOrgID)).SetActive(false).SetStripeSubscriptionStatus(stripe.SubscriptionStatusCanceled).SetExpiresAt(time.Now()).Exec(ctx); err != nil {
```

^ this might not work if you don't have the `stripe` import so don't just take the change :)

pkg/entitlements/subscriptions.go
```diff
@@ -227,3 +233,54 @@ func (sc *StripeClient) MapStripeSubscription(subs *stripe.Subscription) *Subscr
...
+	}
+
+	switch status {
+	case stripe.SubscriptionStatusActive,
+		stripe.SubscriptionStatusTrialing:
+		return true
+	case stripe.SubscriptionStatusPastDue,
+		stripe.SubscriptionStatusIncomplete:
+		// You might want to add a grace period for past_due or incomplete statuses
+		// This could be based on the number of days past due or other criteria
```
golanglemonade: is this a todo comment?

pkg/entitlements/subscriptions.go
```diff
@@ -227,3 +233,54 @@ func (sc *StripeClient) MapStripeSubscription(subs *stripe.Subscription) *Subscr
...
+// SubscriptionStatusTrialing          SubscriptionStatus = "trialing"
+
+// after exhausting all payment retry attempts, the subscription will become canceled or unpaid
+// subscription moves into cancelled if we set cancel_at_period_end: true and the period end passes
+// SubscriptionStatusUnpaid            SubscriptionStatus = "unpaid"
+// SubscriptionStatusCanceled          SubscriptionStatus = "canceled"
+
+func IsSubscriptionActive(status stripe.SubscriptionStatus) bool {
+	// this shouldn't happen but including for sanity
+	if status == "" {
```
golanglemonade: you shouldn't need this; it will fall into the switch case default of `false`

internal/ent/hooks/organization.go
matoszz: well, no, because using the type here is inappropriate and it's just a string constant defined by them; plus there's no risk of a typo when the function call has it in there spelled correctly

pkg/entitlements/subscriptions.go
matoszz: nope

golanglemonade (approved):

golanglemonade (approved):

PR #689, opened 2025-05-03
Feat: summarise policies and procedures

adelowo:
Added two summarisers:

- [x] LexRank algorithm ( http://github.com/didasy/tldr ) Default
- [x] AI 
  - [x] OpenAI 
  - [x] Anthropic/Claude 
  - [x] Mistral
  - [x] Gemini 
  - [x] cloudflare
  - [x] hugging face 

### Resources

Added to both policies and procedures 

UI: https://github.com/theopenlane/openlane-ui/pull/398


internal/ent/entconfig/config.go
```diff
@@ -4,4 +4,76 @@ package entconfig
 type Config struct {
 	// EntityTypes is the list of entity types to create by default for the organization
 	EntityTypes []string `json:"entityTypes" koanf:"entityTypes" default:"" description:"entity types to create for the organization"`
+
+	Summarizer Summarizer `json:"summarizer" koanf:"summarizer"`
+}
+
+type Summarizer struct {
```
matoszz, line 13: You're going to want to include comments on all the exported structs in the format of // StructName is...

adelowo: Will do that now

adelowo: https://github.com/theopenlane/core/pull/689/commits/52ae9321f7c0b5922497c2100cd023bb5cbb6b58

internal/ent/hooks/internalpolicy.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// HookPolicySummarize summarizes the policy and produces a short human readable copy
+func HookPolicySummarize() ent.Hook {
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.InternalPolicyFunc(func(ctx context.Context, m *generated.InternalPolicyMutation) (generated.Value, error) {
+			details, ok := m.Details()
+			if !ok {
+				return nil, fmt.Errorf("details does not exists") // nolint:err113
+			}
```
golanglemonade: I dont think we want to return an error, instead we want to only do the summary:

```
if ok && details != ""
```

adelowo: Ha yes. I actually fixed this but didn't push. Will do that as soon as I get back on desk

internal/ent/hooks/procedure.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+
+	"entgo.io/ent"
+
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/hook"
+)
+
+// HookProcedureSummarize summarizes the procedure and produces a short human readable copy
+func HookProcedureSummarize() ent.Hook {
```
golanglemonade: We can actually use the same hook for both policy + procedure (I would actually say we want to add this to the `mixin_document` so that any schema that uses the document mixin with the `details` has the summary field). The `revision` hook is a good example of using a hook for multiple schemas with the same field using an interface to get the field if you want to look at that as an example.

internal/ent/schema/internalpolicy.go
```diff
@@ -38,7 +40,14 @@ func (InternalPolicy) PluralName() string {
 
 // Fields returns policy fields.
 func (InternalPolicy) Fields() []ent.Field {
-	return []ent.Field{} // fields are defined in the mixins
+	// other fields are defined in the mixins
+	return []ent.Field{
```
golanglemonade: We can go ahead and add this to the `DocumentMixin` instead so its used everywhere

pkg/summarizer/llm.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/tmc/langchaingo/llms/mistral"
+	"github.com/tmc/langchaingo/llms/ollama"
+	"github.com/tmc/langchaingo/llms/openai"
+
+	"github.com/theopenlane/core/internal/ent/entconfig"
+)
+
+// maybe make a config value too?
+const prompt = `
+	Summarise the following text in the line below. Be brief, concise and precise.
```
golanglemonade:
```suggestion
	Summarize the following text in the line below. Be brief, concise and precise. 
```

pkg/summarizer/llm.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/tmc/langchaingo/llms/googleai"
+	"github.com/tmc/langchaingo/llms/huggingface"
+	"github.com/tmc/langchaingo/llms/mistral"
+	"github.com/tmc/langchaingo/llms/ollama"
+	"github.com/tmc/langchaingo/llms/openai"
+
+	"github.com/theopenlane/core/internal/ent/entconfig"
+)
+
+// maybe make a config value too?
```
golanglemonade, line 20: todo comment? if so can you add `TODO: ` in front?

pkg/summarizer/llm.go
```diff
@@ -0,0 +1,218 @@
...
+const prompt = `
+	Summarise the following text in the line below. Be brief, concise and precise. 
+
+	%s
+	`
+
+type LLMSummarizer struct {
```
golanglemonade: Comment exported struct

golanglemonade (commented):
Can you also add a test to policies or procedures create + update to make sure its set correctly when details is set?

db/migrations-goose-postgres/20250503121240_add_summary.sql
```diff
@@ -0,0 +1,19 @@
+-- +goose Up
```
golanglemonade: Can you please squash this migration and the other into one - you should be able to delete the two new migraiton files; resetset the hash, and then rerun `db:create`

internal/ent/schema/internalpolicy.go
```diff
@@ -86,6 +87,7 @@ func (InternalPolicy) Hooks() []ent.Hook {
 			hooks.HookOrgOwnedTuples(),
 			ent.OpCreate|ent.OpUpdateOne|ent.OpUpdateOne,
 		),
+		hooks.HookSummarizeDetails(),
```
golanglemonade: Add this hook the document mixin directly instead of each schema

internal/graphapi/internalpolicy_test.go
```diff
@@ -511,6 +535,10 @@ func (suite *GraphTestSuite) TestMutationUpdateInternalPolicy() {
 				assert.Equal(t, *tc.request.Name, resp.UpdateInternalPolicy.InternalPolicy.Name)
 			}
 
+			if tc.request.Details != nil {
+				assert.NotEmpty(t, resp.UpdateInternalPolicy.InternalPolicy.Summary)
```
golanglemonade, line 539: you'll also want to ensure that `resp.UpdateInternalPolicy.InternalPolicy.Summary` does not equal the original summary when it was created.

golanglemonade (approved):

PR #702, opened 2025-05-05
Schemas + graphql endpoints for vanity domains.

acookin:
Implement the db migrations and graphql apis to support custom domains (vanity domains), allowing organizations to map their own custom domains to Openlane-provided domains

* Added MappableDomain entity: Represents domains that Openlane provides that can be mapped to by custom domains. Should only be write-able by system admins
* Added CustomDomain entity: Represents customer-owned domains that map to Openlane domains

In a future PR, the CustomDomain's status will be updated by an async process. This process will scan the CustomDomain table and compare the hashed txt record value with the evaluated value of the txt record ($txt_record_subdomin.$cname_record) and ensure the cname record is also pointing at the expected mappable domain. If both things are true, the cname record will be updated to a "verified" status.


internal/ent/hooks/custom_domain.go
```diff
@@ -0,0 +1,74 @@
...
+		return "", err
+	}
+
+	encodedString := base64.URLEncoding.EncodeToString(bytes)[:domainValidationSecretLength]
+
+	return encodedString, nil
+}
+
+// VerifyDomainValidationSecret verifies if the provided secretString matches the hashed value
+func VerifyDomainValidationSecret(hashedSecret, secretString string) bool {
```
acookin, line 71: Not used yet, but this is how the async process will validate that the txt record value matches the hashed string we have in the DB

internal/ent/hooks/custom_domain.go
```diff
@@ -0,0 +1,74 @@
...
+				if u, ok := u.(*generated.CustomDomain); ok {
+					// Return the original secretString to the caller, not the hashed value
+					// this will be readable only in the Create mutation response, and not retrievable later
+					u.TxtRecordValue = secretString
+				}
+
+				return u, err
+			})
+		},
+		hook.HasFields("txt_record_value"),
```
golanglemonade: Does this need to also filter on the create op: `hook.HasOp(ent.OpCreate)`?

internal/ent/schema/custom_domain.go
```diff
@@ -0,0 +1,148 @@
...
+			NotEmpty().
+			Immutable(),
+		field.String("txt_record_subdomain").
+			Comment("String to be prepended to the cname_record, used to evaluate domain ownership.").
+			Annotations(
+				entgql.Skip(entgql.SkipWhereInput, entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
+			).
+			Annotations(
+				entgql.Skip(entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
+			).
```
golanglemonade: Duplicate annotations, one should be removed
```suggestion
			Annotations(
				entgql.Skip(entgql.SkipWhereInput, entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
			).
```

internal/ent/schema/mappable_domain.go
```diff
@@ -0,0 +1,95 @@
...
+			Immutable().
+			Annotations(
+				entgql.OrderField("name"),
+			),
+	}
+}
+
+// Mixin of the MappableDomain
+func (e MappableDomain) Mixin() []ent.Mixin {
+	return mixinConfig{}.getMixins()
```
golanglemonade, line 61: does this need the org owned mixin so its mapped back to the organization as the `owner_id`?

golanglemonade (commented):
🎉 nice!

internal/ent/schema/mappable_domain.go
acookin: No, these aren't associated with an organization or an owner, they should be global imo

internal/ent/hooks/custom_domain.go
acookin: Probably _only_ needs to filter on the create op, since this field is immutable.

PR #714, opened 2025-05-06
Clone controls from a standard into a program

adelowo:


fga/model/model.fga
```diff
@@ -413,6 +413,8 @@ type standard
     define parent_editor: admin from parent or owner from parent
     define viewer: [user, service] or editor
     define editor: [user, service]
+    # Add permission to create controls under this standard
+    define can_create_control: can_edit or parent_editor
```
golanglemonade: We shouldn't need this; if you have edit access to a standard (system-admins under system-owned and can_edit for org owned) you should be able to add controls). Ping me on slack if you have questions about this

internal/graphapi/programextended.resolvers.go
```diff
@@ -76,8 +76,26 @@ func (r *mutationResolver) CreateFullProgram(ctx context.Context, input model.Cr
...
+			return nil, parseRequestError(err, action{action: ActionGet, object: "standard"})
+		}
+
+		controls, err := standardRes.QueryControls().All(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, control := range controls {
+			createdProgram.AddControlIDs(control.ID)
```
golanglemonade, line 98: We can't just add the IDs of the controls; the controls have to be _cloned_ under the organization. You should be able to use/call this resolver: 

https://github.com/theopenlane/core/blob/main/internal/graphapi/controlextended.resolvers.go#L19

internal/graphapi/programextended_test.go
```diff
@@ -91,6 +91,19 @@ func (suite *GraphTestSuite) TestMutationCreateProgramWithMembers() {
...
 	t := suite.T()
 
+	numControls := 5
+	controlIDs := []string{}
+	for range numControls {
+		control := (&ControlBuilder{client: suite.client}).MustNew(testUser1.UserCtx, t)
+		controlIDs = append(controlIDs, control.ID)
+	}
+
+	resp, err := suite.client.api.CreateStandard(testUser1.UserCtx, openlaneclient.CreateStandardInput{
```
golanglemonade, line 102: This is a good test for a org owned standard; You should also add a test for a system-owned standard, e.g. from the controls test we have:

```
	publicStandard := (&StandardBuilder{client: suite.client, IsPublic: true}).MustNew(systemAdminUser.UserCtx, t)
```

This way you can ensure you can copy controls from a system-owned standard, into an organization

internal/ent/schema/standard.go
```diff
@@ -126,14 +126,9 @@ func (Standard) Fields() []ent.Field {
...
-			edgeSchema: Control{},
-			annotations: []schema.Annotation{
-				// skip the ability to create and update controls via the standard
-				// TODO: (sfunk) implement permissions on parent edge to allow children to be created
-				// and have the permissions added to fga
-				entgql.Skip(entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
-			},
+			fromSchema:  s,
+			edgeSchema:  Control{},
+			annotations: []schema.Annotation{},
```
golanglemonade:
```suggestion
```

internal/graphapi/controlextended.resolvers.go
```diff
@@ -37,96 +34,9 @@ func (r *mutationResolver) CreateControlsByClone(ctx context.Context, input *mod
...
-	query, err := withTransactionalMutation(ctx).Control.Query().Where(control.IDIn(createdControlIDs...)).CollectFields(ctx)
+	createdControls, err := r.cloneControls(ctx, existingControls, input.ProgramID, false)
 	if err != nil {
-		return nil, parseRequestError(err, action{action: ActionCreate, object: "control"})
-	}
-
-	createdControls, err := query.All(ctx)
-	if err != nil {
-		return nil, parseRequestError(err, action{action: ActionCreate, object: "control"})
+		return nil, err
```
golanglemonade: Resolver should return the parsed error instead of a generic error
```suggestion
		return nil, parseRequestError(err, action{action: ActionCreate, object: "control"})
```

internal/graphapi/control_clone.go
```diff
@@ -0,0 +1,110 @@
...
+			}
+		}
+
+		if programID != nil {
+			controlInput.ProgramIDs = []string{*programID}
+		}
+
+		res, err := withTransactionalMutation(ctx).Control.Create().SetInput(controlInput).Save(ctx)
+		if err != nil {
+			return nil, parseRequestError(err, action{action: ActionCreate, object: "control"})
```
golanglemonade: helper function should just return the error and let the resolver handle the parsing
```suggestion
			return nil, err
```

internal/graphapi/control_clone.go
```diff
@@ -0,0 +1,110 @@
...
+
+		if err := r.cloneSubcontrols(ctx, control, res.ID); err != nil {
+			return nil, err
+		}
+	}
+
+	// get the created controls
+	query, err := withTransactionalMutation(ctx).Control.Query().Where(control.IDIn(createdControlIDs...)).CollectFields(ctx)
+	if err != nil {
+		return nil, parseRequestError(err, action{action: ActionCreate, object: "control"})
```
golanglemonade:
```suggestion
		return nil, err
```

internal/graphapi/controlextended.resolvers.go
```diff
@@ -20,13 +19,11 @@ func (r *mutationResolver) CreateControlsByClone(ctx context.Context, input *mod
 	// set the organization in the auth context if its not done for us
 	if err := setOrganizationInAuthContext(ctx, input.OwnerID); err != nil {
 		log.Error().Err(err).Msg("failed to set organization in auth context")
-
 		return nil, rout.NewMissingRequiredFieldError("owner_id")
 	}
 
 	existingControls, err := withTransactionalMutation(ctx).Control.Query().
 		Where(control.IDIn(input.ControlIDs...)).
-		// WithMappedControls(). // TODO: uncomment once mapped controls are implemented
```
golanglemonade, line 29: okay to remove this comment as long as you remember to add it with the mapped controls stuff :)

golanglemonade (approved):

PR #719, opened 2025-05-07
chore: speed up graphapi tests, use gotestsum

golanglemonade:
- moves away from testify suite for the longest test package (graphapi)
- uses `gotestsum` which gives better output, auto retries, and integrates with buildkite test analytics
- moves to https://github.com/gotestyourself/gotest.tools from testify assert + require as it works well with gotestsum and is a simpler package
- adding `coverage` loses the caching in buildkite, but its not big difference anymore

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -163,6 +163,9 @@ steps:
           - artifacts#v1.9.4:
               download: "coverage.out"
               step: "go_test"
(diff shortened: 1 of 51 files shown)


golanglemonade:
This is intentionally not set to auto merge, so I can check with the team about outstanding branches and test conflicts.

adelowo (approved):

PR #734, opened 2025-05-16
Job runners for agents workflow

adelowo:
> had to break the PRs up as they keep getting larger and larger

### How it works

(self hosted) runners/Agents can be registered to run these control checks. There are 3 main components here:

- registration tokens for a runner 
- tokens the agent actually uses to receive jobs to run ; requires step 1, _a node/device_ must be registered
- actually registering the node  https://github.com/theopenlane/agent/blob/main/config.sh

These agents ( openlane/customers self hosted ) will then execute the jobs. The linked script does that and setups the nodes/job runners correctly


internal/ent/hooks/job.go
```diff
@@ -0,0 +1,85 @@
...
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.JobRunnerRegistrationTokenFunc(
+			func(ctx context.Context, m *generated.JobRunnerRegistrationTokenMutation) (generated.Value, error) {
+				if m.Op().Is(ent.OpCreate) {
+					orgID, err := auth.GetOrganizationIDFromContext(ctx)
+					if err != nil {
+						return nil, err
+					}
+
+					_, err = m.Client().JobRunnerRegistrationToken.Delete().
```
golanglemonade, line 31: Why hard delete the token here but soft delete on line 44?

internal/ent/hooks/job.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				runner := v.(*generated.JobRunner)
+
+				// make sure we cannot reuse the registration token
+				// for cases where there is no "registration token"
+				// like an admin creating a runner via api
+				// or even tests.
+				// Only check the error and make sure it is not a "not found".
+				err = m.Client().JobRunnerRegistrationToken.DeleteOneID(runner.CreatedBy).
```
golanglemonade: Should this be soft deleted instead of hard deletion?

internal/ent/privacy/rule/job_runner.go
```diff
@@ -0,0 +1,54 @@
...
+			return err
+		}
+
+		if isAdmin {
+			return privacy.Allow
+		}
+
+		if m.Op() == ent.OpCreate {
+			return privacy.Allow
+		}
```
golanglemonade: since this always returns allow, it should first; before the check for system admin

internal/ent/privacy/rule/job_runner.go
```diff
@@ -0,0 +1,54 @@
...
+
+		ids, err := m.IDs(ctx)
+		if err != nil {
```
golanglemonade, line 34: I think you may also want to check if `ids` is empty and return early if so, since there would be nothing to do

internal/ent/privacy/token/job_runner.go
```diff
@@ -0,0 +1,47 @@
...
+func (token *JobRunnerRegistrationToken) GetToken() string {
+	return token.token
+}
+
+// SetToken sets the token
+func (token *JobRunnerRegistrationToken) SetToken(s string) {
+	token.token = s
+}
+
+// NewContextWithJobRunnerRegistrationToken  returns a new context with the job runner registration token
```
golanglemonade:
```suggestion
// NewContextWithJobRunnerRegistrationToken returns a new context with the job runner registration token
```

internal/ent/schema/jobrunner.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Policy of the JobRunner
+func (JobRunner) Policy() ent.Policy {
+	return policy.NewPolicy(
+		policy.WithQueryRules(
+			privacy.AlwaysAllowRule(),
+		),
+		policy.WithMutationRules(
+			rule.AllowIfContextHasPrivacyTokenOfType[*token.JobRunnerRegistrationToken](),
+			rule.AllowMutationAfterApplyingOwnerFilter(),
```
golanglemonade: this is going to allow any authenticated user to edit a job runner in their organization, even if they only have read-only access, I dont think you want or need this here.

fga/model/model.fga
```diff
@@ -416,6 +416,19 @@ type standard
     # this is used for organization custom standards that are only available to the organization
     define parent: [organization]
 
+type job_runner
```
golanglemonade, line 419: this works for now; but I think we will want to go later and allow creators + editors based on groups like some of the other objects - this will allow non-org admins to create job runners (which is probably going to be a common case where you'd give an engineering team access just to set this up and not full admin access to the system)

internal/graphapi/jobrunner.resolvers.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"context"
+
+	"github.com/rs/zerolog/log"
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/jobrunner"
+	"github.com/theopenlane/core/internal/graphapi/model"
+	"github.com/theopenlane/utils/rout"
+)
+
```
golanglemonade, line 16: Why is there no `Create` resolver? I assume because you use the registration resolver to actually do the create?

internal/graphapi/jobrunner_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+			assert.NilError(t, err)
+			assert.Assert(t, resp != nil)
+
+			assert.Check(t, is.Len(resp.JobRunners.Edges, tc.expectedCount))
+		})
+	}
+
+	(&Cleanup[*generated.JobRunnerDeleteOne]{client: suite.client.db.JobRunner, ID: systemJob.ID}).MustDelete(systemAdminUser.UserCtx, t)
+	(&Cleanup[*generated.JobRunnerDeleteOne]{client: suite.client.db.JobRunner, ID: firstJob.ID}).MustDelete(testUser1.UserCtx, t)
```
golanglemonade: you don't need to change this, but just an FYI in case you didn't see this you can do multiple in one call: 
```
	(&Cleanup[*generated.JobRunnerDeleteOne]{client: suite.client.db.JobRunner, IDs: []string{firstJob.ID, secondJob.ID}}).MustDelete(testUser1.UserCtx, t)
```

pkg/models/ipaddress_test.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func TestValidateIP(t *testing.T) {
+	tt := []struct {
+		name     string
+		ip       string
+		hasError bool
+	}{
+		{
+			name:     "empty string not allowed",
```
golanglemonade, line 17: can you add an ipv6 test?

golanglemonade (commented):
a lot here! nice work - few comments for you

internal/ent/hooks/job.go
adelowo: the soft delete mixin is added so it soft deletes but I updated

fga/model/model.fga
adelowo: will keep that in mind

internal/graphapi/jobrunner.resolvers.go
adelowo: yes correct

internal/graphapi/jobrunner_test.go
```diff
@@ -91,36 +91,37 @@ func TestMutationDeleteJobRunner(t *testing.T) {
 		{
 			name:          "happy path user",
 			client:        suite.client.api,
-			ctx:           testUser1.UserCtx,
+			ctx:           rule.WithInternalContext(testUser1.UserCtx),
```
golanglemonade: Why bypass the auth checks here?

adelowo: ha copy/paste while I was debugging the new interceptor. will fix

adelowo: 5b9365cfc1bbebff1f4c8d38a93f31f6bf53fa69

golanglemonade (approved):

PR #736, opened 2025-05-17
update old custom domain fields

acookin:
Deprecate custom domain fields.
In a follow up, I'll add new fields to track cloudflare and ssl verification/creation.

These build failures look transient, but I'm not a part of the buildkite org so I don't think I can rebuild them


internal/ent/schema/customdomain.go
```diff
@@ -66,15 +66,15 @@ func (CustomDomain) Fields() []ent.Field {
 				entgql.Skip(entgql.SkipWhereInput, entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
 			).
 			Default("_olverify").
-			NotEmpty().
 			Immutable().
+			Optional().
+			Deprecated("Field no longer used").
```
golanglemonade, line 71: you can just completely drop it; we don't have to worry about a breaking change before we go live

internal/ent/schema/customdomain.go
```diff
@@ -87,13 +87,16 @@ func (CustomDomain) Fields() []ent.Field {
 				// value back with the Create mutation response
 				return str
 			}).
+			Optional().
+			Deprecated("Field no longer used").
```
golanglemonade, line 91: same with these

PR #753, opened 2025-05-21
bug: fix ability to update permissions on managed group to other objects

golanglemonade:
You could previously add a group to an object (e.g. on a program, set a group to have edit permissions) but you couldn't do the opposite - edit a managed group to add a permission to a program. This fixes it to allow it from either object. 

I added tests as well as tested this from the UI


adelowo (approved):
the displayID check is great

PR #755, opened 2025-05-22
Paginate group members and fetch user details

adelowo:
- added the relay connection to the members of a group so we can fetch more details .
- Updated the tests too

CleanShot 2025-05-22 at 16 09 24


cmd/cli/cmd/group/root.go
```diff
@@ -86,7 +86,7 @@ func tableOutput(out []openlaneclient.Group) {
 			isManaged = *i.IsManaged
 		}
 
-		writer.AddRow(i.ID, i.Name, i.DisplayName, *i.Description, i.Setting.Visibility, isManaged, i.Owner.DisplayName, len(i.Members))
+		writer.AddRow(i.ID, i.Name, i.DisplayName, *i.Description, i.Setting.Visibility, isManaged, i.Owner.DisplayName, len(i.Members.Edges))
```
golanglemonade: You are going to want to grab the `totalCount` field here now instead, since `edges` will be limited to 100

adelowo: ha nice catch. true. Thanks!

adelowo: fixed in a4f1f52c

internal/graphapi/query/group.graphql
```diff
@@ -186,12 +213,16 @@ query GetAllGroups {
           visibility
         }
         members {
-          id
-          role
-          user {
-            id
-            firstName
-            lastName
```
golanglemonade, line 194: you'll need to add `totalCount` to whichever query is used by the cli  (and run `task generate:gql`)

golanglemonade (approved):

PR #765, opened 2025-05-24
New fields and tables and endpoints for cloudflare custom domain verification

acookin:
New bits and bobs for tracking custom domain verification in cloudflare.

Need to do some updating for hooks n stuff, but this PR is already pretttttyyyy big, so going to do this in a follow up.


internal/ent/schema/customdomain.go
```diff
@@ -59,14 +60,21 @@ func (CustomDomain) Fields() []ent.Field {
 			Comment("The mappable domain id that this custom domain maps to").
 			NotEmpty().
 			Immutable(),
+		// TODO skip
```
golanglemonade, line 62: What does this TODO mean?

internal/ent/schema/customdomain.go
```diff
@@ -99,8 +114,15 @@ func (CustomDomain) Policy() ent.Policy {
 	return policy.NewPolicy(
 		policy.WithQueryRules(
 			policy.CheckOrgReadAccess(),
+			rule.AllowQueryIfSystemAdmin(),
+		),
+		policy.WithOnMutationRules(
+			ent.OpUpdateOne|ent.OpUpdate,
+			rule.AllowMutationIfSystemAdmin(),
+			privacy.AlwaysDenyRule(),
```
golanglemonade, line 122: You don't need this one, the default rules always have a Deny after everything else: https://github.com/theopenlane/core/blob/d3ddb7632bd6775676b6b78b681d6cb71a555127/internal/ent/privacy/policy/base.go#L21-L29

golanglemonade: You don't need this one, the default rules always have a Deny after everything else: https://github.com/theopenlane/core/blob/d3ddb7632bd6775676b6b78b681d6cb71a555127/internal/ent/privacy/policy/base.go#L21-L29



```suggestion
```

internal/ent/schema/dnsverification.go
```diff
@@ -0,0 +1,135 @@
...
+// Policy of the DNSVerification
+func (DNSVerification) Policy() ent.Policy {
+	return policy.NewPolicy(
+		policy.WithQueryRules(
+			policy.CheckOrgReadAccess(),
+			rule.AllowQueryIfSystemAdmin(),
+		),
+		policy.WithMutationRules(
+			rule.AllowMutationIfSystemAdmin(),
+			privacy.AlwaysDenyRule(),
```
golanglemonade: Same here

```suggestion
```

internal/graphapi/customdomain_test.go
```diff
@@ -258,13 +258,7 @@ func TestMutationDeleteCustomDomain(t *testing.T) {
 			name:   "delete domain",
 			id:     customDomain.ID,
 			client: suite.client.api,
-			ctx:    testUser1.UserCtx,
-		},
-		{
-			name:   "delete domain, admin user",
-			id:     customDomain2.ID,
```
golanglemonade, line 265: you are no longer deleting this domain as part of your tests, and aren't cleaning it up. I think it can just be completely removed from the tests

golanglemonade (approved):

PR #768, opened 2025-05-26
Jobs defintion + cadence + results

adelowo:
this adds the following:

- scheduled jobs ( which can be ___templates___ ) can be created by system level admins or org owners . this would also work for marketplace templates 
- ability to ___clone a job___ from the templates from 1. The templates will define default cadence/cron but you can override if need be
- added a jobs results table so the agent can upload execution results ( no mutations, only queries )
- can link controls/subcontrols to an automated job
- added a ___run history___ to keep track of when a runner picked up a job and others
- added basic fga checks for now to allow creation and limit only to org editors

> I shared a mermaid file, you can check it


fga/model/model.fga
```diff
@@ -56,6 +56,10 @@ type organization
     define can_invite_members: can_view or can_edit or can_invite_members from parent
     # only allow users with edit access to the org to invite other admins
     define can_invite_admins: can_edit or can_invite_admins from parent
+    ## only allow admins/owners be able to create scheduled jobs
+    ## and enable a job that can be run and be linked to a control
+    define can_create_scheduled_job: can_edit
```
golanglemonade: I would go ahead and add a `creator` and update to (for both, just one example):

```
define scheduled_job_creator: [group#member]
define can_create_scheduled_job: can_edit or scheduled_job_creator
```

This allows org admins to assign groups of users that are not admins the ability to also create these objects.

internal/ent/hooks/job_definition.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"entgo.io/ent"
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/hook"
+	"github.com/theopenlane/core/pkg/models"
+	"github.com/theopenlane/entx"
+)
+
+// HookScheduledJobCreate verifies a scheduled job has
+// the a cadence set or a cron and the configuration matches what is expected
```
golanglemonade:
```suggestion
// a cadence set or a cron and the configuration matches what is expected
```

internal/ent/hooks/job_definition.go
```diff
@@ -0,0 +1,92 @@
...
+			if err := validateCadenceOrCron(&cadence, hasCadence, cron, hasCron); err != nil {
+				return nil, err
+			}
+
+			return next.Mutate(ctx, mutation)
+		})
+	}, ent.OpUpdate|ent.OpUpdateOne|ent.OpCreate)
+}
+
+// HookScheduledJobCreate verifies a job that can be attached to a control/subcontrol has
```
golanglemonade:
```suggestion
// HookControlScheduledJobCreate verifies a job that can be attached to a control/subcontrol has
```

internal/ent/hooks/job_definition.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+
+			return next.Mutate(ctx, mutation)
+		})
+	}, ent.OpUpdate|ent.OpUpdateOne|ent.OpCreate)
+}
+
+func validateCadenceOrCron(cadence *models.JobCadence, hasCadence bool, cron string, hasCron bool) error {
+	if !hasCadence && (!hasCron || cron == "") {
+		return errors.New("either cadence or cron must be specified") // nolint:err113
```
golanglemonade: instead of `nolint` these; can you define the errors, we have a bunch defined already in `hooks/errors.go` - this allows us to check error types later if needed; something like a `generated.ValidationError`; with the specific message would be helpful for consistency.

internal/ent/schema/job_definition.go
```diff
@@ -0,0 +1,150 @@
+package schema
```
golanglemonade: File name and schema name don't match; can you update?

internal/ent/schema/job_definition.go
```diff
@@ -0,0 +1,150 @@
...
+			).
+			NotEmpty(),
+		field.String("description").
+			Annotations(
+				entx.FieldSearchable(),
+			).
+			Comment("the description of the job").
+			Optional(),
+		field.Enum("job_type").
+			GoType(enums.JobType("")).
```
golanglemonade: Should this be optional? Right now I think this would mean you'd have to define a job from our list?

internal/ent/schema/job_definition.go
```diff
@@ -0,0 +1,150 @@
...
+			Default(enums.JobTypeSsl.String()).
+			Annotations(
+				entgql.OrderField("job_type"),
```
golanglemonade: I haven't figured out if this is required, but all the docs have the enum order by as all caps
```suggestion
				entgql.OrderField("JOB_TYPE"),
```

internal/ent/hooks/job_definition.go
```diff
@@ -0,0 +1,92 @@
...
+				return nil, err
+			}
+
+			return next.Mutate(ctx, mutation)
+		})
+	}, ent.OpUpdate|ent.OpUpdateOne|ent.OpCreate)
+}
+
+func validateCadenceOrCron(cadence *models.JobCadence, hasCadence bool, cron string, hasCron bool) error {
+	if !hasCadence && (!hasCron || cron == "") {
```
golanglemonade: I think we should probably allow a job to be setup without any of these initially. This would allow:
- users to run without our schedulers
- create a job and then come back and set the cadence
- ability for jobs to just run on demand if they wanted 
- we can always add a warning in the UI for jobs configured on our infra without a schedule that they won't be run unless triggered manually

internal/ent/schema/control_job.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (ControlScheduledJob) PluralName() string {
+	return pluralize.NewClient().Plural(SchemaControlScheduledJob)
+}
+
+// Fields of the ControlScheduledJob
+func (ControlScheduledJob) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("job_id"),
```
golanglemonade: field is missing a comment

internal/ent/schema/control_job.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Annotations of the ControlScheduledJob
+func (ControlScheduledJob) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		// entfga.SelfAccessChecks(),
+		entgql.RelayConnection(),
+		history.Annotations{
+			Exclude: true,
+		},
```
golanglemonade: Why exclude history here?

internal/ent/schema/control_job.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Indexes of the ControlScheduledJob
+func (ControlScheduledJob) Indexes() []ent.Index {
+	return []ent.Index{}
+}
+
+// Annotations of the ControlScheduledJob
+func (ControlScheduledJob) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		// entfga.SelfAccessChecks(),
```
golanglemonade, line 113: Is this because you are coming back to this later?

internal/ent/schema/control_job.go
```diff
@@ -0,0 +1,148 @@
...
+		entgql.RelayConnection(),
```
golanglemonade: You shouldn't need this; it's included by default with `GraphQLAnnotationMixin` which is in the default mixins
```suggestion
```

internal/ent/schema/job_definition.go
```diff
@@ -0,0 +1,150 @@
...
+	return []ent.Index{}
+}
+
+// Annotations of the ScheduledJob
+func (ScheduledJob) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		// entfga.SelfAccessChecks(),
+		history.Annotations{
+			Exclude: true,
+		},
```
golanglemonade: Same comments here as control job - why exclude history?

internal/ent/schema/jobresult.go
```diff
@@ -0,0 +1,159 @@
...
+
+// JobResult holds the schema definition for the JobResult entity
+type JobResult struct {
+	SchemaFuncs
+
+	ent.Schema
+}
+
+// SchemaJobResultis the name of the schema in snake case
+const SchemaJobResult = "scheduled_job_result"
```
golanglemonade: I think this can just be `job_result` for consistency with the name of the schema + file?
```suggestion
const SchemaJobResult = "job_result"
```

internal/ent/schema/jobresult.go
```diff
@@ -0,0 +1,159 @@
...
+
+// PluralName returns the plural name of the schema
+func (JobResult) PluralName() string {
+	return pluralize.NewClient().Plural(SchemaJobResult)
+}
+
+// Fields of the JobResult
+func (JobResult) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("scheduled_job_id"),
```
golanglemonade: Missing comment on most of these field definitions

internal/ent/schema/jobresult.go
```diff
@@ -0,0 +1,159 @@
...
+			Default(time.Now).
+			Annotations(
+				entgql.Skip(
+					entgql.SkipMutationUpdateInput,
+				),
+				entgql.OrderField("started_at"),
+			).
+			Comment("The time the job started it's execution. This is different from the db insertion time"),
+
+		field.String("file_id").
```
golanglemonade, line 86: Comment but just checking - these are the log files? or what is the file here?

internal/ent/schema/scheduledjobrun.go
```diff
@@ -0,0 +1,116 @@
...
+			Optional().
+			Comment("The runner that this job will be executed on. Useful to know because of self hosted runners"),
+
+		field.Enum("status").
+			GoType(enums.ScheduledJobRunStatus("")).
+			Default(enums.ScheduledJobRunStatusPending.String()).
+			Comment(`The status of the job to be executed. By default will be pending but when 
+			scheduled on a runner, this will change to acquired.`),
+
+		field.String("scheduled_job_id"),
```
golanglemonade: Missing field Comment

internal/ent/schema/scheduledjobrun.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Indexes of the ScheduledJobRun
+func (ScheduledJobRun) Indexes() []ent.Index {
+	return []ent.Index{}
+}
+
+// Annotations of the ScheduledJobRun
+func (ScheduledJobRun) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		history.Annotations{
```
golanglemonade, line 88: same question about excluding history - this one might make sense because it's already kind of a history of runs though - just wanted to be sure.

internal/graphapi/controlscheduledjob.resolvers.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, parseRequestError(err, action{action: ActionGet, object: "scheduledjob"})
+	}
+
+	if input.Cadence == nil && input.Cron == nil {
+		if !scheduledJob.Cadence.IsZero() {
+			input.Cadence = &scheduledJob.Cadence
+		} else if scheduledJob.Cron != nil {
+			input.Cron = scheduledJob.Cron
+		}
+	}
```
golanglemonade: Let's do this check before the query so we don't run the query if we don't need it

internal/graphapi/controlscheduledjob.resolvers.go
```diff
@@ -0,0 +1,164 @@
...
+		Where(scheduledjob.ID(input.JobID)).
+		CollectFields(ctx)
+	if err != nil {
+		return nil, parseRequestError(err, action{action: ActionGet, object: "scheduledjob"})
+	}
+
+	scheduledJob, err := query.Only(ctx)
+	if err != nil {
+		return nil, parseRequestError(err, action{action: ActionGet, object: "scheduledjob"})
+	}
```
golanglemonade: I dont think you need `CollectFields` here - this is used to limit the columns returned from a graph request; but since you are writing the query you can simply do:

```suggestion
	// by default copy the cadence or cron from the "scheduled job" template
	// if this is not provided here
	scheduledJob, err := withTransactionalMutation(ctx).ScheduledJob.Query().
	        Select("candence", "cron"). // just select the fields you want
		Where(scheduledjob.ID(input.JobID)).
		Only(ctx)
	if err != nil {
		return nil, parseRequestError(err, action{action: ActionGet, object: "scheduled_job"})
	}
```

internal/graphapi/controlscheduledjob.resolvers.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, rout.ErrPermissionDenied
+	}
+
+	if input.Cadence == nil && input.Cron == nil {
+		if !res.Cadence.IsZero() {
+			input.Cadence = &res.Cadence
+		} else if res.Cron != nil {
+			input.Cron = res.Cron
+		}
+	}
```
golanglemonade: this is an update so these should already be set - we shouldn't need to reset them on an update

pkg/models/job_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/theopenlane/core/pkg/models"
+	"gotest.tools/v3/assert"
+)
+
+func TestValidateCronExpression(t *testing.T) {
```
golanglemonade: this test should be in a file called `jobcadence_test.go`

pkg/models/job_cadence.go
```diff
@@ -0,0 +1,150 @@
...
+var validWeekdaysSet = func() map[string]struct{} {
+	m := make(map[string]struct{})
+	for _, v := range enums.JobWeekdayFriday.Values() {
+		m[v] = struct{}{}
+	}
+	return m
+}()
+
+// Validate makes sure we have a usable job cadence setting
+func (c *JobCadence) Validate() error {
```
golanglemonade, line 70: Can you add a unit test for this one too; its a pretty complex function and should be fairly easy to test

internal/ent/schema/job_definition.go
adelowo: I don't think it should be optional, we will have an _others_ job type for community templates/marketplace

internal/ent/schema/jobresult.go
adelowo: yes correct. this will hold the log file we spoke about

internal/ent/schema/control_job.go
adelowo: yes correct.

golanglemonade (approved):

PR #770, opened 2025-05-26
acme solver handler

acookin:
Handler to solve acme challenges, so cloudflare handles SSL for the custom domains.

When we set up the custom hostname in cloudflare, we specify that we want an HTTP challenge for the ssl verification. We do this so the user doens't have to add 2 txt records to their zone.

Once the SSL bits in cloudflare are initialized, we just have to correctly respond to the challenge, then the cloudflare proxy will handle ssl for the custom hostname.

adds new auth context:
https://github.com/theopenlane/iam/pull/255


internal/httpserve/handlers/acmesolver.go
```diff
@@ -0,0 +1,39 @@
...
+// ACMESolverHandler handles ACME challenge requests by looking up the challenge path
+// and returning the expected challenge value for domain verification
+func (h *Handler) ACMESolverHandler(ctx echo.Context) error {
+	var in models.AcmeSolverRequest
+	if err := ctx.Bind(&in); err != nil {
+		return h.InvalidInput(ctx, err)
+	}
+
+	allowCtx := privacy.DecisionContext(ctx.Request().Context(), privacy.Allow) // bypass privacy policy
+	allowCtx = contextx.With(allowCtx, auth.OrgSubscriptionContextKey{})        // bypass org owned interceptor
```
golanglemonade: We shouldn't overload the `OrgSubscriptionContextKey` to bypass the org interceptor for this - we wan't its own context key - we currently define these in the `iam` repo in `auth/context.go`. This way we have more flexibility in bypass rules. (It has very much hurt me in the past to reuse the same key for very different things before 😭 ). This one in particular already is used in more places than just the org interceptor bypass we don't want to end up conflicting with later

internal/httpserve/route/acmesolver.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// registerAcmeSolverHandler registers the acme solver handler and route
+func registerAcmeSolverHandler(router *Router) (err error) {
+	path := "/.well-known/acme-challenge/:path"
+	method := http.MethodGet
+	name := "AcmeSolver"
+
+	route := echo.Route{
+		Name:   name,
```
golanglemonade, line 16: just confirming; this route doesn't need authentication?

acookin: nope. This should respond with the value cloudflare expects entirely unauthenticated.

internal/httpserve/handlers/acmesolver.go
acookin: ah sounds good. sorry, i just kinda followed a pattern to see if it would work and then forgot about it heh

go.mod
```diff
@@ -76,7 +76,7 @@ require (
 	github.com/theopenlane/entx v0.7.2
 	github.com/theopenlane/gqlgen-plugins v0.6.1
 	github.com/theopenlane/httpsling v0.2.2
-	github.com/theopenlane/iam v0.12.5
+	github.com/theopenlane/iam v0.12.6-0.20250527134840-519b23d83421
```
golanglemonade: I tagged this one too, `v0.12.6`

golanglemonade (approved):

PR #778, opened 2025-05-27
Scheduled job fixes

adelowo:
Bunch of tiny issues I have found while doing real world integrations.

- runs table now has an `execution_time_at`,  the agents might receive jobs earlier so they can use this to determine when to execute locally. Logic is here, we might ___pre schedule___ some jobs instead of having the agent ping way too often ( see https://github.com/theopenlane/riverboat/pull/106 )
- `runs` table now has a script column which is the templated version from that of the original job and the configuration you provided. see https://github.com/theopenlane/riverboat/pull/106 too 
- `script` should not be optional on the scheduled job.
- fixed: `Cron` model causing infinite recursion when retrieving from the db because of the `Value` method
- Since cadence and cron can be empty and valid, make sure to skip validation for them if empty


internal/ent/schema/scheduled_job.go
```diff
@@ -69,8 +69,7 @@ func (ScheduledJob) Fields() []ent.Field {
 						entgql.SkipWhereInput,
 				),
 			).
-			Comment("the script to run").
-			Optional(),
+			Comment("the script to run"),
```
golanglemonade, line 73: Why is this required? Won't this mean a customer couldn't create a job to track what they run on their own infrastructure and just send result without giving us a script even if they don't need to? 

We don't necessarily need/want the script to run in a lot of cases. Even in the case where its something we create - I think we would end up with a lot of duplicate data here instead of just storing a pointer to what would be executed?

Also, wondering if this should be a `TEXT` field instead of `String`?

golanglemonade (approved):

PR #790, opened 2025-05-30
feat: control mapping  - additional fields, tests, and permissions

golanglemonade:
- updates mapped control schema to have `to` and `from` controls and subcontrols, instead of just a single list
- adds confidence, source, and enums for mapping type for mapped controls
- adds cli commands for control mapping:

```
go run cmd/cli/main.go mapped-control create --to-control-ids 01JWSY84RYP4078SNECD3ZBNKJ --from-control-ids 01JWSY8545PSG8TQ8PAM74EQZA --confidence 84 --relation "example mapping" --mapping-type intersect         
  ID                          TOCONTROLS  FROMCONTROLS  RELATION         CONFIDENCE  MAPPINGTYPE  SOURCE  
  01JWSYAW5C5WYRSZ9DFM7NJEM8  A1.1        C1.1          example mapping  84%         INTERSECT    MANUAL  
```

```
go run cmd/cli/main.go mapped-control update -i 01JWSYAW5C5WYRSZ9DFM7NJEM8 -c 87
  ID                          TOCONTROLS  FROMCONTROLS  RELATION         CONFIDENCE  MAPPINGTYPE  SOURCE  
  01JWSYAW5C5WYRSZ9DFM7NJEM8  A1.1        C1.1          example mapping  87%         EQUAL        MANUAL  
```

```
go run cmd/cli/main.go mapped-control delete -i 01JWSYAW5C5WYRSZ9DFM7NJEM8 
  DELETEDID                   
  01JWSYAW5C5WYRSZ9DFM7NJEM8  
```
- adds command to fga task file to seed all the tests tuples for easy checking
- updates fga model for mapped controls, cleans up some other relations
- updates `controls` to be `org owned` instead of requiring access to a program
- similarly, mapped controls will be org owned, so org admins can create (or defined groups) but all users will be able to see the relations
- updates the parent tuple functions to work for `Organization` based on the `OwnerID` field, this is used for controls because it is both org owned and program inherited (for program admins to be able to edit controls)
- adds a `reference_framework` field to controls and subcontrols, this will allow us to track the framework the controls came from (in conjunction with the source of type `FRAMEWORK` - more on this on a follow-up PR
- removes the clone of mapped controls from clone controls resolver - need to ensure we can grab the correct controls if this is done from a standard -> org so this will be a follow-up PR


adelowo (dismissed):

adelowo (approved):

internal/ent/hooks/authzmutationhelpers.go
```diff
@@ -18,6 +18,10 @@ import (
 	"github.com/theopenlane/core/internal/ent/privacy/utils"
 )
 
+const (
+	ownerFieldName = "owner_id"
```
matoszz, line 22: any reason to re-define this instead of using the one in the org owned mixin ?

golanglemonade: went back and forth on this; wasn't sure I wanted an exported variable from either of these - but could define it in the schema package

golanglemonade: oh thats where the mixin is; I'll update :-D

golanglemonade: now I remember; I can't because thats a cyclical import - because schema imports hooks. I'd have to define it somewhere else completely

PR #796, opened 2025-06-01
Evidence: creation date cannot be in the future

adelowo:
- made sure the `creation_date` cannot be in the future
- validate if present only for updates but hard requirement on new creations
- updated tests


golanglemonade (approved):
good catch!

PR #810, opened 2025-06-06
org members should view other user's avatars in the same org

adelowo:
- [x] added fga tests 
- [x] updated models
- [x] updated permissions

### DIFF
--- a/internal/middleware/objects/permissions.go
+++ b/internal/middleware/objects/permissions.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/theopenlane/echox/middleware/echocontext"
+	"github.com/theopenlane/iam/auth"
 	"github.com/theopenlane/iam/fgax"
 
 	"github.com/theopenlane/core/internal/ent/privacy/utils"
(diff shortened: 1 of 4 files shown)


internal/middleware/objects/permissions.go
```diff
@@ -37,7 +38,26 @@ func AddFilePermissions(ctx context.Context) error {
...
+					return err
+				}
+
+				orgReq := fgax.GetTupleKey(fgax.TupleRequest{
+					SubjectID:   orgID + "#member",
+					SubjectType: "organization",
+					ObjectID:    f.ID,
+					ObjectType:  "file",
+					Relation:    "can_view",
+				})
```
golanglemonade: I realize the lines a few above aren't doing this, but we should use the defined constants, also should use the `SubjectRelation` field instead:
```suggestion
				orgReq := fgax.GetTupleKey(fgax.TupleRequest{
					SubjectID:   orgID,
					SubjectType: generated.TypeOrganization,
				        SubjectRelation: fgax.MemberRelation,
					ObjectID:    f.ID,
					ObjectType:  generated.TypeFile,
					Relation:    fgax.CanView
				})
```

internal/middleware/objects/permissions.go
```diff
@@ -37,7 +38,26 @@ func AddFilePermissions(ctx context.Context) error {
 				Relation:    "parent", // this will always be parent in an object owned permission setup
 			})
 
-			if _, err := utils.AuthzClientFromContext(ctx).WriteTupleKeys(ctx, []fgax.TupleKey{req}, nil); err != nil {
+			tuples := []fgax.TupleKey{req}
+
+			// if the parent of the file is a user, explicitly add view permissions for org members
+			if f.Parent.Type == "User" {
```
golanglemonade: this could cause issues if other file types we don't want public to the org are added. Can we also ensure this is an `avatarFile`?

internal/middleware/objects/permissions.go
adelowo: resolved in https://github.com/theopenlane/core/pull/810/commits/dc66ffcc3eef0b1577beebb9b55ccdedcb7eac29

internal/middleware/objects/permissions.go
adelowo: ha true. thanks for the callout

golanglemonade (approved):

PR #814, opened 2025-06-07
add CSRF wrapper

matoszz:
The core repository’s session cookies are currently configured with `SameSiteStrictMode`, which blocks cross‑site requests (e.g. blocks us trying to use Vercel's preview functionality). Our current server config provide CORS and standard security headers, but there is no code implementing CSRF tokens or verifying Origin/Referer headers. Changing the cookie to `SameSite=None` (or `lax`) would therefore remove the primary CSRF protection without any alternative mitigation in place, so this middleware is being added in advance of modification of the IAM repo / Cookie configuration to ensure we have adequate CSRF protection outside of the secure cookie `SameSite` config, so that we can loosen up the `SameSite` setting to `lax` and allow use of Vercel's preview functionality.


golanglemonade (approved):
sweet, this was what I was thinking we should do here as well 🚀

PR #816, opened 2025-06-07
Feat cert rotation

matoszz:
- Adds keydir and key watcher configurations and routines
- additionally adds with secretmanagerkeys which can call GCP secret manager directly


internal/httpserve/serveropts/keydir.go
```diff
@@ -0,0 +1,47 @@
...
+		maps.Copy(keys, s.Config.Settings.Auth.Token.Keys)
+
+		s.Config.Settings.Auth.Token.Keys = keys
+	})
+}
+
+// WithKeyDirFromEnv loads keys from the directory specified by the
+// CORE_AUTH_TOKEN_KEYDIR environment variable, if set
+func WithKeyDirFromEnv() ServerOption {
+	dir := os.Getenv("CORE_AUTH_TOKEN_KEYDIR")
```
golanglemonade: This should use the config instead of the env variable. This way it can provided via the yaml or env var like all other config vars.

internal/httpserve/serveropts/keydir.go
```diff
@@ -0,0 +1,47 @@
...
+		if err != nil {
+			log.Panic().Err(err).Msg("unable to read key directory")
+		}
+
+		keys := make(map[string]string)
+		for _, e := range entries {
+			if e.IsDir() || !strings.HasSuffix(e.Name(), ".pem") {
+				continue
+			}
+			kid := strings.TrimSuffix(e.Name(), filepath.Ext(e.Name()))
```
golanglemonade, line 29: If there is no `kid` found - e.g. no file without extension is found, we should:
1. Check if the config has the `kid` set in `auth.token.kid`
2. Error because the kid was not set and so you'd have an empty string in your map

internal/httpserve/serveropts/keydir_test.go
```diff
@@ -0,0 +1,79 @@
...
+				},
+			},
+		},
+	}
+}
+
+func TestWithKeyDirLoadsKeys(t *testing.T) {
+	dir := t.TempDir()
+	kid := ulid.Make().String()
+	writeKey(t, dir, kid)
```
golanglemonade, line 51: Add a test for when the `kid` isn't there

internal/httpserve/serveropts/keywatch.go
```diff
@@ -0,0 +1,86 @@
...
+			if err != nil {
+				log.Error().Err(err).Msg("unable to reload jwt keys")
+				return
+			}
+			s.Config.Handler.JWTKeys = keys
+			s.Config.Handler.TokenManager = tm
+		}
+
+		if _, err := os.Stat(dir); err != nil {
+			log.Error().Err(err).Msg("key directory not found")
```
golanglemonade: Can you add the `dir` to the log too? 
```suggestion
			log.Error().Str("dir", dir).Err(err).Msg("key directory not found")
```

internal/httpserve/serveropts/keydir.go
golanglemonade: This function doesn't look to be used, the `WithKeyDir` is called via `WithKeyDirWatcherFromEnv` instead?

internal/httpserve/serveropts/keywatch.go
```diff
@@ -0,0 +1,86 @@
...
+					log.Error().Err(err).Msg("watcher error")
+				}
+			}
+		}()
+	})
+}
+
+// WithKeyDirWatcherFromEnv calls WithKeyDirWatcher if CORE_AUTH_TOKEN_KEYDIR is set.
+func WithKeyDirWatcherFromEnv() ServerOption {
+	dir := os.Getenv("CORE_AUTH_TOKEN_KEYDIR")
```
golanglemonade: This should use the config instead of the env variable. This way it can provided via the yaml or env var like all other config vars.

internal/httpserve/serveropts/option.go
```diff
@@ -504,6 +504,32 @@ func WithSummarizer() ServerOption {
...
+		if s.Config.Settings.Server.SecretManagerSecret != "" {
+			WithSecretManagerKeys(s.Config.Settings.Server.SecretManagerSecret).apply(s)
+		}
+	})
+}
+
+// WithKeyDirWatcherOption allows the key directory watcher to be set via server config.
+func WithKeyDirWatcherOption() ServerOption {
+	return newApplyFunc(func(s *ServerOptions) {
+		if s.Config.Settings.Server.KeyDir != "" {
```
golanglemonade: empty if statement?

internal/httpserve/serveropts/keywatch.go
matoszz: When you say "use the config instead of the env variable" - there already is a config variable you can set with the key location, this function is in addition to that so that if you were manually rotating it you'd want to set this variable in *addition* to the other so that you could se the mount point on a running deployment and have it pick up the new setting. This is described in the readme

matoszz: I think I see a way to make this more managable - updating shortly!

internal/httpserve/serveropts/keydir.go
```diff
@@ -0,0 +1,46 @@
...
+				continue
+			}
+
+			kid := strings.TrimSuffix(e.Name(), filepath.Ext(e.Name()))
+			keys[kid] = filepath.Join(dir, e.Name())
+		}
+
+		// If no kid found, check config for kid
+		if len(keys) == 0 {
+			configKid := s.Config.Settings.Auth.Token.KID
```
golanglemonade, line 35: this `configKid ` isn't ever used besides to see if it was empty. If its found, and there are files in the key dir should we also set they keys? 

```
keys[configKid] = filepath.Join(dir, e.Name())
```

golanglemonade: Nevermind - chatted in slack - I was misunderstanding here

golanglemonade (dismissed):

golanglemonade (dismissed):

adelowo (approved):

PR #818, opened 2025-06-08
acookin/trustcenter

acookin:
- **trust center stuff**
- **stuff**
- **tc stuff**


internal/ent/schema/trustcenter_settings.go
```diff
@@ -0,0 +1,108 @@
...
+			Comment("primary color for the trust center").
+			Optional(),
+	}
+}
+
+// Mixin of the TrustCenterSetting
+func (t TrustCenterSetting) Mixin() []ent.Mixin {
+	return mixinConfig{
+		additionalMixins: []ent.Mixin{
+			newOrgOwnedMixin(t),
```
golanglemonade: I know this is still a draft, but thought I'd comment early on this - instead of having this as org owned, it should be "owned" by the trust center object. We have an annotation you can use, if you look at `groups` as an example, I think this should easily be able to work the same. And then you also don't need the edge to trust center settings on the organization object. 

```
		entfga.SettingsChecks("group"),
```

golanglemonade: You would set that ^ first, and then you can add this to your policy on a second run:

```
entfga.CheckEditAccess[*generated.TrustCenterSettingMutation](),
```

acookin: On the first generate, there's a failure with the authcheck dereferencing a string, and I honestly do not understand why. Not sure if i need to generate from scratch, but not even sure how to do that because if i delete the generated directory(s), i get massive errors

acookin: I retried this whole thing bc i was going around in circles, this seems to fail with different errors, so that's fun:
https://github.com/theopenlane/core/pull/822

PR #840, opened 2025-06-12
Trust Center APIs

acookin:
Create Trust Center and Trust Center Settings APIs. I expect we'll add and amend the settings, but this'll get us goin.

I plan on implementing anonymous users (issuing jwts for anonymous users) so that the "public" trust center content can be public, and this will also help us with user tracking.


internal/ent/schema/trustcenter.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Fields of the TrustCenter
+func (TrustCenter) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("slug").
+			Comment("Slug for the trust center").
+			MaxLen(trustCenterNameMaxLen).
+			Annotations(
+				entgql.Skip(entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
```
golanglemonade, line 56: Should this be `NotEmpty()`?

internal/graphapi/schema/trustcenter.graphql
```diff
@@ -0,0 +1,103 @@
...
+    ): TrustCenterBulkCreatePayload!
+    """
+    Create multiple new trustCenters via file upload
+    """
+    createBulkCSVTrustCenter(
+        """
+        csv file containing values of the trustCenter
+        """
+        input: Upload!
+    ): TrustCenterBulkCreatePayload!
```
golanglemonade: You shouldn't be allowed to bulk create trust centers, since each org should only have one. You should be able to remove this, and after regen the resolvers will be commented out to delete. 
```suggestion
    """
    Create multiple new trustCenters
    """
    createBulkTrustCenter(
        """
        values of the trustCenter
        """
        input: [CreateTrustCenterInput!]
    ): TrustCenterBulkCreatePayload!
    """
    Create multiple new trustCenters via file upload
    """
    createBulkCSVTrustCenter(
        """
        csv file containing values of the trustCenter
        """
        input: Upload!
    ): TrustCenterBulkCreatePayload!
```

internal/ent/schema/trustcenter.go
```diff
@@ -0,0 +1,124 @@
...
+func (t TrustCenter) Edges() []ent.Edge {
+	return []ent.Edge{
+		// Add relationships to other entities as needed
+		// Example: defaultEdgeToWithPagination(t, File{}),
+		uniqueEdgeTo(&edgeDefinition{
+			fromSchema: t,
+			edgeSchema: CustomDomain{},
+			field:      "custom_domain_id",
+			required:   false,
+			immutable:  false,
```
golanglemonade, line 84: This should cascade delete as well
```
			annotations: []schema.Annotation{
				entx.CascadeAnnotationField("TrustCenter"),
			},
```

internal/graphapi/schema/trustcentersetting.graphql
```diff
@@ -0,0 +1,103 @@
...
+    ): TrustCenterSettingBulkCreatePayload!
+    """
+    Create multiple new trustCenterSettings via file upload
+    """
+    createBulkCSVTrustCenterSetting(
+        """
+        csv file containing values of the trustCenterSetting
+        """
+        input: Upload!
+    ): TrustCenterSettingBulkCreatePayload!
```
golanglemonade: Same here
```suggestion
```

acookin:
sorry for not just adding another commit. I couldn't get the go generate to do its thing cleanly with a pure rebase, so I kinda had to start over, and pull my changes in and run generate a few times to get this to work. I feel like i am surely holding this incorrectly, but rebasing this to main took me far too long.

internal/ent/schema/trustcenter.go
acookin: NotEmpty forces this to be provided in the graphql input I'm pretty sure, and I'm autogenerating this in the hook. Is there a way to say "do not require for graphql but do require in DB"? I have searched for this in ent and could not figure it out if it does exist.

internal/ent/schema/trustcenter.go
acookin: I get a really fun error with this:
```
6:28PM FTL internal/ent/generate/entc.go:175 | generating history schema error=failed to generate template: failed loading ent graph: entc/load: # github.com/theopenlane/core/internal/ent/generated
internal/ent/generated/edge_cleanup.go:963:78: undefined: customdomain.HasTrustCenterWith
internal/ent/generated/edge_cleanup.go:964:90: undefined: customdomain.HasTrustCenterWith
```
I've added edges in various formats and lost a lot of time.
I think getting this branch in will make the iteration process much easier, every time i need to update my branch it takes a _very_ long time to regenerate all the code correctly.

matoszz (approved):

internal/ent/schema/trustcenter.go
golanglemonade: hm, it shouldn't. `NotEmpty` is just a DB validator (whereas when the field is required its in the graphql input as a requirement`

golanglemonade: Pretty sure `NotEmpty` is what does that for you - it's a validator on the db level; I have a comment on the `ownerID` field, which is not normally a required input for GraphQL, but required in the database:

```
			field.String(ownerFieldName).
				Comment("the ID of the organization owner of the object").
				Immutable(). // Immutable because it is set on creation and never changes
				Optional().  // Optional because it doesn't need to be provided as input
				NotEmpty())  // NotEmpty because it is required to be set in the database
```

PR #846, opened 2025-06-13
Allow adding a user to an org on sign up if they provide the invitation token

adelowo:
frontend at https://github.com/theopenlane/openlane-ui/pull/599

### DIFF
--- a/internal/httpserve/handlers/invite.go
+++ b/internal/httpserve/handlers/invite.go
@@ -62,102 +62,8 @@ func (h *Handler) OrganizationInviteAccept(ctx echo.Context) error {
 		return h.BadRequest(ctx, err)
 	}
 
-	inv := &Invite{
-		Token: in.Token,
-	}
-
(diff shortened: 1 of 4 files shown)


internal/httpserve/handlers/invite.go
```diff
@@ -62,28 +62,60 @@ func (h *Handler) OrganizationInviteAccept(ctx echo.Context) error {
...
+
 	inv := &Invite{
-		Token: in.Token,
+		Token: invitationToken,
 	}
 
 	// ensure the user that is logged in, matches the invited user
 	if err := inv.validateInviteRequest(); err != nil {
-		return h.BadRequest(ctx, err)
+		return nil, nil, nil, err
```
golanglemonade, line 102: this is so many returns, but I think refactoring the related things isn't worth it 😅

golanglemonade (approved):
🚀

PR #847, opened 2025-06-13
validate token prior to creating user on registration route

adelowo:
- **validate token before creating user**
- **simplify check**

### DIFF
--- a/internal/httpserve/handlers/register.go
+++ b/internal/httpserve/handlers/register.go
@@ -56,7 +56,7 @@ func (h *Handler) RegisterHandler(ctx echo.Context) error {
 			return h.BadRequest(ctx, err)
 		}
 
-		if !strings.EqualFold(strings.ToLower(invitedUser.Recipient), strings.ToLower(input.Email)) {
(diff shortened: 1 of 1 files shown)


internal/httpserve/handlers/register.go
```diff
@@ -45,6 +47,24 @@ func (h *Handler) RegisterHandler(ctx echo.Context) error {
...
+	if in.Token != nil {
+		ctxWithToken = token.NewContextWithOrgInviteToken(ctxWithToken, *in.Token)
+
+		invitedUser, err := h.getUserByInviteToken(ctxWithToken, *in.Token)
+		if err != nil {
+			log.Error().Err(err).Msg("error retrieving invite token")
+			return h.BadRequest(ctx, err)
+		}
+
+		if !strings.EqualFold(strings.ToLower(invitedUser.Recipient), strings.ToLower(input.Email)) {
```
golanglemonade: you shouldn't need `strings.ToLower(` when using `.EqualFold`:

```suggestion
		if !strings.EqualFold(invitedUser.Recipient, input.Email) {
```

golanglemonade (approved):

PR #851, opened 2025-06-15
Graceful Shutdown of DB Clients, Redis Clients, Event Pools

matoszz:
During testing and validation of running Openlane on Kubernetes, I noticed strange behavior with how pods were being treated during rolling updates and node evictions. After some testing, it was evident that liveness and readiness endpoints always returned a 200 even if a shutdown was in progress. Event pools were created independently across the codebase with no centralized mechanism for closing them. The database drivers also seemingly did not block new queries once shutdown began and the shutdown helper could only break out of the `select` statement potentially looping forever. This PR attempts to address those concerns with the following changes:

- `Execute()` now creates a context using `signal.NotifyContext` and logs when the context finishes, enabling graceful SIGTERM handling
- the server closes the redis client, database client, and all registered event pools when the context is cancelled
- New blocking drivers, hooks, and interceptors prevent new queries once the shutdown process starts, and `GracefulCloseWithFlag` waits for in-flight operations before closing connections with a labeled loop
- Liveness and Readiness routes now return `503` if shutdown is underway
- Event pools (in soiree) are registered globally so `ShutDownAll` can close them together during shutdown
- the metrics server starts with a custom http.Server that shuts down on context cancellation and logs shutdown errors

With these updates, when Kubernetes sends SIGTERM, the service logs “shutting down gracefully…”, stops accepting new database queries, waits for running queries to finish, closes event pools, the metrics server, Redis and database connections, and reports non‑ready status via /livez and /ready. The system therefore terminates cleanly without disrupting in‑flight work.

The behavior prior this PR / on current main was easily witnessed when running the service locally, and hitting `ctrl + c` :
```
^Ctask: Signal received: "interrupt"
signal: interrupt
task: Failed to run task "run-dev": exit status 1
```

AFTER:

```
^C10:29AM INF cmd/root.go:42 | shutting down gracefully...
task: Signal received: "interrupt"
10:29AM ERR cmd/serve.go:196 | failed to run server error=http: Server closed
task: Failed to run task "run-dev": exit status 1
```

When the server receives `SIGTERM` the context created in `cmd/root.go` is cancelled, which signals the HTTP server to stop. `StartEchoServer` then returns `http.ErrServerClosed`, indicating a graceful shutdown.

```
if err := srv.StartEchoServer(ctx); err != nil {
        log.Error().Err(err).Msg("failed to run server")
}
```
Any graceful shutdown should have resulted in the log line `... failed to run server error=http: Server closed`. It occurs because `Start` returns `http.ErrServerClosed` after `Shutdown` completes, so the _lack_ of this error today means effectively that we were not actually gracefully shutting the server down as we may have thought.

### DIFF
--- a/internal/entdb/shutdown.go
+++ b/internal/entdb/shutdown.go
@@ -2,98 +2,174 @@ package entdb
 
 import (
 	"context"
-	"database/sql"
+	stdsql "database/sql"
 	"errors"
+	"fmt"
 	"sync/atomic"
 	"time"
 
+	"ariga.io/entcache"
 	"entgo.io/ent/dialect"
+	entsql "entgo.io/ent/dialect/sql"
 
 	ent "github.com/theopenlane/core/internal/ent/generated"
 	"github.com/theopenlane/core/internal/ent/generated/intercept"
 )
 
-var (
-	// ErrShuttingDown is returned when operations are attempted during shutdown
-	ErrShuttingDown = errors.New("database shutting down")
-	// ErrDriverSupportBeginTx is returned when the driver does not support BeginTx
-	ErrDriverSupportBeginTx = errors.New("driver does not support BeginTx")
-	// shuttingDown is set to true when GracefulClose is called
-	shuttingDown atomic.Bool
-)
+// ShutdownFlag tracks whether a shutdown is in progress
+type ShutdownFlag struct {
+	// flag is an atomic boolean that indicates if a shutdown is in progress
+	// atomic.Bool is used to ensure thread-safe access to the flag
+	// without the need for explicit locks
+	// this is important in a concurrent environment where multiple goroutines
+	// may be checking or setting the flag at the same time
+	flag atomic.Bool
+}
 
-// BeginShutdown marks the system as shutting down. It is safe to call multiple times.
+// Begin sets the shutdown flag to true, indicating that a shutdown is in progress
+func (s *ShutdownFlag) Begin() {
+	s.flag.Store(true)
+}
+
+// Reset clears the shutdown flag, indicating that a shutdown is no longer in progress
+func (s *ShutdownFlag) Reset() {
+	s.flag.Store(false)
+}
+
+// IsSet checks if the shutdown flag is set to true
+func (s *ShutdownFlag) IsSet() bool {
+	return s.flag.Load()
+}
+
+// newShutdownFlag creates a new instance of ShutdownFlag
+func newShutdownFlag() *ShutdownFlag {
+	return &ShutdownFlag{}
+}
+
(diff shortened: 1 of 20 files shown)


internal/entdb/shutdown.go
```diff
@@ -0,0 +1,252 @@
...
+	if c == nil {
+		return nil
+	}
+
+	return extractSQLDB(c.Driver())
+}
+
+// extractSQLDB is a utility designed to retrieve the underlying *sql.DB
+// (aliased here as *stdsql.DB) from a potentially wrapped Ent database driver
+func extractSQLDB(d dialect.Driver) *stdsql.DB {
```
golanglemonade, line 77: This function already exists in the `generated` code as `.DB()`, any reason you can't just us the `.DB()` function here?

https://github.com/theopenlane/core/blob/main/internal/ent/generate/templates/ent/client.tmpl#L20

internal/graphapi/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/99designs/gqlgen/graphql/handler"
+
+	"github.com/theopenlane/core/pkg/metrics"
+)
+
+// WithMetrics adds Prometheus instrumentation around GraphQL operations.
+func WithMetrics(h *handler.Server) {
+	h.AroundOperations(func(ctx context.Context, next graphql.OperationHandler) graphql.ResponseHandler {
+		opCtx := graphql.GetOperationContext(ctx)
```
golanglemonade, line 16: it should be fine since this is used only for graph requests, but just an FYI `GetOperationContext` will `panic` if the context isn't set and `HasOperationContext` can be called before if there is a chance this would be called without a graphql context.

internal/graphapi/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+		opName := "unknown"
+		if opCtx != nil {
```
golanglemonade: `GetOperationContext` will never be `nil`; it would have panic'ed if that was the case.

internal/entdb/shutdown.go
matoszz: I wanted to ensure if we added other drivers to connect to different postgreSQL instances I could reuse the same methods to ensure connections were drained

matoszz: additionally, as I said in our slack thread, `DB()` is missing a key part of functionality required to implement this logic which is `blockingDriver` and the receiver functions it uses. Leveraging `DB()` in its current state, from everything I can tell, would not be possible (or at least not without a wholesale refactor for this PR which I'm personally unwilling to entertain)

golanglemonade (approved):

PR #862, opened 2025-06-17
Module, Product, and Price schemas for Entitlements

matoszz:
- Adds schemas which roughly match stripe's model, where you have 1 customer with 1 subscription, but with many products and with many prices attached to each product
- Kept the fields relatively basic but built under the context of "module aware"
- Added annotations using the new entx helper created to label schemas with which module they are related to

This PR is schema changes only, so all changes were derived from `internal/ent/schema` - further automation for Hooks, Interceptors, and Policies will be added in later PR's


internal/ent/schema/customdomain.go
```diff
@@ -124,9 +126,17 @@ func (CustomDomain) Policy() ent.Policy {
...
 
+// Hooks of the CustomDomain
 func (CustomDomain) Hooks() []ent.Hook {
 	return []ent.Hook{
 		hooks.HookCreateCustomDomain(),
 		hooks.HookDeleteCustomDomain(),
 	}
 }
+
+// Annotations of the DNSVerification
```
golanglemonade:
```suggestion
// Annotations of the CustomDomain
```
```suggestion
// Annotations of the DNSVerification
```

internal/ent/schema/mappable_domain.go
```diff
@@ -98,3 +100,10 @@ func (MappableDomain) Policy() ent.Policy {
 func (MappableDomain) Hooks() []ent.Hook {
 	return []ent.Hook{}
 }
+
+// Annotations of the Invite
```
golanglemonade:
```suggestion
// Annotations of the MappableDomain
```

internal/ent/schema/organization.go
```diff
@@ -213,6 +213,21 @@ func (o Organization) Edges() []ent.Edge {
 			Annotations(
 				entx.CascadeAnnotationField("Owner"),
 			),
+		edgeToWithPagination(&edgeDefinition{
+			fromSchema:         o,
+			edgeSchema:         OrgProduct{},
+			cascadeDeleteOwner: true,
```
golanglemonade, line 219: I believe we need an annotation on each of these (and I'm not sure if we have this every where we should today, I just noticed it in your output) so that you can't do the update via the [organization](https://github.com/theopenlane/core/pull/862/files) resolvers:

```
Annotations(
    entgql.Skip(entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
)
```

golanglemonade: I'm going to follow-up and make sure our other schemas don't also have this issue.

golanglemonade (approved):

PR #866, opened 2025-06-18
users should be able to leave orgs themselves

adelowo:
- **Revert "fix: hard delete needs to delete tuples first (#857)"**
- **allow user self delete themselves from an org**
- **fix bug where after deletion, random orgs are just assigned as the default**
- **update tests to make sure the user actually leaves the org**

### DIFF
--- a/internal/ent/hooks/memberhelpers.go
+++ b/internal/ent/hooks/memberhelpers.go
@@ -2,6 +2,8 @@ package hooks
 
 import (
 	"context"
+	"fmt"
+	"slices"
 	"strings"
 
 	"entgo.io/ent"
@@ -9,6 +11,7 @@ import (
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/rs/zerolog"
 	"github.com/theopenlane/iam/auth"
+	"github.com/theopenlane/iam/fgax"
 
 	"github.com/theopenlane/core/internal/ent/generated"
 	"github.com/theopenlane/core/internal/ent/generated/privacy"
@@ -20,6 +23,8 @@ type MutationMember interface {
 	UserID() (string, bool)
 	ID() (string, bool)
 	IDs(ctx context.Context) ([]string, error)
+	Op() ent.Op
+	Client() *generated.Client
 }
 
 // HookMembershipSelf is a hook that runs on membership mutations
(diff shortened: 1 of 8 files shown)


internal/ent/hooks/memberhelpers.go
```diff
@@ -59,17 +64,58 @@ func HookMembershipSelf(table string) ent.Hook {
...
+			if err != nil {
+				return nil, err
+			}
+
+			retVal, err := next.Mutate(ctx, m)
+			if err != nil {
+				return nil, err
+			}
+
+			if table == "org_memberships" && (m.Op().Is(ent.OpDelete | ent.OpDeleteOne)) {
```
golanglemonade: I think you should be able to use `migrate.OrgMembershipsTable.Name` instead of the string `org_memberships` here. 

Does this issue not affect group and program memberships too or is it specific to _self_ removal?

internal/ent/hooks/organization.go
```diff
@@ -385,10 +386,17 @@ func checkAndUpdateDefaultOrg(ctx context.Context, userID string, oldOrgID strin
 	// if the user's default org was deleted this will now be nil
 	if userSetting.Edges.DefaultOrg == nil || userSetting.Edges.DefaultOrg.ID == oldOrgID {
 		// set the user's default org another org
-		// get the first org that was not the org being deleted
+		// get the first org that was not the org being deleted and where the user is a member
 		newDefaultOrgID, err := client.
 			Organization.
 			Query().
+			Where(
```
golanglemonade: doesn't the interceptor filter this already? or is it an order of operations because we use list objects from fga for organizations?

internal/graphapi/orgmembers_test.go
```diff
@@ -179,22 +180,18 @@ func TestQueryOrgMembers(t *testing.T) {
 }
 
 func TestMutationCreateOrgMembers(t *testing.T) {
-	t.Parallel()
-
-	testUser := suite.userBuilder(context.Background(), t)
```
golanglemonade: this is going to cause conflicts with the other tests, thats why I had created a brand new user for this test, any test thats touches the org level things cannot use the same seeded data.

internal/ent/hooks/memberhelpers.go
adelowo: it doesn't. all good there so far from tests

golanglemonade (approved):

PR #871, opened 2025-06-19
Bump gqlgen-plugins

adelowo:
Needed to cut a release and not so sure when rennovate bot would do that


adelowo:
> needs a `go mod tidy`

Done

PR #889, opened 2025-06-24
Add new schemas in anticipation of upcoming features

matoszz:
- The Assets, Scan schemas are being in anticipation of performing domain + vulnerability scanning and presenting to the user the ability to confirm technologies or assets we discover
- The Identity Provider fields in organization settings are being added in anticipation of adding org-level SSO
- The Webhook Secret field in organization setting is being added in anticipation of creating webhook endpoints capable of receiving payloads from 3d party systems and then performing updates on controls, evidence, tasks, etc.
- The Suggested and Dismissed tag fields are being added in anticipation of an AI-based integration capable of automatically tagging policies and procedures / presenting possible tags to the user
- The additional edges / relationships for OrgModules, OrgProduct, OrgPrice, and OrgSubscription are in anticipation for the module-based entitlement work
- Sorting on name for controlOwners was added related to UI table sorting

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -78,7 +78,7 @@ require (
 	github.com/theopenlane/entx v0.8.2
 	github.com/theopenlane/gqlgen-plugins v0.6.2
 	github.com/theopenlane/httpsling v0.2.2
-	github.com/theopenlane/iam v0.13.0
+	github.com/theopenlane/iam v0.12.7
 	github.com/theopenlane/newman v0.1.6
 	github.com/theopenlane/riverboat v0.1.6
 	github.com/theopenlane/utils v0.4.7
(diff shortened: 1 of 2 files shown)


internal/ent/schema/asset.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (Asset) Policy() ent.Policy {
+	return policy.NewPolicy(
+		policy.WithQueryRules(
+			rule.AllowQueryIfSystemAdmin(),
+			// object owner check done via scan
+			policy.CheckOrgReadAccess(),
+		),
+		policy.WithMutationRules(
+			rule.AllowMutationIfSystemAdmin(),
```
golanglemonade: this is generally used when we add system-owned objects - this this schema is org owned. Based on the schema though - this might be to add date on behalf of the organization?

golanglemonade (approved):

PR #917, opened 2025-06-27
anonymous JWTs for trust center users + allow view on trust centers

acookin:
Implements anonymous user sessions for accessing the "public" trust center info.

Creates a handler for issuing anonymous JWTs that are tied to a specific trust center, so we can track the users and filter the right content. We can later use this JWT and convert the user to a real User when they sign an NDA to access documents and stuff.

Depends on https://github.com/theopenlane/iam/pull/281


go.mod
```diff
@@ -80,7 +80,7 @@ require (
 	github.com/theopenlane/entx v0.8.2
 	github.com/theopenlane/gqlgen-plugins v0.6.2
 	github.com/theopenlane/httpsling v0.2.2
-	github.com/theopenlane/iam v0.13.0
+	github.com/theopenlane/iam v0.13.1-0.20250627192841-c29f1d849f7e
```
acookin: TODO: update when https://github.com/theopenlane/iam/pull/281 is merged/tagged

cmd/cli/cmd/client.go
```diff
@@ -28,6 +28,9 @@ func TokenAuth(ctx context.Context, k *koanf.Koanf) (*openlaneclient.OpenlaneCli
 	token := k.String("token")
 	if token == "" {
 		token = k.String("pat")
+		if token == "" {
+			token = k.String("jwt")
+		}
 	}
```
golanglemonade: not really you're code, but might be better instead of keeping nesting to refactor this to the following, so we can just extend for more later if we needed. ( indentation isn't correctly so wouldn't recommend just committing the suggestion)

```suggestion
var token string
for _, key := range []string{"token", "pat", "jwt"} {
	token = k.String(key)
	if token != "" {
		break
	}
}
```

internal/httpserve/authmanager/authmanager.go
```diff
@@ -84,6 +85,40 @@ func (a *Client) GenerateUserAuthSession(ctx context.Context, w http.ResponseWri
 	return a.GenerateUserAuthSessionWithOrg(ctx, w, user, "")
 }
 
+func (a *Client) GenerateAnonymousTrustCenterSession(ctx context.Context, w http.ResponseWriter, targetOrgID string, targetTrustCenterID string) (*models.AuthData, error) {
```
golanglemonade, line 89: comment exported function 🙏

internal/httpserve/handlers/trustcenter_anonymousjwt.go
```diff
@@ -0,0 +1,95 @@
...
+		if len(pathSegments) == 0 || pathSegments[0] == "" {
+			return h.BadRequest(ctx, ErrMissingSlugInPath)
+		}
+		slug := pathSegments[0]
+
+		// 4a. query the database for trust centers with the slug and the default hostname
+		trustCenter, err = h.DBClient.TrustCenter.Query().
+			Where(trustcenter.SlugEQ(slug)).
+			Where(trustcenter.Not(trustcenter.HasCustomDomain())).
+			First(allowCtx)
```
golanglemonade: there should never be more than one right? Thoughts on use `Only` instead here and line 76?

internal/httpserve/route/trustcenter_anonymousjwt.go
```diff
@@ -0,0 +1,29 @@
...
+// registerTrustCenterAnonymousJWTHandler registers the trust center anonymous JWT handler and route
+func registerTrustCenterAnonymousJWTHandler(router *Router) (err error) {
+	path := "/trustcenter/auth/anonymous"
+	method := http.MethodPost
+	name := "TrustCenterAnonymousJWT"
+
+	route := echo.Route{
+		Name:   name,
+		Method: method,
+		Path:   path,
```
golanglemonade: Did you intentionally leave out the default middleware here from any reason?  (default here is the transaction mw, recovery middleware, and csrf middleware)

```suggestion
		Path:   path,
		Middlewares: mw,
```

pkg/models/models.go
```diff
@@ -979,3 +979,7 @@ var ExampleJobRunnerRegistrationResponse = JobRunnerRegistrationReply{
 type AcmeSolverRequest struct {
 	Path string `param:"path" description:"The path to the acme challenge" example:"01J4HMNDSZCCQBTY93BF9CBF5D"`
 }
+
+type CreateTrustCenterAnonymousJWTResponse struct {
```
golanglemonade, line 984: comment exported struct

internal/httpserve/route/trustcenter_anonymousjwt.go
acookin: nope, I think i was having random troubles and removed it, and never added back

cmd/cli/cmd/client.go
acookin: wowwww can't even do correct tabbing in suggested code 😆

go.mod
golanglemonade: tagged at v0.13.3

internal/httpserve/handlers/trustcenter_anonymousjwt.go
acookin: yep yep. I think i just have too many ORM syntaxes cataloged and i can't keep it straight. First is before Only alphabetically haha

golanglemonade (approved):

PR #937, opened 2025-07-03
Migrate csv exports to background jobs

adelowo:
```
  createExport(input: $createExportInput3)  {
    export {
      id
      ownerID
    }
  }
```

Frontend can then query all exports. 

```
query {
  exports {
    edges {
      node {
        id
      }
    }
  }
}
```

then grab the file and download.

There is a job worker that uses the client to convert the controls to csv then saves to the file. I have also added a `format` - it supports only csv at this time


db/migrations/20250707142637_exports.sql
```diff
@@ -0,0 +1,12 @@
+-- Create "export_history" table
+CREATE TABLE "export_history" ("id" character varying NOT NULL, "history_time" timestamptz NOT NULL, "ref" character varying NULL, "operation" character varying NOT NULL, "created_at" timestamptz NULL, "updated_at" timestamptz NULL, "created_by" character varying NULL, "updated_by" character varying NULL, "deleted_at" timestamptz NULL, "deleted_by" character varying NULL, "owner_id" character varying NULL, "export_type" character varying NOT NULL, "status" character varying NOT NULL DEFAULT 'PENDING', "requestor_id" character varying NULL, "file_id" character varying NULL, PRIMARY KEY ("id"));
```
golanglemonade: I dont think we need a history table for this table

internal/ent/hooks/export.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/theopenlane/core/pkg/objects"
+	"github.com/theopenlane/iam/auth"
+)
+
+var (
+	errExportTypeInvalid     = errors.New("export type is invalid")
+	errExportTypeNotProvided = errors.New("provide export type")
+)
+
+func HookExport() ent.Hook {
```
golanglemonade, line 21: comment exported function 🙏

pkg/enums/exporttype.go
```diff
@@ -0,0 +1,56 @@
...
+	"io"
+	"strings"
+)
+
+// ExportType is a custom type representing the various states of ExportType.
+type ExportType string
+
+var (
+	// ExportTypeControl indicates the control.
+	ExportTypeControl ExportType = "CONTROL"
```
golanglemonade, line 14: does this mean we need to manually add all types? This should work for any object type; or it should be automatic opt-in with a schema annotation?

internal/ent/schema/export.go
```diff
@@ -0,0 +1,118 @@
...
+const SchemaExport = "export"
+
+func (Export) Name() string       { return SchemaExport }
+func (Export) GetType() any       { return Export.Type }
+func (Export) PluralName() string { return pluralize.NewClient().Plural(SchemaExport) }
+
+// Fields returns export fields.
+func (Export) Fields() []ent.Field {
+	return []ent.Field{
+		field.Enum("export_type").
```
golanglemonade, line 34: We need to also be able to set the fields + filters; looks like right now it will just grab all data of that type?

pkg/corejobs/export.go
```diff
@@ -0,0 +1,228 @@
...
+	CategoryID         string `csv:"category_id"`
+	Subcategory        string `csv:"subcategory"`
+	MappedCategories   string `csv:"mapped_categories"`
+	Tags               string `csv:"tags"`
+	ControlQuestions   string `csv:"control_questions"`
+	CreatedAt          string `csv:"created_at"`
+	UpdatedAt          string `csv:"updated_at"`
+	CreatedBy          string `csv:"created_by"`
+	UpdatedBy          string `csv:"updated_by"`
+}
```
golanglemonade: this shouldn't be hardcoded; we want the request to be able to set which fields they want (And any filters). If we need to, we should be able to add `csv` struct tags with gqlgen, so we have the tags there we need. 

https://github.com/theopenlane/core/blob/main/internal/graphapi/generate/.gqlgen.yml#L18

fga/model/model.fga
```diff
@@ -485,6 +485,18 @@ type trust_center_setting
...
+  relations
+    define can_view: [user, service] or can_edit or (viewer but not blocked)
+    define can_edit: [service] or system_admin from system
+    define can_delete: [service] or system_admin from system
+    define audit_log_viewer: ([user, service] or audit_log_viewer from parent) and can_view
+    define parent: [organization]
+    define system: [system]
+    define viewer: [group#member] or editor or can_view from parent
+    define editor: [group#member] or can_edit from parent
+    define blocked: [user, group#member]
```
golanglemonade: these aren't quite right and would allow users in the organization to see exported data they might not be allowed to; only the user or service that created the export should be able to view it.

golanglemonade (approved):

PR #944, opened 2025-07-04
feat: optimize clone controls

golanglemonade:
> because extending the gateway timeout is not only not an option, but because performance matters here

tl;dr; - in prod, the batch checks were piling up; and timing out so the controls never even started to be cloned before we hit 30 seconds. I want to follow-up and remove some other duplicate checks to fga I've seen for awhile that should have an overall performance improvement (e.g. dont check the organization using listobjects more than once in a call, etc)

- adds a pond pool to the ent client; this was setup to look at using for the ent history, but that didn't end up being the bottleneck. leaving here for now because I plan on optimizing this as well but not a blocker for cloning controls
- optimizing `Get()` queries used in hooks to only pool field required, instead of all fields
- allows reference framework to be set on create, eliminating an additional query to the control when creating controls
- bypass bulk checks when we know the user has permissions to the object (e.g. if the user can see a standard ,they have permissions to all related controls under the standard.
- added additional client functions to return less data after clone
- by default, all requests will use the optimize latency method - checking the cache over the database (ttl I was testing with is 60s)  - we may need to adjust this for certain queries - where `CheckAccessHighConsistency` can be used over `CheckAccess()`
- **NOTE**: a lot of the changes are in this file, which github hides by default because it has a lot of changes and is "generated": https://github.com/theopenlane/core/pull/944/files#diff-71922d8b0335689ea9b83e9200083e7a33c3d366694022abafd502e10a6ed356

tested locally:

```
📄  Creating controls from seed data...
> creating controls...  10% [>              ]  [0s:0s]
{"level":"error","app":"openlane-startup","time":"2025-07-05T21:31:37-06:00","message":"starting to create controls"}
{"level":"warn","app":"openlane-startup","standard":"ISO 27001","time":"2025-07-05T21:31:38-06:00","message":"creating program for standard"}
{"level":"warn","app":"openlane-startup","standard":"ISO 27001","time":"2025-07-05T21:31:38-06:00","message":"creating controls by clone"}
{"level":"warn","app":"openlane-startup","standard":"NIST 800-53","time":"2025-07-05T21:31:39-06:00","message":"creating program for standard"}
{"level":"warn","app":"openlane-startup","standard":"NIST 800-53","time":"2025-07-05T21:31:39-06:00","message":"creating controls by clone"}
{"level":"warn","app":"openlane-startup","standard":"NIST CSF","time":"2025-07-05T21:31:47-06:00","message":"creating program for standard"}
{"level":"warn","app":"openlane-startup","standard":"NIST CSF","time":"2025-07-05T21:31:47-06:00","message":"creating controls by clone"}
{"level":"warn","app":"openlane-startup","standard":"SOC 2","time":"2025-07-05T21:31:47-06:00","message":"creating program for standard"}
{"level":"warn","app":"openlane-startup","standard":"SOC 2","time":"2025-07-05T21:31:47-06:00","message":"creating controls by clone"}
{"level":"error","app":"openlane-startup","time":"2025-07-05T21:31:49-06:00","message":"finished creating controls"}
> seeded environment created 100% [===============]  [11s]Sat Jul  5 21:31:49 MDT 2025
``` 

Also tested in prod, but had some temp optimizations around history inserts - I think we should be good for now without them so want to get this out before moving onto that


Taskfile.yaml
```diff
@@ -168,7 +168,7 @@ tasks:
     silent: true
     aliases: ['testsum:ci']
     cmds:
-      - go run gotest.tools/gotestsum@latest --junitfile junit.xml --rerun-fails --packages "./..." --format-hide-empty-pkg --rerun-fails --format-icons hivis --hide-summary output -- -p 20 -coverprofile=coverage.out
+      - go run gotest.tools/gotestsum@latest --junitfile junit.xml --packages "./..." --format-hide-empty-pkg --rerun-fails-run-root-test --format-icons hivis --hide-summary output -- -p 20 -coverprofile=coverage.out
```
golanglemonade, line 171: test
```suggestion
      - go run gotest.tools/gotestsum@latest --junitfile junit.xml --packages "./..." --format-hide-empty-pkg --rerun-fails-run-root-test --format-icons hivis --hide-summary output -- -p 5 -coverprofile=coverage.out
```

matoszz (approved):

PR #945, opened 2025-07-05
Soiree event persistence store(s) 

matoszz:
- Introduced an `EventStore` interface and an optional `EventQueue` extension for persisting events and results
- Updated `EventPool` to support pluggable event stores, background queue consumption, and listener retries with customizable backoff strategies
- Optional Redis store can be done per pool via `WithEventStore`, `WithRedisStore`, and `WithRetry`
- misc update; the event interface now uses `any `instead of `interface{}`
- docs update to include how to persist events in Redis and retry failed handlers using a backoff policy
- Metrics wrappers for prometheus good-ness


pkg/events/soiree/redis_store.go
```diff
@@ -0,0 +1,199 @@
...
+	// context is not needed here since this is a fire-and-forget operation
+	pushErr := s.client.RPush(context.Background(), "soiree:results", data).Err()
+	if pushErr == nil {
+		s.metrics.redisResultsPersisted.Inc()
+	}
+
+	return pushErr
+}
+
+// DequeueEvent pops a soiree event from the event queue - party line !
```
golanglemonade, line 102: conga line? 😆

golanglemonade (approved):

PR #947, opened 2025-07-05
Organization-wide SSO Configuration and Enforcement

matoszz:
This PR introduces Organization‑controlled single sign‑on (SSO) with OpenID Connect (OIDC). Documentation has been added at `docs/SSO.md` (likely to be integrated into `docs.theopenlane.io` at a later date) explains how admins configure providers such as Okta or Google Workspace and describes the WebFinger endpoint used by the login flow. Additionally overall diagrams / flows of the various login methods have been added for clearer understanding of the differences.

# Overview
- New REST handlers implement the OIDC flow. `SSOLoginHandler` builds the IdP redirect, while `SSOCallbackHandler` validates state/nonce, provisions the user, and issues a session. 
- WebFinger support is provided via `WebfingerHandler`, which returns the organization’s SSO policy and discovery URL
- Tokens can be authorized for SSO via dedicated endpoints; the handlers store authorization timestamps on the token record
- Middleware now checks if an organization enforces SSO and redirects non‑owner users to the IdP when unauthorized requests are detected
- Basic browser-initiated authentication flows have been added to the CLI to aid in authentication via the CLI (since if your org is set to a default org with SSO enabled and you are NOT an owner user but wish to use the CLI, it would basically be unusable without some method implemented).
- If a user leveraging social login providers is attempting to authenticate into an organization with SSO enforced, they are similarly redirected to the IdP for authentication
- A local dex instance + setup has been provided so that the process can be verified end to end
- Small HTML helpers / UIs have been added to `pkg/testutils` for testing of the code in this area

## Key assumptions to note:
- Assumes owner users can bypass the enforcement using credential based authentication
- The SSO initiated login flow occurs based on a check to the user's default organization, so membership in other organizations with SSO enforcement only occur at the time of a /switch call
- We make available to the UI (and other clients) the user or organizations SSO status via a .well-known webfinger endpoint
- The OIDC library (zitadel) is only implemented at this time specific to login intersections

## Caveats
- The PAT + API token authorization flows may not work 100% as it was difficult to test / verify the behavior without a UI and additional graphql queries / wrappers 
- Verification hooks on the org settings requiring client ID and client secret and discovery URL being present before allowing an org to flip the enforcement were added, but we should consider adding some kind of pre-verification (or staged method of allowing the user to authenticate with the settings they've added before performing org wide enforcement)
- No consideration / updates were made around existing issued sessions or tokens, so this may also need to be added in the future
- There are no checks / enforcements around what role a user must possess to make updates to these organization settings, so we should consider specific role designations in the future
- SSO is being considered a part of the "base" system functionality and not a separate feature we require authorization checks for


cmd/cli/cmd/login/root.go
```diff
@@ -39,6 +39,21 @@ func login(ctx context.Context) (*oauth2.Token, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		fmt.Println("\nAuthentication Successful!")
+		if err := cmd.StoreToken(tokens); err == nil {
+			fmt.Println("auth tokens successfully stored in keychain")
+		}
+
+		return tokens, nil
```
golanglemonade, line 54: We don't bubble up the error if the tokens aren't stored successfully, I also think we need to store the session cookie as well, similar to the password auth below? Something like this? 

```suggestion
		fmt.Println("\nAuthentication Successful!")
	        err = cmd.StoreToken(tokens)
	        cobra.CheckErr(err)
	        
	        cmd.StoreSessionCookies(client)
	       
		fmt.Println("auth tokens successfully stored in keychain")
		
		return tokens, nil
```

cmd/cli/cmd/login/sso.go
```diff
@@ -0,0 +1,165 @@
...
+	resp, err := client.HTTPSlingRequester().ReceiveWithContext(ctx, &out,
+		httpsling.Get("/.well-known/webfinger"),
+		httpsling.QueryParam("resource", fmt.Sprintf("acct:%s", email)),
+	)
+	if err != nil {
+		return out, err
+	}
+
+	if resp != nil {
+		resp.Body.Close()
```
golanglemonade, line 73: nit: doesn't look like you ever read the body; but usually doing a `defer` before ensures its available, as needed within the function, and always closing no matter how the function exists
```suggestion
		defer resp.Body.Close()
```

internal/httpserve/handlers/switch.go
```diff
@@ -47,6 +52,22 @@ func (h *Handler) SwitchHandler(ctx echo.Context) error {
...
+	allowCtx := privacy.DecisionContext(reqCtx, privacy.Allow)
+	status, err := h.fetchSSOStatus(allowCtx, in.TargetOrganizationID)
+
+	if err == nil && status.Enforced {
+		member, mErr := transaction.FromContext(allowCtx).OrgMembership.Query().Where(
+			orgmembership.UserID(user.ID),
+			orgmembership.OrganizationID(in.TargetOrganizationID),
+		).Only(allowCtx)
+		if mErr == nil && member.Role != enums.RoleOwner {
+			loginURL := sso.SSOLogin(ctx.Echo(), in.TargetOrganizationID)
```
golanglemonade, line 66: not sure how this will work with the UI; but thats a tomorrow problem

internal/httpserve/handlers/webauthn.go
```diff
@@ -380,3 +387,28 @@ func (h *Handler) userHandler(ctx context.Context) webauthn.DiscoverableUserHand
 		return authnUser, nil
 	}
 }
+
+func (h *Handler) HasValidSSOSession(ctx echo.Context, userID string) bool {
```
golanglemonade, line 391: does this need to be exported? looks like its only used within the `handlers` package? Either unexport it or add a comment 🙏

pkg/middleware/auth/auth.go
```diff
@@ -341,11 +362,156 @@ func getSubjectName(user *ent.User) string {
...
+		}
+	}
+
+	if jsonErr := c.JSON(http.StatusUnauthorized, rout.ErrorResponse(err)); jsonErr != nil {
+		log.Error().Err(jsonErr).Msg("failed to write unauthorized JSON response")
+		return jsonErr
+	}
+
+	return nil
+}
```
golanglemonade, line 397: nit: the amount of nested `ifs` makes this function really. hard to read, I would suggest flippingn the checks to make it easier to follow:

```
// returnUnauthorized returns a 401 Unauthorized response with the error message.
func returnUnauthorized(c echo.Context, err error) error {
	if err := c.JSON(http.StatusUnauthorized, rout.ErrorResponse(err)); err != nil {
		log.Error().Err(jsonErr).Msg("failed to write unauthorized JSON response")

		return err
	}

	return nil
}

func unauthorized(c echo.Context, err error, conf *Options, v tokens.Validator) error {
	reqCtx := c.Request().Context()

	if v != nil && conf != nil && conf.DBClient != nil {
		orgID := orgIDFromToken(c, v)
		if orgID == "" {
			return returnUnauthorized(c, err)
		}

		userID := userIDFromToken(c, v)
		if userID == "" {
			return returnUnauthorized(c, err)
		}

		enforced, err := isSSOEnforcedFunc(reqCtx, conf.DBClient, orgID)
		if err != nil {
			return returnUnauthorized(c, err)
		}

		if enforced {
			role, mErr := orgRoleFunc(reqCtx, conf.DBClient, userID, orgID)
			if mErr == nil && role != enums.RoleOwner {
				if redirErr := c.Redirect(http.StatusFound, sso.SSOLogin(c.Echo(), orgID)); redirErr != nil {
					return redirErr
				}

				return nil
			}
		}

	}

	return returnUnauthorized(c, err)
}
```

pkg/middleware/auth/auth.go
```diff
@@ -341,11 +362,156 @@ func getSubjectName(user *ent.User) string {
...
+	}
+
+	return ""
+}
+
+// isSSOEnforced checks if SSO is enforced for the given organization ID
+func isSSOEnforced(ctx context.Context, db *ent.Client, orgID string) (bool, error) {
+	setting, err := db.OrganizationSetting.Query().Where(organizationsetting.OrganizationID(orgID)).
+		Only(privacy.DecisionContext(ctx, privacy.Allow))
+	if err != nil {
```
golanglemonade, line 433: You should just pull the field you need
```suggestion
	setting, err := db.OrganizationSetting.Query().Where(organizationsetting.OrganizationID(orgID)).
	        Select(organizationsetting.IdentityProviderLoginEnforced).
		Only(privacy.DecisionContext(ctx, privacy.Allow))
	if err != nil {
```

golanglemonade (approved):
nice! this is great

PR #951, opened 2025-07-06
trust center advanced color/font settings

acookin:
Add some fields to support the "advanced" section in the figma:
image

I am intentionally not doing validations on the font because I feel like it will be easier to just let the frontend handle this, at least from the offset, so we don't have to keep making backend changes to this.
Also allowing "advanced" settings to be set even in "easy" mode, because I think the frontend can handle this, and this makes it easier for the settings to persist even when you're toggling between the two modes. I can easily fix this in the mutation hook if that is needed, but I feel like this will just create less friction for the frontend work. (or maybe i'm being lazy, but I guess why not both?)


internal/ent/schema/trustcenter_settings.go
```diff
@@ -83,6 +81,29 @@ func (TrustCenterSetting) Fields() []ent.Field {
...
 			).
 			Nillable(),
+		// Color/font settings
+		field.Enum("theme_mode").
+			Comment("Theme mode for the trust center").
+			GoType(enums.TrustCenterThemeMode("")).
+			Default(enums.TrustCenterThemeModeEasy.String()).
+			Optional(),
+		// Easy options
+		field.String("primary_color").
```
golanglemonade, line 91: I assume these should be hex codes? should we do validation on these?

golanglemonade (dismissed):
Just one question on the color fields; needs a rebase because I merged another PR - I can handle this if you don't get to it before me though.

matoszz (approved):

acookin:
> Just one question on the color fields; needs a rebase because I merged another PR - I can handle this if you don't get to it before me though.

Good point, I'll follow up and add the hex validation to all the color fields

PR #954, opened 2025-07-06
subprocessor schema initialization

acookin:
Initialize the schema for trust center processors.

The idea is that there are "subprocessors" that can by system or org owned (these would be actual companies n stuff, like AWS, GCP, etc etc), then there is the "trust center subprocessor" which is basically a mapping table from trust center to subproccessor.


golanglemonade (dismissed):
let's rebase this one after the other one you have out and we can 🚀  it

PR #968, opened 2025-07-09
Subprocessor schema + api implementation

acookin:
Schema implementation for subprocessors.

Supports system-owned and org-owned subprocessors.

Basically just title (e.g. AWS), description (e.g. cloud computing platform), logo.


internal/ent/schema/subprocessor.go
```diff
@@ -36,47 +46,98 @@ func (Subprocessor) PluralName() string {
...
+			policy.CheckOrgWriteAccess(),
 		),
 	)
 }
 
 // Indexes of the Subprocessor
 func (Subprocessor) Indexes() []ent.Index {
-	return []ent.Index{}
+	return []ent.Index{
+		index.Fields("name").
```
golanglemonade: I think this needs `owner_id` as well? If not, an organization could create "Cool Company" subprocessor, and then no one else using the platform could add that subprocesso.

golanglemonade (dismissed):
looks good

PR #973, opened 2025-07-09
Feat org check

golanglemonade:


pkg/catalog/README.md
```diff
@@ -97,7 +97,7 @@ independently from session data and can be fetched with `Get` or written with
 `Set`.
 
 ```go
-c := features.NewCache(redisClient, time.Minute)
+c := features.NewCache(redisClient, permissioncache.WithCacheTTL(time.Minute))
```
matoszz: you should remove the references that are in here to the cache since the package is moved

pkg/permissioncache/cache.go
```diff
@@ -0,0 +1,153 @@
...
+
+// SetRole stores the role for a subject ID.
+func (c *Cache) SetRole(ctx context.Context, subjectID, orgID string, value string) error {
+	roles, err := c.get(ctx, c.roleKey(subjectID, orgID))
+	if err == nil {
+		if slices.Contains(roles, value) {
+			return nil
+		}
+
+		roles = append(roles, value)
```
matoszz:
```suggestion
if err != nil {
  return err
}

if slices.Contains(roles, value) {
  return nil
}
 
roles = append(roles, value)

```

internal/ent/interceptors/organization.go
```diff
@@ -23,25 +25,67 @@ func InterceptorOrganization() ent.Interceptor {
...
 
-		// if this is an API token, only allow the query if it is for the organization
-		if auth.IsAPITokenAuthentication(ctx) {
-			orgID, err := auth.GetOrganizationIDFromContext(ctx)
-			if err != nil {
-				return err
+		// Authenticated users should have their organizations IDs set in their auth context
+		// after logging in, check this first before using the AddIDPredicate and requiring a
+		// query to fga
+		authorizedOrgs, err := auth.GetOrganizationIDsFromContext(ctx)
```
matoszz, line 40: no check on the `err`?

internal/ent/interceptors/organization.go
```diff
@@ -23,25 +25,67 @@ func InterceptorOrganization() ent.Interceptor {
...
+			if authType != auth.JWTAuthentication {
+				q.WhereP(organization.IDIn(authorizedOrgs...))
+
+				return nil
+			}
+
+			// if the request is using a JWT and is not org owned, for example user profile, personal access tokens, etc,
+			// as well as a query on all organizations for a user,
+			// we need to restrict on all organization instead of just the current one
+			useListObjectsTypes := []string{
```
matoszz, line 57: instead of strings, maybe use a map to get better performance? 

something like...

```
			useListObjectsTypes := map[string]struct{}{
				"personalAccessToken": {},
				"organization":        {},
				"orgMemberships":      {},
				"userSetting":         {},
				"user":                {},
			}
```

internal/ent/interceptors/organization.go
```diff
@@ -23,25 +25,67 @@ func InterceptorOrganization() ent.Interceptor {
...
+			fieldCheck := ""
+			if fCtx != nil {
+				if fCtx.Object == "Query" || fCtx.Object == "Mutation" {
+					fieldCheck = fCtx.Field.Name
+				} else {
+					fieldCheck = fCtx.Object
+				}
+
+				if fieldCheck != "" {
+					for _, t := range useListObjectsTypes {
```
matoszz, line 76: are the `ToLower` manipulations going to cause a problem with camelcase objects? e.g. `userSetting` vs. `usersetting`

internal/ent/interceptors/organization.go
```diff
@@ -23,25 +25,67 @@ func InterceptorOrganization() ent.Interceptor {
...
+		// after logging in, check this first before using the AddIDPredicate and requiring a
+		// query to fga
+		authorizedOrgs, err := auth.GetOrganizationIDsFromContext(ctx)
+
+		authType := auth.GetAuthTypeFromContext(ctx)
+		if err == nil && len(authorizedOrgs) > 0 {
+			// if the request is not using a JWT, we can restrict to the authorized orgs
+			// from the context
+			if authType != auth.JWTAuthentication {
+				q.WhereP(organization.IDIn(authorizedOrgs...))
```
matoszz, line 47: do you actually need this check here since it's also done on line 83?

internal/ent/interceptors/organization.go
golanglemonade: there is, two lines down for `== nil`. If we dont get it back, I want it to continue because there is other logic that could allow it to pass in `AddIDPredicate`

internal/ent/interceptors/organization.go
golanglemonade: no, thats why I have them - so it doesn't matter.

internal/ent/interceptors/organization.go
golanglemonade: yes, because here it will prevent any use of AddPredicateID for non-JWT auth, line 83 is for all JWT auth that is an org owned object.

internal/ent/interceptors/organization.go
golanglemonade: hm it needs to include not just the plural, but the operations, e.g. `deleteOrganization`, etc. So I'm pattern matching rather than listing every specific type.

pkg/permissioncache/cache.go
golanglemonade: Good call, neither are quite right though since `roles` would be nil if not found. Pushing a fix.

internal/ent/interceptors/organization.go
golanglemonade: I added a comment to make that clearer what it was doing, because even I forgot why I had it that way :)

matoszz (approved):

PR #980, opened 2025-07-10
Trust center subprocessor schema/api implementation

acookin:
Schema for associating a subprocessor with a trust center. Also allows for a bit of metadata to attach to the subprocessor, like what countries/regions the subprocessor is used/operating in, and the category that the company is using the subprocessor for.


internal/ent/interceptors/trustcenter.go
```diff
@@ -19,3 +22,52 @@ func InterceptorTrustCenter() ent.Interceptor {
...
+				sql.In(
+					s.C("trust_center_id"),
+					sql.Select(t.C(trustcenter.FieldID)).From(t).Where(
+						sql.In(
+							t.C(trustcenter.FieldOwnerID), anys...,
+						),
+					),
+				),
+			)
+		})
```
golanglemonade, line 69: any reason you are using the generic `sql` functions instead of the typed variables from the generate files like you did in the subprocessor intercetpor?

internal/ent/schema/trustcenter_subprocessor.go
```diff
@@ -58,25 +97,34 @@ func (TrustCenterSubprocessor) Hooks() []ent.Hook {
 func (TrustCenterSubprocessor) Policy() ent.Policy {
 	return policy.NewPolicy(
 		policy.WithQueryRules(
-			privacy.AlwaysDenyRule(),
+			privacy.AlwaysAllowRule(),
 		),
 		policy.WithMutationRules(
-			privacy.AlwaysDenyRule(),
+			entfga.CheckEditAccess[*generated.TrustCenterSubprocessorMutation](),
+			privacy.AlwaysAllowRule(),
```
golanglemonade:
```suggestion
```

Why always allow?

internal/ent/interceptors/trustcenter.go
acookin: because i wanted to make it generic so i don't have to copy paste this each time (updated trust center setting to use this as well). I should probably allow this to do the system admin thing then update the subprocessor to use this interceptor instead.

internal/ent/schema/trustcenter_subprocessor.go
acookin: err whoops haha. left over from setting it to always allow so i could do the initial generate

internal/ent/interceptors/trustcenter.go
acookin: oh, nvm bc the subprocessor isn't trust center owned. (and we're going to replace this with entity anyway?)

golanglemonade (approved):

PR #982, opened 2025-07-11
Dynamic exports for all schemas 

adelowo:
- Dynamic exports now supported. You just need to add `entx.Exportable{}` to any schema's annotation.
- Dynamic export values now supported. it will default to only `id`

```
mutation($input: CreateExportInput!) {
 createExport(input: $input) {
  export {
    id
    fields
  }
 }
}

{
  "input": {
    "exportType": "CONTROL",
    "format": "CSV",
    "fields": ["id", "createdAt", "displayID"]
  }
}

```

- Added a new enum `NODATA` to indicate to client that there is no need to download any file since no data was exported ( because it does not exists )
- Fixed a bug where `ownerID` was not added to the filters for the exported data so you could export data for another org :)) 
- Evidences can now be exported 
- Updated scheduled worker to delete `NODATA` exports too alongside `READY` exports
- Updated queries and client to include pagination which would be needed to export all data and not get limited to 100 items


internal/ent/hooks/export.go
```diff
@@ -8,14 +8,13 @@ import (
...
 	"github.com/theopenlane/core/pkg/corejobs"
-	"github.com/theopenlane/core/pkg/enums"
 	"github.com/theopenlane/core/pkg/objects"
 	"github.com/theopenlane/iam/auth"
 )
 
 var (
-	errExportTypeInvalid     = errors.New("export type is invalid")
 	errExportTypeNotProvided = errors.New("provide export type")
+	errFieldsNotProvided     = errors.New("atleast one field must be provided for the schema to export")
```
golanglemonade:
```suggestion
	errFieldsNotProvided     = errors.New("at least one field must be provided for the schema to export")
```

internal/ent/schema/export.go
```diff
@@ -61,6 +61,11 @@ func (Export) Fields() []ent.Field {
 			Annotations(
 				entgql.Skip(entgql.SkipMutationCreateInput),
 			),
+		field.JSON("fields", []string{}).
```
golanglemonade, line 64: we need filters in addition to fields  (we can do this in a follow-up PR)

pkg/corejobs/export.go
```diff
@@ -73,39 +79,46 @@ func (w *ExportContentWorker) Work(ctx context.Context, job *river.Job[ExportCon
...
 
-func (w *ExportContentWorker) exportControls(ctx context.Context, exportID string) error {
-	controls, err := w.olClient.GetAllControls(ctx)
+	exportType := strings.ToLower(export.Export.ExportType.String())
+
+	fields := export.Export.Fields
+
+	// since the exporttype would be correct. just append "s" to build up the root query
+	// e.g control becomes controls , evidence becomes evidences
+	rootQuery := exportType + "s"
```
golanglemonade: Look at `github.com/gertd/go-pluralize` - we use this in `ent/schema` - not all schemas will just be "s", e.g `internalPolicy` -> `internalPolicies`

internal/ent/hooks/exportable_generated.go
```diff
@@ -9,6 +9,8 @@ import (
 
 // ExportableSchemas contains all schemas that have Exportable annotation
 var ExportableSchemas = map[string]bool{
+	"control": true,
+	"evidence": true,
```
golanglemonade, line 13: can you add the annotation to these schemas as well: 
- internal policies
- risks
- procedures
- subscribers 
- tasks

pkg/corejobs/export.go
```diff
@@ -119,15 +132,161 @@ func (w *ExportContentWorker) exportControls(ctx context.Context, exportID strin
...
+
+func (w *ExportContentWorker) executeGraphQLQuery(ctx context.Context, query string) (map[string]any, error) {
+	body := map[string]string{"query": query}
+
+	jsonBody, err := json.Marshal(body)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", w.Config.OpenlaneAPIHost+"/query", bytes.NewReader(jsonBody))
```
golanglemonade: graph queries are limited to the first 100 results, you'll need to paginate the results to get them all

pkg/enums/exporttype.go
```diff
@@ -12,6 +12,8 @@ type ExportType string
 var (
 	// ExportTypeControl indicates the control.
 	ExportTypeControl ExportType = "CONTROL"
+	// ExportTypeEvidence indicates the evidence.
+	ExportTypeEvidence ExportType = "EVIDENCE"
```
golanglemonade, line 16: Any reason we can't just use the `ExportableSchemas` to validate instead of making an enum that we have to manually update?

golanglemonade (commented):
nice work!

pkg/enums/exporttype.go
adelowo: this is really only here because it provides _typing_ to clients to know the export types they can use on the frontend but it is also not manually updated else we will just make it a string and that won't be the best for clients. 

`task generate` will always update this enum to match the schemas that have `entx.Exportable` so it is not manually maintained

We use `ExportableSchemas` in the hooks to validate and that is also updated automatically by `task generate`

golanglemonade: ah sweet; thanks for clarifying -  missed that it wasn't manually updated :)

PR #983, opened 2025-07-11
Initial Module Catalog Implementation

matoszz:
This PR encompasses the initial consumption of the catalog framework created in [PR 909](https://github.com/theopenlane/core/pull/909). The majority of the additions in the CLI within `cmd/catalog/main.go` are related to functional testing results / failures experienced when repeatedly using the utility in a Stripe sandbox. Overall goal(s) are to perform as many pre-checks as possible to ensure the validity of the catalog as it relates to the configured stripe account, and write back results of the catalog to the file itself (and generated types). Logic was added to create (and link) the added schemas for OrgModule, OrgProduct, and OrgPrice. Aside from the larger changes, this PR additionally includes:

- Generated types for each of the respective `OrgModules` in `pkg/models`
- Flags for specifying which modules should be used in certain situations, ex: `include_with_trial` and `personal_org`
- Helper functions for accessing modules with the above specified flags, ex: "I want all modules flagged for trials"
- CLI now performs Product and Feature "takeover" so that any combination of existing Products, Prices, or Features can be reconciled
- Reconciliation logic which will compare org subscriptions within the Openlane system that do not exist in Stripe
- Creation of default Orgmodule, OrgProduct, and OrgPrice within the organization ent hook
- Refactor / collapsing of several of the `pkg/subscription` functions related to creation of subscriptions so that we are consistently using the same function to create stripe subscriptions but passing various option(s) to control if it should be a trial or not, if multiple prices should be included or not, etc.

High level remaining TODO:
- Ensure feature tuples are created properly for each module the customer is entitled to
- Add interceptors and hooks already created for features to ent schemas to be able to confirm the action being taken against a schema / object is allowed via the entitlements
- Update the received webhooks to properly reconcile OrgModule, OrgProduct, and OrgPrice (today is primarily subscriptions)
- Update GraphQL scalars / URL propagation to populate Organization Settings billing page with module purchase buttons (and cancellation)
- Test / verify behaviors within the UI when modules are enabled (e.g. access is granted on return to the page) and also when they are cancelled / expired (e.g. a user clicks cancellation button, cancels a module, they should then be denied access to those features and presented with the option to re-purchase)
- Fully remove the "tier" references from the OrgSubscription module
- Update the feature maps / entitlements queries to return the products / prices attached to a subscription and not the feature lists
- Add functions, overall capability, and tests, which enable / allow a customer to move from a monthly subscription to a yearly (you cannot have mixed billing in stripe where the subscription contains some monthly prices and some yearly, so you have to convert all prices on the subscription to yearly) 
- Test / verify the behavior of logging in and accessing the billing page when all modules are expired (customer should be able to re-activate subscription if all expired)
- Add UI capabilities which only display components the user is entitled to via modules, and if a section within the UI is visible but the customer cannot access it, they should be presented with the option to purchase it
- Convert the entx.Features schema annotations to use the generated types added in this PR
- Test / verify the behavior related to staggered module additions / subscription updates; the desired behavior if a user were to add a module mid-billing cycle is that they are charged a pro-rated amount, and all billing is co-term
- End-to-end test behaviors, code, logic, user experience related to price migrations. If we want to change an underlying price associated with a module, we need to define a new price, and choose (or not) to migrate existing customers to this new price as Stripe prices are immutable. This should be a seamless and easy experience for us and the user.
- Add metering functionality that would allow us to meter an org module by various attributes, e.g. "Maximum number of programs an organization can have is 10" (stripe has Metering functionality that is not yet taken advantage of)


internal/ent/hooks/organization.go
```diff
@@ -599,3 +617,117 @@ func createFeatureTuples(ctx context.Context, authz fgax.Client, orgID string, f
...
+
+		if cfg.trial && !mod.IncludeWithTrial {
+			continue
+		}
+
+		if !cfg.personalOrg && !cfg.trial {
+			continue
+		}
+
+		// Find the first price with "month" interval
```
golanglemonade: nit: if you check the cfg first, you can make the others simpler: 

```suggestion
		if !cfg.personalOrg && !cfg.trial {
			continue
		}
		
		if !mod.PersonalOrg {
			continue
		}

		if !mod.IncludeWithTrial {
			continue
		}

		// Find the first price with "month" interval
```

internal/ent/hooks/organization.go
```diff
@@ -599,3 +617,117 @@ func createFeatureTuples(ctx context.Context, authz fgax.Client, orgID string, f
...
+			SetProductID(orgProduct.ID).
+			SetPrice(models.Price{Amount: float64(monthlyPrice.UnitAmount), Interval: monthlyPrice.Interval}).
+			SetOwnerID(orgCreated.ID).
+			SetSubscriptionID(orgSubs.ID).
+			SetStripePriceID(monthlyPrice.PriceID).
+			SetActive(true).
+			Save(newCtx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create OrgPrice for module %s: %w", moduleName, err)
+		}
```
golanglemonade: Use `Exec` if you don't need the response, this saves an additional query to get the updated result back. 

```suggestion
		err = m.Client().OrgPrice.Create().
			SetProductID(orgProduct.ID).
			SetPrice(models.Price{Amount: float64(monthlyPrice.UnitAmount), Interval: monthlyPrice.Interval}).
			SetOwnerID(orgCreated.ID).
			SetSubscriptionID(orgSubs.ID).
			SetStripePriceID(monthlyPrice.PriceID).
			SetActive(true).
			Exec(newCtx)
		if err != nil {
			return nil, fmt.Errorf("failed to create OrgPrice for module %s: %w", moduleName, err)
		}
```

internal/entitlements/reconciler/reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil, ErrMissingClients
+	}
+
+	if r.dryRun && r.writer == nil {
+		r.writer = os.Stdout
+	}
+
+	return r, nil
+}
+
```
golanglemonade, line 82: You want a `DeletedAtIsNil()` added to a where clause here

internal/entitlements/reconciler/reconciler.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	customerMissing := sub == nil || sub.StripeCustomerID == ""
+	subscriptionMissing := sub == nil || sub.StripeSubscriptionID == ""
+
+	if sub != nil && sub.StripeCustomerID != "" {
+		if _, err := r.stripe.GetCustomerByStripeID(ctx, sub.StripeCustomerID); err != nil {
+			customerMissing = true
+		}
+	}
```
golanglemonade: I think you can just use your bools defined above: 
```suggestion
	if !customerMissing {
		if _, err := r.stripe.GetCustomerByStripeID(ctx, sub.StripeCustomerID); err != nil {
			customerMissing = true
		}
	}

	if !subscriptionMissing {
		if _, err := r.stripe.GetSubscriptionByID(ctx, sub.StripeSubscriptionID); err != nil {
			subscriptionMissing = true
		}
	}
```

cmd/catalog/main.go
```diff
@@ -65,7 +69,7 @@ func catalogApp() *cli.Command {
 			&cli.StringFlag{
 				Name:  "catalog",
 				Usage: "catalog file path",
-				Value: "./pkg/catalog/catalog.yaml",
+				Value: "./pkg/catalog/catalog.yaml", // set the value to sandbox catalog by default to avoid disasters
```
golanglemonade: your comment and value are mismatched. Should this be? 
```suggestion
				Value: "./pkg/catalog/catalog_sandbox.yaml", // set the value to sandbox catalog by default to avoid disasters
```

pkg/catalog/precheck_test.go
```diff
@@ -0,0 +1,80 @@
...
+	feat := &stripe.EntitlementsFeature{ID: "feat_x", LookupKey: "dup_feat"}
+	prod := &stripe.Product{ID: "m1", Name: "mod1"}
+
+	conflicts, err := cat.LookupKeyConflicts(context.Background(), &fakeLookup{price: price, feature: feat, product: prod, products: []*stripe.Product{prod}})
+	require.NoError(t, err)
+	require.Len(t, conflicts, 3)
+}
+
+func TestLookupKeyConflictsByName(t *testing.T) {
+	cat := &catalog.Catalog{
```
golanglemonade, line 68: not a blocker, but just fyi I think most of your tests can be run with `t.Parallel()` which will help speed up the test suite

pkg/entitlements/subscriptions.go
```diff
@@ -106,82 +63,54 @@ func (sc *StripeClient) GetSubscriptionByID(ctx context.Context, id string) (*st
...
-
-	return subscription, nil
-}
-
 var trialdays int64 = 30
 
-// CreateTrialSubscription creates a trial subscription with the configured price
-func (sc *StripeClient) CreateTrialSubscription(ctx context.Context, cust *stripe.Customer) (*Subscription, error) {
+// CreateTrialSubscriptionWithPrices creates a trial subscription using the provided prices
+func (sc *StripeClient) CreateSubscriptionWithPrices(ctx context.Context, cust *stripe.Customer, o *OrganizationCustomer) (*Subscription, error) {
```
golanglemonade:
```suggestion
// CreateSubscriptionWithPrices creates a subscription using the provided prices
func (sc *StripeClient) CreateSubscriptionWithPrices(ctx context.Context, cust *stripe.Customer, o *OrganizationCustomer) (*Subscription, error) {
```

PR #988, opened 2025-07-13
actually make the default trust center domain configurable

acookin:


internal/httpserve/handlers/tools_test.go
```diff
@@ -220,6 +220,7 @@ func handlerSetup(db *ent.Client) *handlers.Handler {
 		OauthProvider: handlers.OauthProviderConfig{
 			RedirectURL: "http://localhost",
 		},
+		DefaultTrustCenterDomain: "trust.openlane.com",
```
golanglemonade, line 223: this should be in the config instead of just hardcoded here, approving for now though

golanglemonade (approved):

PR #993, opened 2025-07-14
Windmill integration

adelowo:
- Create jobs in windmill. These are regular code/scripts that will be wrapped into Windmill compatible layer
- Created a tiny windmill client/sdk
- create scheduled job in windmill when a control is added to a job
- add support for both on success and failure scripts which would be appended to scheduled jobs - globally applied

```yml

entConfig:

  windmill:
    enabled: true
    baseURL: "http://localhost"
    workspace: "test"
    token: "xyz"
    defaultTimeout: "30s"
    onFailureScript: "u/admin/glimmering_script"
    onSuccessScript: "u/admin/glimmering_script"

```

```

mutation($input: CreateScheduledJobInput!) {
  createScheduledJob(input: $input) {
    scheduledJob {
      id
    }
  }
}

{
  "input": {
    "cron": "0 22 * * 1-5",
    "downloadURL": "https://gist.githubusercontent.com/",
    "title": "SSL scanning test",
  },
}

```


internal/ent/hooks/scheduled_job.go
```diff
@@ -22,69 +33,190 @@ func HookScheduledJobCreate() ent.Hook {
...
-
-			if err := validateCadenceOrCron(&cadence, hasCadence, cron, hasCron); err != nil {
-				return nil, err
+			if mutation.Op() == ent.OpCreate {
+				if err := createWindmillFlow(ctx, mutation); err != nil {
+					return nil, fmt.Errorf("failed to create windmill flow: %w", err)
+				}
 			}
 
 			return next.Mutate(ctx, mutation)
```
golanglemonade: I think this should happen after the next.Mutate, so we can make sure the validation happens correctly and then the flow is created in windmill
```suggestion
                        retVal, err := next.Mutate(ctx, mutation)
                        if err != nil {
                            return err
                         }
                          
			if mutation.Op() == ent.OpCreate {
				if err := createWindmillFlow(ctx, mutation); err != nil {
					return nil, fmt.Errorf("failed to create windmill flow: %w", err)
				}
			}

			return retVal, err
```

internal/ent/hooks/scheduled_job.go
```diff
@@ -22,69 +33,190 @@ func HookScheduledJobCreate() ent.Hook {
...
 
+// createWindmillFlow creates a Windmill flow for the scheduled job
+func createWindmillFlow(ctx context.Context, mutation *generated.ScheduledJobMutation) error {
+	windmillClient := mutation.Client().Windmill
+	if windmillClient == nil {
+		return nil
+	}
+
+	// Get job details
+	title, hasTitle := mutation.Title()
```
golanglemonade: this is a required field right? as long as this happens after the db create, you shouldn't need to also validate it here.

internal/ent/hooks/scheduled_job.go
```diff
@@ -22,69 +33,190 @@ func HookScheduledJobCreate() ent.Hook {
...
+	if downloadURL == "" {
+		return errors.New("download_url is required for windmill flow creation") // nolint:err113
+	}
+
+	rawCode, err := downloadRawCode(ctx, downloadURL)
+	if err != nil {
+		return err
+	}
+
+	flowPath := fmt.Sprintf("f/openlane/scheduled_job_%s", generateFlowPath())
```
golanglemonade: this should be configurable, and not hardcoded

internal/ent/hooks/scheduled_job.go
golanglemonade: How does this work if a user updates the job?

internal/ent/hooks/scheduled_job.go
```diff
@@ -22,69 +33,190 @@ func HookScheduledJobCreate() ent.Hook {
...
 			if entx.CheckIsSoftDelete(ctx) {
 				return next.Mutate(ctx, mutation)
 			}
 
-			if err := validateCadenceOrCron(&cadence, hasCadence, cron, hasCron); err != nil {
+			if err := validateCron(cron, hasCron); err != nil {
 				return nil, err
 			}
 
+			if mutation.Op() == ent.OpCreate && hasCron && cron != "" {
```
golanglemonade, line 211: same here, how do I update the cron schedule if this is wrapped in OpCreate only?

internal/ent/hooks/scheduled_job.go
```diff
@@ -22,69 +33,190 @@ func HookScheduledJobCreate() ent.Hook {
...
 	}
 
-	if hasCron {
-		return cron.Validate()
+	entConfig := mutation.Client().EntConfig
+	if entConfig == nil {
+		return errors.New("ent config is required") // nolint:err113
 	}
 
+	scheduledJobPath := fmt.Sprintf("s/openlane/control_scheduled_job_%s", generateFlowPath())
```
golanglemonade: should be configurable

internal/ent/hooks/scheduled_job.go
```diff
@@ -22,69 +33,190 @@ func HookScheduledJobCreate() ent.Hook {
...
+		Enabled:  &enabled,
+		Summary:  fmt.Sprintf("Control scheduled job %s", jobID),
+	}
+
+	resp, err := windmillClient.CreateScheduledJob(ctx, scheduleReq)
+	if err != nil {
+		return err
+	}
+
+	_ = resp.Path
```
golanglemonade: what is this for? 

```suggestion
```

internal/graphapi/models_test.go
```diff
@@ -1521,22 +1521,22 @@ type ScheduledJobBuilder struct {
 }
 
 func (w *ScheduledJobBuilder) MustNew(ctx context.Context, t *testing.T) *ent.ScheduledJob {
+	const testScriptURL = "https://gist.githubusercontent.com/adelowo/5559b84fabf87f02502166a6467e345e/raw/b317494adde0f9ba16e09dd019681d1dac8d4d05/test.go"
```
golanglemonade: probably be better if this is in this org itself, instead of depending on your user's gist - might be good to just go ahead and create a `jobs` repo and stash it in there?

internal/ent/hooks/scheduled_job.go
adelowo: removed. was debugging earlier

internal/graphapi/models_test.go
adelowo: makes sense. we can use it as a community sourced version or examples for users really.

internal/ent/hooks/scheduled_job.go
adelowo: we do not need to ever update the scheduled job in windmill since it is just a pointer to a flow/original job - which we can already update and create.

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/100

### Changes Preview:
<br/>- 🔄 Merged Helm values.yaml<br/>- 🆕 Created External Secrets templates<br/>- ✨ Created ConfigMap template

internal/ent/hooks/scheduled_job.go
adelowo: for this, I do not want to make fields like the windmill path optional - as they aren't. the job creation also does all possible validation _asides the context_.

let me know if to move still

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/100

### Changes Preview:
🔧 Applying Helm configuration changes...<br/> <br/>- 🔄 Merged Helm values.yaml🔄 Merging Helm values.yaml<br/>   🔀 Merging with existing chart values...<br/>   📋 Replacing core section...<br/>   🔐 Merging external secrets configuration...<br/>   📊 Analyzing changes...<br/> <br/>    • External secrets configuration updated<br/> 

internal/ent/hooks/scheduled_job.go
golanglemonade: gotcha, that makes sense - but we would still need to be able to roll this back if the transaction fails so we don't end up with a hanging flow if something causes the db transaction to fail

golanglemonade (approved):

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/111

### Changes Preview:
🔧 Applying Helm configuration changes...<br/> Updating ConfigMap template<br/> ✅ Updated ConfigMap template<br/> <br/>- ✅ Updated ConfigMap template<br/> 

PR #1011, opened 2025-07-16
user defined filters for exports

adelowo:
- added filters for exports
- dropped `job_type` for jobs
- added support for logging failure reasons so UI can notify user

```
{
  "input": {
    "exportType": "TASK",
    "format": "CSV",
    "filters": "{\"statusIn\":[\"OPEN\",\"IN_PROGRESS\",\"IN_REVIEW\"]}"
  }
}
```


internal/ent/schema/export.go
```diff
@@ -66,6 +66,15 @@ func (Export) Fields() []ent.Field {
...
 			Immutable(),
+
+		field.String("filters").
+			Comment("the specific filters to run against the exported data. This should be a well formatted graphql query").
+			Optional().
+			Immutable(),
+
+		field.String("error_message").
+			Optional().
+			Comment("if we try to export and it fails, the error message will be stored here"),
```
golanglemonade: this should have a skip on create annotation

pkg/corejobs/export.go
```diff
@@ -117,24 +121,47 @@ func (w *ExportContentWorker) Work(ctx context.Context, job *river.Job[ExportCon
...
+		if filters != "" {
+			if err := json.Unmarshal([]byte(filters), &filterMap); err != nil {
+				log.Error().Err(err).Msg("failed to parse filters")
+				return w.updateExportStatus(ctx, job.Args.ExportID, enums.ExportStatusFailed, err)
+			}
+
+			if _, exists := filterMap["ownerID"]; exists {
+				log.Error().Err(errInvalidFilters).Msg("invalid filters")
+				return w.updateExportStatus(ctx, job.Args.ExportID, enums.ExportStatusFailed, errInvalidFilters)
+			}
```
golanglemonade: why can't ownerID be in the filters? Not sure this should error, instead just dropped?

adelowo: Because the job adds it automatically before making the request. We don't want anyone to try to set that field. 

FGA checks would fail for it regardless but it's just a defensive check

pkg/corejobs/export.go
```diff
@@ -31,13 +33,15 @@ const (
...
-	errMissingEndCursor   = fmt.Errorf("missing endCursor in pageInfo")
+	errUnexpectedStatus    = errors.New("unexpected HTTP status")
+	errInvalidFilters      = errors.New("filters cannot define ownerID")
+	errGraphQLMessage      = errors.New("GraphQL error")
+	errUnknownGraphQLError = errors.New("an unknown error occurred")
+	errMissingRoot         = errors.New("missing root in response")
+	errMissingEdges        = errors.New("missing edges in response")
+	errMissingPageInfo     = errors.New("missing pageInfo in response")
+	errMissingHasNextPage  = errors.New("missing hasNextPage in pageInfo")
+	errMissingEndCursor    = errors.New("missing endCursor in pageInfo")
```
golanglemonade: nit: the reason for defining errors is so that when other packages use this, they can do `errors.Is`, which isn't possible if they aren't exported. I would consider exporting these

golanglemonade (commented):
just a few comments

pkg/corejobs/export.go
golanglemonade: it would only fail if the owner id wasn't valid - not if its set at all. When using a personal access token; you might have a token valid for multiple orgs, and you would set the owner ID

adelowo: yes but the ownerID is also already known from the time you created the export. We store the requestor_id and owner_id in the database and just reuse that. 

at the time of creating the export, you have set the ownerID for that specific export. 

but let me know if to remove still

golanglemonade: I think we should just ignore the filter instead of erroring, its a valid filter on the schema so it would feel weird to have it fail on something that is just ignored? And once we add the token exchange stuff; we dont need to actually set the owner ID in the filter ourselves

golanglemonade (approved):

PR #1014, opened 2025-07-16
Wrap build pipeline steps in conditional checks based on files changed

matoszz:
The overall intent here is to only run steps relevant to the scope of the changes being implemented, but to fall back to running the entire pipeline if necessary.


.buildkite/pipeline.yaml
```diff
@@ -42,6 +99,7 @@ steps:
                 - "GOTOOLCHAIN=auto"
       - label: ":yaml: generate config"
         key: "generate_config"
+        if: build.env("CONFIG_CHANGES") == "true" || build.env("BUILDKITE_CHANGES") == "true" || build.env("RUN_ALL") == "true"
```
golanglemonade: I don't think these are build envs since they are set from a step prior;  you'd need to escape them I believe but not sure if that actually works in an if statement here

PR #1019, opened 2025-07-17
Cascade delete controls

adelowo:
If an admin deletes a standard, it should cascade delete all controls under it.

org owned controls linked to the standard do not get deleted, they just get unlinked from the standard

### DIFF
--- a/internal/ent/hooks/standard.go
+++ b/internal/ent/hooks/standard.go
@@ -20,7 +20,12 @@ import (
 func HookStandardDelete() ent.Hook {
 	return hook.On(func(next ent.Mutator) ent.Mutator {
 		return hook.StandardFunc(func(ctx context.Context, m *generated.StandardMutation) (generated.Value, error) {
-			if !entx.CheckIsSoftDelete(ctx) && !auth.IsSystemAdminFromContext(ctx) {
+
+			if !entx.CheckIsSoftDelete(ctx) {
+				return next.Mutate(ctx, m)
+			}
+
+			if !auth.IsSystemAdminFromContext(ctx) {
 				return next.Mutate(ctx, m)
 			}
 
@@ -28,6 +33,15 @@ func HookStandardDelete() ent.Hook {
 			var err error
 
 			ctx = privacy.DecisionContext(ctx, privacy.Allow)
+
+			retrievedStandard, err := m.Client().Standard.Get(ctx, id)
+			if err != nil {
+				return nil, err
+			}
+
+			if !retrievedStandard.SystemOwned {
+				return next.Mutate(ctx, m)
+			}
 
 			// remove standard_id mapping from org owned controls
 			err = m.Client().Control.Update().ClearStandardID().Where(
--- a/internal/ent/hooks/standard.go
+++ b/internal/ent/hooks/standard.go
@@ -2,18 +2,23 @@ package hooks
 
 import (
 	"context"
+	"errors"
 
 	"entgo.io/ent"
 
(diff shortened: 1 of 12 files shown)


internal/graphapi/control_test.go
```diff
@@ -1010,9 +1010,7 @@ func TestMutationCreateControlsByClone(t *testing.T) {
 	(&Cleanup[*generated.ControlDeleteOne]{client: suite.client.db.Control, IDs: controlIDsToDelete}).MustDelete(testUser1.UserCtx, t)
 	(&Cleanup[*generated.StandardDeleteOne]{client: suite.client.db.Standard, IDs: []string{publicStandard.ID, publicStandard2.ID}}).MustDelete(systemAdminUser.UserCtx, t)
 	(&Cleanup[*generated.StandardDeleteOne]{client: suite.client.db.Standard, ID: orgStandard.ID}).MustDelete(testUser1.UserCtx, t)
-	(&Cleanup[*generated.ControlDeleteOne]{client: suite.client.db.Control, IDs: controlIDs}).MustDelete(systemAdminUser.UserCtx, t)
```
adelowo, line 1013: deleted these since we cascade deletes for system owned standards now

internal/ent/hooks/standard.go
```diff
@@ -5,13 +5,71 @@ import (
...
+			if !auth.IsSystemAdminFromContext(ctx) {
+				return next.Mutate(ctx, m)
+			}
+
+			id, _ := m.ID()
+			var err error
+
+			ctx = privacy.DecisionContext(ctx, privacy.Allow)
+
+			retrievedStandard, err := m.Client().Standard.Get(ctx, id)
```
golanglemonade: the user in context should have access or this should fail anyways? Why do we need the privacy allow? 

We should also just grab the fields we need, not all of them. 
```suggestion
			retrievedStandard, err := m.Client().Standard.Query().Where(standard.ID(id)).Select(standard.FieldSystemOwned,standard.IsPublic).Only(ctx)
```

Another thing, I think we shouldn't allow the delete of public standards, so can you add if retrievedStandard.IsPublic - return error public statndard not allowed to be deleted.

internal/ent/hooks/standard.go
```diff
@@ -5,13 +5,71 @@ import (
...
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.StandardFunc(func(ctx context.Context, m *generated.StandardMutation) (generated.Value, error) {
+
+			if !entx.CheckIsSoftDelete(ctx) {
+				return next.Mutate(ctx, m)
+			}
+
+			if !auth.IsSystemAdminFromContext(ctx) {
+				return next.Mutate(ctx, m)
+			}
```
golanglemonade, line 30: Why do we need this? this should already be checked on the standard privacy policy? And if its a org own standard, I'd expect the same cascade delete of controls as well.

internal/ent/hooks/standard.go
```diff
@@ -5,13 +5,71 @@ import (
...
+				return nil, err
+			}
+
+			// delete all controls not linked to an org (system-owned controls)
+			_, err = m.Client().Control.Delete().Where(
+				control.And(
+					control.StandardID(id),
+					control.OwnerIDIsNil(),
+				),
+			).Exec(ctx)
```
golanglemonade, line 63: this will hard delete; we want to update to DeletedAt  + DeletedBy instead of hard deleting.

matoszz (approved):

PR #1028, opened 2025-07-19
add config

golanglemonade:
allows for easier configuration of riverboat when working in core


matoszz (commented):
won't this get overwritten the next time we run config:generate?

golanglemonade:
> won't this get overwritten the next time we run config:generate?

No, these are riverboat configs, which generate from the repo. Just added it here to make things easier if you want to run it directly out of this repo

matoszz (approved):

PR #1031, opened 2025-07-21
Inherit cron schedule from parent job where possible

adelowo:
When adding a control to a job, if you don't provide a cron schedule, It should inherit that of the parent


adelowo:
haha should have done improvement not bug but either works

golanglemonade (approved):

PR #1050, opened 2025-07-29
Bulk updates

adelowo:
adds bulk updates and tests to a few models where we support bulk creation


golanglemonade (approved):
thanks for including tests here!

PR #1069, opened 2025-07-31
trustcenter compliance implementation

acookin:
Create a mapping table for standards and trust centers so that we can display them in the trust center as "compliance" items.

The trust center compliance hook also manages fga tuples for the standard <> trust center mapping. I'm less confident that I did that entirely "correct", but it seems to be doing what i want it to do.


internal/ent/hooks/trustcenter_compliance.go
```diff
@@ -0,0 +1,83 @@
...
+
+		standardID, standardExists := m.StandardID()
+		if standardExists {
+
+			standardReq := fgax.TupleRequest{
+				SubjectID:   tcID,
+				SubjectType: "trust_center",
+				ObjectID:    standardID,
+				ObjectType:  "standard",
+				Relation:    "associated_with",
```
golanglemonade, line 50: Before I suggest a different way; want to confirm the use case here - this would be to allow trust centers view access to an org own standard?

internal/ent/interceptors/standard.go
```diff
@@ -5,48 +5,89 @@ import (
...
+	req := fgax.ListRequest{
+		SubjectID:   tcID,
+		SubjectType: "trust_center",
+		ObjectType:  "standard",
+		Relation:    "associated_with",
+	}
+
+	zerolog.Ctx(ctx).Debug().Interface("req", req).Msg("getting authorized object ids")
+
+	resp, err := utils.AuthzClientFromContext(ctx).ListObjectsRequest(ctx, req)
```
golanglemonade: `ListObject` requests are pretty expensive and I've been generally trying to avoid them. You should just be able to to add a predicate to the query to only allow ones associated with the trust center to be returned instead?

golanglemonade: Something like this? 

```
		anon, isAnon := auth.AnonymousTrustCenterUserFromContext(ctx)
		if isAnon {
			q.Where(
				standard.HasTrustCenterCompliancesWith(
					trustcentercompliance.HasTrustCenterWith(
						trustcenter.OwnerID(anon.OrganizationID),
					),
				),
			)
			return nil
		}
```

golanglemonade: I went ahead and pushed this, but left the existing function commented out, when I was rebasing this for you: https://github.com/theopenlane/core/pull/1069/commits/c05a93cd427d5d925a6d3bb2f8a1c18ccc448f72#diff-bcfcec9ff9e3477e68e58e2de3d3d11eb064a0eb06c78231fd0c6e725947ed7d 

let me know what you think.

acookin: This makes sense, I think i was just getting too caught up in getting the data from fga. Thanks!

internal/ent/hooks/trustcenter_compliance.go
acookin: yes, and anonymous users access to any standard that associated with a trust center

PR #1079, opened 2025-08-05
Interceptors + policies for entititlements

adelowo:
- Added new interceptors that will always return an empty result if no access to module
- If no access to module, enrich the error graph e.g 

```json
 {
      "message": "feature not enabled for organization",
      "path": [
        "organizations",
        "tasks"
      ],
      "extensions": {
        "code": "MODULE_NO_ACCESS",
        "message": "feature not enabled for organization"
      }
    },

```

Full example  here https://gist.github.com/adelowo/75030912c5a4cc127364706d281262bd

- Updated all tests to include module checks 
- Added interceptor and policy tests to verify module access
- Every schema now attached to atleast one feature. and come with a `Features`  method
- Allow modules to be toggleable 

```
entConfig:
  modules:
    enabled: true

```
- handle subscription creation event for use cases where a sub is added directly from stripe's dashboard
- make sure hook for org modules clean up the tuples if available 
- simplified `stripe_customer_id` logic. Instead of searching everything, I have moved it to `organisations` table as it will always be unique per customer/organization


internal/ent/interceptors/features.go
```diff
@@ -2,31 +2,53 @@ package interceptors
...
+
+			// force an evaluation to false always
+			// so the data to be returned will always be empty or not found
+			q.WhereP(func(s *sql.Selector) {
+				s.Where(sql.ExprP("1=0"))
+			})
+
+			entity := strings.ToLower(q.Type())
+			if !strings.HasSuffix(entity, "s") {
+				entity += "s"
```
golanglemonade: This is to make it plural? You should use go-pluralize; this doesn't work for things like `internalPolicy`; it would give you `internalpolicys` instead of `internalpolicies`

internal/ent/privacy/rule/features.go
```diff
@@ -57,53 +53,62 @@ func orgFeatures(ctx context.Context) ([]string, error) {
...
+				feats = append(feats, ent.Identifier)
+			}
 		}
+	}
+
+	// if EntitlementManager was not usable or features still empty,
+	// try to fallback to the OrgModule
+	if len(feats) == 0 && client != nil {
+		modules, err := client.OrgModule.Query().
+			Select(orgmodule.FieldModule, orgmodule.FieldOwnerID, orgmodule.FieldActive).
```
golanglemonade:
```suggestion
			Select(orgmodule.FieldModule).
```

You shouldn't need either of these fields since you are filtering on them

internal/ent/privacy/rule/features.go
```diff
@@ -57,53 +53,62 @@ func orgFeatures(ctx context.Context) ([]string, error) {
...
+		client = c
+	} else if tx := transaction.FromContext(ctx); tx != nil {
+		client = tx.Client()
+	}
+
 	var feats []string
 
-	ac := utils.AuthzClientFromContext(ctx)
+	// attempt to use the EntitlementManager
+	if client != nil && client.EntitlementManager != nil {
```
golanglemonade: Why do we need this check here? this if doesn't have any calls to the client or entitlement manager?

internal/ent/privacy/rule/features.go
```diff
@@ -57,53 +53,62 @@ func orgFeatures(ctx context.Context) ([]string, error) {
 		}
 	}
 
+	if c := generated.FromContext(ctx); c != nil {
```
golanglemonade: We have similar wrapper functions like `hooks.transactionFromContext` - we should centralize this or reuse one of the existing ones.

internal/ent/privacy/rule/features.go
```diff
@@ -57,53 +53,62 @@ func orgFeatures(ctx context.Context) ([]string, error) {
...
+				ent, parseErr := fgax.ParseEntity(obj)
+				if parseErr != nil {
+					continue
+				}
+				feats = append(feats, ent.Identifier)
+			}
 		}
+	}
+
+	// if EntitlementManager was not usable or features still empty,
```
golanglemonade: What use case are you trying to cover with this? If the entitlement manager isn't is use; we shouldn't be checking this anyways?

internal/ent/privacy/rule/features.go
```diff
@@ -131,20 +136,48 @@ func AllowIfHasFeature(feature string) privacy.QueryMutationRule {
...
+//
+// If false, at least one must be enabled.
+func checkFeatures(ctx context.Context, requireAll bool, feats ...models.OrgModule) (bool, error) {
 	enabled, err := orgFeatures(ctx)
 	if err != nil {
 		return false, err
 	}
 
+	if len(enabled) == 0 {
+		return true, nil
```
golanglemonade: Should this return false if there aren't any?

internal/ent/privacy/rule/features.go
```diff
@@ -131,20 +136,48 @@ func AllowIfHasFeature(feature string) privacy.QueryMutationRule {
...
 	enabledSet := make(map[string]struct{}, len(enabled))
 
 	for _, f := range enabled {
 		enabledSet[f] = struct{}{}
 	}
 
+	if requireAll {
+		// all features must be enabled
+		for _, f := range feats {
+			if _, ok := enabledSet[string(f)]; !ok {
```
golanglemonade: OrgModule has a `String()` function 
```suggestion
			if _, ok := enabledSet[f.String()]; !ok {
```

internal/ent/privacy/rule/features.go
```diff
@@ -167,3 +200,86 @@ func AllowIfHasAnyFeature(features ...string) privacy.QueryMutationRule {
...
+	}
+
+	if tok := token.VerifyTokenFromContext(ctx); tok != nil {
+		return true
+	}
+
+	if tok := token.JobRunnerRegistrationTokenFromContext(ctx); tok != nil {
+		return true
+	}
+
```
golanglemonade: All the `privacyToken` types can use the `SkipTokenInContext` function; e..g something like:

```
	skipTokenType := []token.PrivacyToken{
		&token.OauthTooToken{},
		&token.VerifyToken{},
		&token.SignUpToken{},
		&token.OrgInviteToken{},
		&token.ResetToken{},
		&token.JobRunnerRegistrationToken{},
	}

	if skip := rule.SkipTokenInContext(ctx, skipTokenType); skip {
		return true
	}
	```

File: internal/ent/privacy/rule/features.go [hunk 9]
```diff
@@ -167,3 +200,86 @@ func AllowIfHasAnyFeature(features ...string) privacy.QueryMutationRule {
...
+	// check for bypass
+	// For unauthenticated users, this interceptor
+	// will still run when a query is done to fetch the data such as an api
+	// token or personal access token
+	// And would lead to a situation where the features cannot be
+	// retrieved from the database and a failure occurrs
+	if _, allowCtx := privacy.DecisionFromContext(ctx); allowCtx {
+		return true
+	}
```
golanglemonade: Do we need to skip system admins? 

```
	if auth.IsSystemAdminFromContext(ctx) {
		return true
	}
```

internal/ent/schema/actionplan.go
```diff
@@ -83,18 +85,31 @@ func (a ActionPlan) Mixin() []ent.Mixin {
 		}}.getMixins(a)
 }
 
+func (ActionPlan) Features() []models.OrgModule {
```
golanglemonade: Can you add `Features()` to the example schema template? https://github.com/theopenlane/core/blob/main/internal/ent/base/entinit.tmpl

internal/ent/schema/actionplan.go
```diff
@@ -83,18 +85,31 @@ func (a ActionPlan) Mixin() []ent.Mixin {
...
 	return []schema.Annotation{
-		entx.Features("compliance", "continuous-compliance-automation"),
 		entfga.SelfAccessChecks(),
 	}
 }
 
+// Interceptors of the ActionPlan
+func (a ActionPlan) Interceptors() []ent.Interceptor {
+	return []ent.Interceptor{
+		interceptors.InterceptorFeatures(a.Features()...),
```
golanglemonade: Wondering if we can add the interceptor as a runtime interceptor in the client: https://github.com/theopenlane/core/blob/main/internal/entdb/client.go#L152 so we don't have to remember to add this on every schema? You can use the existing generated `features.go` to lookup the features required for the schema in the interceptor https://github.com/theopenlane/core/blob/main/internal/features/features.go

internal/ent/schema/actionplan.go
```diff
@@ -83,18 +85,31 @@ func (a ActionPlan) Mixin() []ent.Mixin {
...
+	}
+}
+
 // Policy of the ActionPlan
-func (ActionPlan) Policy() ent.Policy {
+func (a ActionPlan) Policy() ent.Policy {
 	return policy.NewPolicy(
 		policy.WithMutationRules(
+			rule.DenyIfMissingAllFeatures(a.Features()...),
```
golanglemonade: Same here; can this just be a base policy? https://github.com/theopenlane/core/blob/main/internal/ent/privacy/policy/base.go#L13

internal/ent/schema/procedure.go
```diff
@@ -74,10 +75,19 @@ func (p Procedure) Mixin() []ent.Mixin {
 	}.getMixins(p)
 }
 
+func (Procedure) Features() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogBaseModule,
```
golanglemonade:
```suggestion
```

internal/ent/schema/risk.go
```diff
@@ -192,19 +194,36 @@ func (r Risk) Mixin() []ent.Mixin {
 	}.getMixins(r)
 }
 
+func (Risk) Features() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogBaseModule,
```
golanglemonade:
```suggestion
```

internal/ent/schema/task.go
```diff
@@ -213,10 +215,19 @@ func (t Task) Edges() []ent.Edge {
 	}
 }
 
+func (Task) Features() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogBaseModule,
+		models.CatalogEntityManagementModule,
```
golanglemonade: I believe tasks should be just base? @matoszz ?
```suggestion
	return []models.OrgModule{
		models.CatalogBaseModule,
```

internal/graphapi/control_clone.go
```diff
@@ -83,6 +83,10 @@ func (r *mutationResolver) cloneControlsFromStandard(ctx context.Context, standa
 		return nil, err
 	}
 
+	if std.IsPublic {
+		ctx = allowCtx
+	}
```
golanglemonade: Why this change?

internal/graphapi/models_test.go
```diff
@@ -462,9 +471,60 @@ func (o *OrganizationBuilder) MustNew(ctx context.Context, t *testing.T) *ent.Or
...
+		assert.NilError(t, err)
+
+		// if no rows were updated, the module doesn't exist - create it
+		if n == 0 {
+			_, err := o.client.db.OrgModule.Create().
+				SetOwnerID(orgID).
+				SetModule(string(feature)).
+				SetActive(true).
+				SetPrice(models.Price{Amount: 0, Interval: "month"}).
+				Save(newCtx)
```
golanglemonade:
```suggestion
			err := o.client.db.OrgModule.Create().
				SetOwnerID(orgID).
				SetModule(string(feature)).
				SetActive(true).
				SetPrice(models.Price{Amount: 0, Interval: "month"}).
				Exec(newCtx)
```

you can use `Exec` instead of `Save` here and it won't return the result

internal/graphapi/models_test.go
```diff
@@ -462,9 +471,60 @@ func (o *OrganizationBuilder) MustNew(ctx context.Context, t *testing.T) *ent.Or
...
+			Relation:    "enabled",
+		}))
+	}
+
+	_, err := o.client.db.Authz.WriteTupleKeys(ctx, tuples, nil)
+	assert.NilError(t, err)
+
+	// create a context with the organization ID properly set for authentication
+	newCtx := auth.NewTestContextWithOrgID(o.UserID, orgID)
+
```
golanglemonade: Why do we need this here? The context passed in should already be authorized for that user and org?

internal/graphapi/organization_test.go
```diff
@@ -126,8 +126,8 @@ func TestQueryOrganizations(t *testing.T) {
 	// so it doesn't interfere with the other tests
 	orgUser := suite.userBuilder(context.Background(), t)
 
-	org1 := (&OrganizationBuilder{client: suite.client}).MustNew(orgUser.UserCtx, t)
-	org2 := (&OrganizationBuilder{client: suite.client}).MustNew(orgUser.UserCtx, t)
+	org1 := (&OrganizationBuilder{client: suite.client, UserID: orgUser.ID}).MustNew(orgUser.UserCtx, t)
```
golanglemonade: We shouldn't need to pass in UserID; we know this from the User context that is passed in.

internal/graphapi/organization_test.go
```diff
@@ -503,12 +503,14 @@ func TestMutationCreateOrganization(t *testing.T) {
...
+			// groups and programs require the compliance module which is not added by default
+			_, err = suite.client.api.GetGroups(newCtx, &testclient.GroupWhereInput{
 				IsManaged: lo.ToPtr(true),
 			})
 
-			// admins, viewers, all users should be created
-			assert.Check(t, is.Len(managedGroups.Groups.Edges, 3))
+			// while group is in the base module, this query includes programs and others
+			// which are in other modules
+			assert.Assert(t, err != nil)
```
golanglemonade: This changes what this test is supposed to be checking; and it no longer confirms the managed groups were created and able to be returned. 

We should still be able to get the groups back, you should just be missing the edges to programs, etc. that weren't enabled? If not; that should be the case, right?

internal/graphapi/seed_test.go
```diff
@@ -67,8 +67,11 @@ func (suite *GraphTestSuite) userBuilder(ctx context.Context, t *testing.T) test
 	// setup user context with the personal org
 	userCtx := auth.NewTestContextWithOrgID(testUser.ID, testUser.PersonalOrgID)
 
+	// ensure the personal org has modules enabled so org-owned interceptors don't block operations
+	(&OrganizationBuilder{client: suite.client, UserID: testUser.ID, Features: features}).enableModules(userCtx, t, testUser.PersonalOrgID)
```
golanglemonade: Personal orgs should only ever have `base` modules enabled, and it should always be enabled by default, right @matoszz? I wouldn't think we should be manually setting this in a test?

internal/httpserve/handlers/invite_test.go
```diff
@@ -58,6 +58,7 @@ func (suite *HandlerTestSuite) TestOrgInviteAcceptHandler() {
 	require.NoError(t, err)
 
 	recipientCtx := auth.NewTestContextWithOrgID(recipient.ID, userSetting.Edges.DefaultOrg.ID)
+	suite.enableModules(recipientCtx, recipient.ID, userSetting.Edges.DefaultOrg.ID, models.AllOrgModules)
```
golanglemonade: This should only require the base module
```suggestion
	suite.enableModules(recipientCtx, recipient.ID, userSetting.Edges.DefaultOrg.ID, models.CatalogBaseModule)
```

internal/ent/privacy/rule/features.go
adelowo: @golanglemonade if stripe isn't enabled as an example, module access will still be usable and enforced with this

golanglemonade: What does stripe have to do with the features being in FGA?

internal/graphapi/seed_test.go
matoszz: correct

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/183

### Changes Preview:

config/config-dev.example.yaml
```diff
@@ -102,29 +102,31 @@ sessions:
...
+    root: https://www.theopenlane.io
+    docs: https://docs.theopenlane.io
+    product: https://console.theopenlane.io
 
 ratelimit:
   enabled: false
 
 entConfig:
+  modules:
+    enabled: true
```
golanglemonade:
```suggestion
    enabled: false
```

Should this be false as the default in dev? Otherwise you always need to have stripe turned on, right?

internal/ent/generate/entc.go
```diff
@@ -361,6 +369,138 @@ func generateEnum(name string, values []string) error {
 	return nil
 }
 
+// generateModulePerSchema walks through the schema files, parses them
+// and generates a static file containing the schema and it's associated modules
+func generateModulePerSchema() error {
```
golanglemonade: Can these either be moved to their own files, or included with entitlements package? It's a bit weird these are all just in the entc.go main file?

internal/ent/generate/entc.go
```diff
@@ -361,6 +369,138 @@ func generateEnum(name string, values []string) error {
...
+
+	data := struct {
+		Entries []moduleEntry
+	}{
+		Entries: entries,
+	}
+
+	if err := tmpl.Execute(file, data); err != nil {
+		return fmt.Errorf("executing feature map template: %w", err)
+	}
```
golanglemonade: Since this is creating a go file; you need to ensure it is properly gofmt, we have other examples of this:

```
			var buf bytes.Buffer

			if err := tmpl.ExecuteTemplate(&buf, "features.tmpl", data); err != nil {
				log.Fatalf("Unable to execute template: %v", err)
			}

			// run gofmt and goimports on the file contents
			formatted, err := imports.Process(filePath, buf.Bytes(), nil)
			if err != nil {
				return fmt.Errorf("%w: failed to format file", err)
			}

			if _, err := file.Write(formatted); err != nil {
				log.Fatalf("Unable to write to file: %v", err)
			}
```

internal/ent/generate/entc.go
```diff
@@ -96,7 +99,6 @@ func main() {
 			genhooks.GenQuery(graphQueryDir),
 			genhooks.GenQuery(graphSimpleQueryDir),
 			genhooks.GenSearchSchema(graphSchemaDir, graphQueryDir),
-			genhooks.GenFeatureMap(featureMapDir),
```
golanglemonade, line 99: Is this existing one going to be deleted from entx?

internal/ent/generate/entc.go
```diff
@@ -361,6 +369,138 @@ func generateEnum(name string, values []string) error {
...
+		return true
+	})
+
+	return schemaName, modules
+}
+
+func extractModules(funcDecl *ast.FuncDecl) []string {
+	var modules []string
+
+	ast.Inspect(funcDecl.Body, func(n ast.Node) bool {
```
golanglemonade: this is a crazy nested function, but I'll assume it works 😅

internal/ent/schema/organization.go
```diff
@@ -122,6 +122,17 @@ func (Organization) Fields() []ent.Field {
 			Annotations(
 				entgql.Skip(entgql.SkipWhereInput, entgql.SkipMutationUpdateInput, entgql.SkipOrderField),
 			),
+		field.String("stripe_customer_id").
```
golanglemonade, line 125: This is on the org subscription table, why are you moving it here?

internal/ent/schema/orgsubscription.go
```diff
@@ -71,7 +70,6 @@ func (OrgSubscription) Fields() []ent.Field {
 			Default(true),
 		field.String("stripe_customer_id").
 			Comment("the customer ID the subscription is associated to").
-			Unique().
```
golanglemonade, line 74: Why doesn't this need to be unique any more?

internal/ent/hooks/errors.go
```diff
@@ -71,6 +71,8 @@ var (
 	ErrFutureTimeNotAllowed = errors.New("time cannot be in the future")
 	// ErrFieldRequired is returned when a field is required but not provided
 	ErrFieldRequired = errors.New("field is required but not provided")
+	// ErrOwnerIDNotExists is returned when an owner_id cannot be found
+	ErrOwnerIDNotExists = errors.New("owner_id is required")
```
golanglemonade, line 75: Instead of making a new error, you can resue the above one with:

```
fmt.Errorf("%w: owner_id", ErrFieldRequired)
```

internal/ent/generate/entc.go
adelowo: yes I removed it already http://github.com/theopenlane/entx/pull/175

internal/ent/schema/orgsubscription.go
adelowo: I was meant to drop it completely.

internal/ent/schema/organization.go
adelowo: We are duplicating the value everywhere. I moved it to `organizations` table instead so we store once

internal/ent/generate/entc.go
adelowo: 😭😭 I didn't want to introduce another format to keep/store the Modules. This seems like the most straightforward way to extract it directly from the schemas

internal/ent/schema/organization.go
adelowo: also it simpifies the stripe client. Currently the client does a search/list to stripe to find the organization. which seems like not so optimal especially since every org will have a unique stripe customer id - even if created with the same email multiple times

internal/ent/generate/entc.go
golanglemonade: yeah I wasnt gonna have you touch it right now...just pretty crazy lol

internal/ent/hooks/organization.go
```diff
@@ -321,16 +315,19 @@ func postOrganizationCreation(ctx context.Context, orgCreated *generated.Organiz
...
+		}
 
-	// create default feature tuples for base functionality when entitlements are enabled
-	if m.Client().EntitlementManager != nil {
-		if err := createFeatureTuples(ctx, m.Authz, orgCreated.ID, modulesCreated); err != nil {
-			zerolog.Ctx(ctx).Error().Err(err).Msg("error creating default feature tuples")
+		// create default entity types, if configured
+		if err := createEntityTypes(ctx, orgCreated.ID, m); err != nil {
+			return err
+		}
```
golanglemonade: This is already done on line 307
```suggestion
```

internal/ent/hooks/orgmodule.go
```diff
@@ -0,0 +1,121 @@
...
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.OrgModuleFunc(func(ctx context.Context, omm *generated.OrgModuleMutation) (generated.Value, error) {
+			if !omm.EntConfig.Modules.Enabled {
+				return next.Mutate(ctx, omm)
+			}
+
+			op := omm.Op()
+
+			if op == ent.OpUpdateOne {
+				if active, _ := omm.Active(); active {
```
golanglemonade: Do we need to add the tuple here? What happens if it goes !Active -> Active?

internal/ent/hooks/orgmodule.go
```diff
@@ -0,0 +1,121 @@
...
+					return next.Mutate(ctx, omm)
+				}
+			}
+
+			id, exists := omm.ID()
```
golanglemonade: Do we need to check `omm.IDs()` if `omm.ID()` doesn't exist?

internal/ent/hooks/orgmodule.go
```diff
@@ -0,0 +1,121 @@
...
+			})
+
+			_, err = omm.Authz.WriteTupleKeys(ctx, nil, []fgax.TupleKey{deleteTuple})
+			if err != nil {
+				return nil, err
+			}
+
+			return v, nil
+		})
+	}, ent.OpUpdateOne|ent.OpDeleteOne)
```
golanglemonade: on my comment above on omm.IDs() would only apply for `ent.Update`; any reason we aren't handle the delete + update cases here?

adelowo: even if an org is subscribed to multiple modules, they come in separate distinct webhooks hence I chose to handle each since we will not be doing bulk updates here

internal/ent/hooks/standard_test.go
```diff
@@ -26,9 +28,24 @@ func (suite *HookTestSuite) TestAddOrDeleteStandardTuple() {
...
+			ObjectType:  "feature",
+			Relation:    "enabled",
+		}),
+	}, nil)
+	require.NoError(t, err)
+
 	ctx := auth.NewTestContextForSystemAdmin(user.ID, orgID)
 	ctx = generated.NewContext(ctx, suite.client)
 
+	ctx = contextx.With(ctx, auth.OrganizationCreationContextKey{})
```
golanglemonade: Why do we need this context key here? This isn't an org creation.

adelowo: not needed actually. this seems to be left over of earlier implementations

internal/ent/interceptors/modules.go
```diff
@@ -0,0 +1,80 @@
...
+
+			entity := pluralize.NewClient().Plural(strings.ToLower(q.Type()))
+
+			path := graphql.GetPath(ctx)
+
+			if len(path) == 0 {
+				path = ast.Path{ast.PathName(entity)}
+			} else if lastName, ok := path[len(path)-1].(ast.PathName); !ok || string(lastName) != entity {
+				path = append(path, ast.PathName(entity))
+			}
```
golanglemonade, line 108: What is this doing? Can you add a comment?

internal/ent/privacy/rule/modules.go
```diff
@@ -0,0 +1,302 @@
...
+	if c := generated.FromContext(ctx); c != nil {
+		client = c
+	} else if tx := transaction.FromContext(ctx); tx != nil {
+		client = tx.Client()
+	}
+
+	var feats []string
+
+	// attempt to use the EntitlementManager
+	if client != nil && client.EntitlementManager != nil {
```
golanglemonade: This was marked as resolved but still see this here and I'm not sure why: https://github.com/theopenlane/core/pull/1079#discussion_r2267297606

internal/ent/privacy/rule/modules.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+			return feats, nil
+		}
+	}
+
+	if c := generated.FromContext(ctx); c != nil {
+		client = c
+	} else if tx := transaction.FromContext(ctx); tx != nil {
+		client = tx.Client()
+	}
```
golanglemonade: same here, comment marked as resolved without a comment or code change? https://github.com/theopenlane/core/pull/1079#discussion_r2267306576

internal/ent/privacy/rule/modules.go
```diff
@@ -0,0 +1,302 @@
...
+
+		if err != nil {
+			return nil, err
+		}
+
+		feats = make([]string, 0, len(modules))
+		for _, m := range modules {
+			feats = append(feats, m.Module.String())
+		}
+	}
```
golanglemonade: I thought we talked about removing this? We should never need to fall back to the db, we should be able to use FGA tuples

internal/ent/privacy/rule/modules.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+		}
+
+		mutationType := m.Type()
+
+		if strings.HasSuffix(mutationType, "History") || ShouldSkipFeatureCheck(ctx) {
+			return privacy.Skip
+		}
+
+		fmt.Println(m.Type(), "MUTATION")
```
golanglemonade:
```suggestion
```

internal/ent/privacy/token/webauthn_token.go
```diff
@@ -0,0 +1,27 @@
+package token
+
+import (
+	"context"
+
+	"github.com/theopenlane/utils/contextx"
+)
+
+type WebauthnCreationContextKey struct {
```
golanglemonade, line 9: I dont see this used?

internal/ent/schema/jobresult.go
```diff
@@ -150,9 +155,13 @@ func (JobResult) Interceptors() []ent.Interceptor {
...
 
 // Policy of the JobResult
-func (JobResult) Policy() ent.Policy {
+func (j JobResult) Policy() ent.Policy {
 	return policy.NewPolicy(
 		policy.WithMutationRules(
+			policy.CheckCreateAccess(),
+			// entfga.CheckEditAccess[*generated.JobResultMutation](),
+			policy.CheckCreateAccess(),
+			// entfga.CheckEditAccess[*generated.JobResultMutation](),
```
golanglemonade:
```suggestion
```

not sure why this change, but its duplicated

internal/ent/schema/risk.go
```diff
@@ -191,17 +192,25 @@ func (r Risk) Mixin() []ent.Mixin {
 	}.getMixins(r)
 }
 
+func (Risk) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogEntityManagementModule,
```
golanglemonade:
```suggestion
		models.CatalogComplianceModule,
		models.CatalogRiskManagementAddon,
```

Confirm with @matoszz but I think we'd only need these two?

internal/ent/schema/scheduledjob.go
```diff
@@ -166,3 +165,9 @@ func (ScheduledJob) Policy() ent.Policy {
 		),
 	)
 }
+
+func (ScheduledJob) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogBaseModule,
```
golanglemonade:
```suggestion
		models.CatalogComplianceModule,
```

internal/ent/schema/subcontrol.go
```diff
@@ -158,8 +159,11 @@ func (Subcontrol) Policy() ent.Policy {
...
-func (Subcontrol) Annotations() []schema.Annotation {
-	return []schema.Annotation{
-		entx.Features("compliance"),
+func (Subcontrol) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogEntityManagementModule,
 	}
```
golanglemonade:
```suggestion
	return []models.OrgModule{
		models.CatalogComplianceModule,
	}
```

You need the compliance module to have controls and subcontrols; cc @matoszz to confirm

internal/ent/schema/control.go
```diff
@@ -184,10 +185,18 @@ func (Control) Policy() ent.Policy {
 	)
 }
 
+func (Control) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogEntityManagementModule,
```
golanglemonade:
```suggestion
		models.CatalogComplianceModule,
```

Believe this should be `compliance` only; cc @matoszz

internal/ent/privacy/token/webauthn_token.go
adelowo: It is used in the webauth handler

internal/ent/schema/program.go
```diff
@@ -173,10 +174,19 @@ func (p Program) Edges() []ent.Edge {
 	}
 }
 
+func (Program) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogBaseModule,
+		models.CatalogEntityManagementModule,
```
golanglemonade: if something is in the base; I dont think it should ever require more than that - all users have the base module. That being said I dont think this is correct. 

```suggestion
		models.CatalogComplianceModule,
		models.CatalogRiskManagementAddon,
```

Programs are pretty well tied to the compliance Module, maybe risk assement addon? @matoszz ?

internal/ent/schema/standard.go
```diff
@@ -171,9 +171,12 @@ func (Standard) Policy() ent.Policy {
...
-func (Standard) Annotations() []schema.Annotation {
-	return []schema.Annotation{
-		entx.Features("compliance"),
+func (Standard) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogBaseModule,
+		models.CatalogEntityManagementModule,
```
golanglemonade, line 180:
```suggestion
		models.CatalogComplianceModule,
		models.CatalogTrustCenterModule,
```

I think this should be these two? cc @matoszz

internal/ent/schema/task.go
```diff
@@ -240,3 +239,12 @@ func (Task) Policy() ent.Policy {
 		),
 	)
 }
+
+func (Task) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+		models.CatalogPolicyManagementAddon,
+		models.CatalogRiskManagementAddon,
+		models.CatalogEntityManagementModule,
```
golanglemonade: this should be in the base I think
```suggestion
	return []models.OrgModule{
		models.CatalogBaseModule,
   }
```

internal/graphapi/apitoken_test.go
```diff
@@ -321,8 +314,6 @@ func TestMutationDeleteAPIToken(t *testing.T) {
 
 			if tc.errorMsg != "" {
 				assert.ErrorContains(t, err, tc.errorMsg)
-				assert.Check(t, is.Nil(resp))
```
golanglemonade, line 324: What is the resp now if not nil?

internal/graphapi/control_clone.go
```diff
@@ -104,6 +106,25 @@ func (r *mutationResolver) cloneControls(ctx context.Context, controlsToClone []
...
 
+	ac, err := auth.GetAuthenticatedUserFromContext(ctx)
+	if err != nil || ac.OrganizationID == "" {
+		return nil, rout.NewMissingRequiredFieldError("owner_id")
+	}
+
+	orgID := ac.OrganizationID
+
+	// check if the organization has the required modules for Control entities before the parallel execution
+	// this prevents the database connection issues that occur when multiple goroutines all try to check modules
```
golanglemonade: what connection issues? is this something that could happen in prod?

internal/httpserve/handlers/invite_test.go
```diff
@@ -142,9 +146,9 @@ func (suite *HandlerTestSuite) TestOrgInviteAcceptHandler() {
 
 			assert.Equal(t, testUser1.OrganizationID, user.User.Setting.DefaultOrg.ID)
 
-			// Test the user was added to the group
+			// test the user was added to the group
 			group, err := suite.api.GetGroupByID(recipientCtx, group.ID)
-			require.NoError(t, err)
+			require.ErrorContains(t, err, interceptors.ErrFeatureNotEnabled.Error())
```
golanglemonade, line 151: Why do we want this error check in the test? Isn't this part of the base module?

internal/ent/privacy/token/webauthn_token.go
golanglemonade: yeah sorry, found it later. it was hidden cause github :-D

internal/httpserve/handlers/webhook.go
```diff
@@ -436,38 +459,69 @@ func (h *Handler) syncOrgSubscriptionWithStripe(ctx context.Context, subscriptio
...
+
+		if err := h.clearFeatureCache(ctx, orgSubscription.OwnerID, stripeSub.FeatureLookupKeys); err != nil {
+			log.Error().Err(err).Msg("failed to ensure feature tuples")
+		}
 	}
 
+	return &orgSubscription.OwnerID, h.syncSubscriptionItemsWithStripe(ctx, subscription)
+}
+
+func (h *Handler) clearFeatureCache(ctx context.Context, orgID string, _ []string) error {
```
golanglemonade: should the last param just be removed? looks like its not used?

internal/ent/schema/organization.go
golanglemonade: talked in slack; good here

internal/ent/schema/organization.go
```diff
@@ -122,6 +122,17 @@ func (Organization) Fields() []ent.Field {
...
+			Comment("the customer ID this organization is associated to").
+			Optional().
+			Nillable().
+			Unique().
+			// Immutable().
```
golanglemonade:
```suggestion
```

internal/ent/schema/organization.go
```diff
@@ -122,6 +122,17 @@ func (Organization) Fields() []ent.Field {
...
+			Annotations(
+				entgql.Skip(entgql.SkipMutationCreateInput |
+					entgql.SkipWhereInput | entgql.SkipOrderField),
+			).
```
golanglemonade: should this have a skip on update as well? Customers shouldn't be updating this via the API

internal/ent/schema/organization.go
```diff
@@ -122,6 +122,17 @@ func (Organization) Fields() []ent.Field {
...
+			Optional(),
```
golanglemonade:
```suggestion
```
This is here twice

internal/ent/schema/organization.go
```diff
@@ -122,6 +122,17 @@ func (Organization) Fields() []ent.Field {
 			Annotations(
 				entgql.Skip(entgql.SkipWhereInput, entgql.SkipMutationUpdateInput, entgql.SkipOrderField),
 			),
+		field.String("stripe_customer_id").
+			Comment("the customer ID this organization is associated to").
```
golanglemonade:
```suggestion
			Comment("the stripe customer ID this organization is associated to").
```

internal/httpserve/handlers/invite_test.go
adelowo: it is part of the base module but it calls programs and the permission nodes which are not part of the the base module.

the assertion on the group is still there and works correctly for the member checks

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/185

### Changes Preview:

internal/graphapi/control_clone.go
adelowo: there was an issue locally where the db gets overloaded. but it is unrelated. removed comment now

golanglemonade (approved):

PR #1126, opened 2025-08-22
Orgmodule hook: fix wrong return type

adelowo:
remove sub.created since we actually already reconcile the subscription on customer.subscription.updated.
it is actually not needed and might have the unintented consequence of duplicating the subscriptions ( also important to note that
we will only ever have one subscription and multiple prices inside that
subscription).

also fix return a bug where the return object on updating/deleting
tuples is not the expected type

```txt
 Error error=unexpected node type <nil> returned from *generated.OrgModuleMutation
```


adelowo:
This handler was only added in the previous pr

golanglemonade (approved):

PR #1129, opened 2025-08-22
Fix riverboat not starting up

adelowo:


docker/docker-compose-riverboat.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: ghcr.io/theopenlane/riverboat:${RIVERBOAT_TAG:-arm64-latest}
     pull_policy: always
     volumes:
-      - ${PWD}/docker/configs/riverboat/.config.yaml:/config/.config.yaml
+      - ${PWD}/docker/configs/riverboat/.config.dev.yaml:/config/.config.yaml
```
golanglemonade, line 8: This shouldn't change; then someone might accidentally commit secrets; you should copy this to `.config.yaml` locally 
```suggestion
      - ${PWD}/docker/configs/riverboat/.config.yaml:/config/.config.yaml
```

adelowo: closing this. thanks

PR #1135, opened 2025-08-25
Sync organization's stripe customer id

adelowo:


golanglemonade (approved):

internal/httpserve/handlers/webhook_helpers.go
```diff
@@ -56,6 +63,35 @@ func (h *Handler) syncSubscriptionItemsWithStripe(ctx context.Context, sub *stri
...
+	}
+
+	if org.StripeCustomerID != nil && *org.StripeCustomerID != "" {
+		return nil
+	}
+
+	return tx.Organization.Update().Where(organization.ID(orgSub.OwnerID)).
+		SetStripeCustomerID(customerID).
+		Exec(allowCtx)
+}
```
golanglemonade: I mean this put this at the top; so we don't run a org query when we don't need to: 

```suggestion
func upsertOrgStripeCustomer(ctx context.Context, orgSub *ent.OrgSubscription, customerID string) error {
	if customerID == "" {
		return nil
	}

	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)

	tx := transaction.FromContext(ctx)

	org, err := tx.Organization.Query().Select(organization.FieldStripeCustomerID).
		Where(organization.ID(orgSub.OwnerID)).Only(ctx)
	if err != nil {
		return err
	}

	if org.StripeCustomerID != nil && *org.StripeCustomerID != "" {
		return nil
	}

	return tx.Organization.Update().Where(organization.ID(orgSub.OwnerID)).
		SetStripeCustomerID(customerID).
		Exec(allowCtx)
}
```

golanglemonade (approved):

PR #1140, opened 2025-08-26
update graph client + add cli commands for custom domains

acookin:
* add `get` cli for mappable domains
* add `delete` cli for custom domains
* put DNS verification info in the client


cmd/cli/cmd/trustcenter/create.go
```diff
@@ -38,6 +39,11 @@ func createValidation() (input openlaneclient.CreateTrustCenterInput, err error)
 		input.CustomDomainID = &customDomainID
 	}
 
+	orgID := cmd.Config.String("org-id")
+	if orgID != "" {
+		input.OwnerID = &orgID
+	}
```
golanglemonade, line 45: This should only be needed in the case of using a personal access token - otherwise we auto-set this based on the context of the user; we should probably add this add the top level instead of per object; but I can do that later.

golanglemonade (approved):

cmd/cli/cmd/trustcenter/create.go
acookin: Yeah, that was the exact scenario i ran into (re: PAT)

PR #1150, opened 2025-08-27
Block all updates to archived programs

adelowo:
Only allow passthrough if the update contains a status update that moves it from archived to another status

### DIFF
--- a/internal/ent/hooks/program.go
+++ b/internal/ent/hooks/program.go
@@ -17,9 +17,16 @@ import (
 )
 
 // HookProgramAuthz runs on program mutations to setup or remove relationship tuples
+// and prevents updates to archived programs - except if the update contains status changes too
 func HookProgramAuthz() ent.Hook {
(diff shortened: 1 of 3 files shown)


internal/ent/hooks/program.go
```diff
@@ -132,3 +139,22 @@ func programDeleteHook(ctx context.Context, m *generated.ProgramMutation) error
...
+
+	program, err := m.Client().Program.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	status, exists := m.Status()
+	// if status is not getting updated and the current status is archived
+	// prevent all updates
+	if (!exists && program.Status == enums.ProgramStatusArchived) || status == enums.ProgramStatusArchived {
```
golanglemonade, line 155: you should be able to set the status to !Archived - all other updates should be blocked.

golanglemonade (approved):

PR #1188, opened 2025-09-04
bug: create subs with entitlement manager on, not module enforcement; sandbox; personal org

golanglemonade:
- creates subscriptions based on the entitlement manager; not module enforcement
- module enforcement should only be on the response side of query/mutation - not gating the creation of subs
- removes subscription creation for personal orgs 
- adds helper functions for "modules enabled" and "entitlement manager enabled" - these are purely for code readable more than anything else
- fixes account/features endpoint to use fga instead of db query
- other random things I had on my branch: 
    - task command to just create the required docker containers and not everything 
- fixes a bug with get control by ref code on the cli
- sets a sandbox env so you don't have to do some file copying; instead you can set a config var:
```
entConfig:
  modules:
    enabled: true
    useSandbox: true
```
- fixes a bug in the catalog cli that would error on creation of product + prices if the price wasn't found
- removes unused price ids from the config
- adds `GenerateModulePerSchema` back to entc, not sure how that got removed
- udpates schemas so that if it really only requires the base module, it doesn't have other modules listed since all organizations will have a base module included
- updates interceptors to only require _one_ of the modules, not all on a schema


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/255

### Changes Preview:

pkg/catalog/catalog_sandbox.yaml
```diff
@@ -11,14 +11,14 @@ modules:
           unit_amount: 0
           nickname: base_module_monthly
           lookup_key: base_module_monthly
-          price_id: price_1RjUGVQrAxy77qa71BBYvS0S
+          price_id: price_1S3qXEJIzM4Pa2Zc3bZXrIC8
```
matoszz, line 14: Any reason why we need to change these since the Development sandbox already had existing price ID's (which were the previous values?)

pkg/catalog/catalog_sandbox.yaml
```diff
@@ -11,14 +11,14 @@ modules:
...
         - interval: year
           unit_amount: 0
           nickname: base_module_free
           lookup_key: base_module_free
-          price_id: price_1Rj7WKQrAxy77qa7zm8okFhE
+          price_id: price_1S3qXEJIzM4Pa2ZcYaKVbHkk
     audience: public
-    product_id: prod_SeQNFK7TFhCH5K
+    product_id: prod_Szq52Y33Hzvxc6
     personal_org: true
```
matoszz, line 22: this should be removed if you're also removing it from the gencatalog

pkg/catalog/genyaml/yamlgen.go
```diff
@@ -262,16 +273,12 @@ func writeModuleConstants(path string, c catalog.Catalog) error {
 	for k, v := range c.Addons {
 		modKey := string(k)
 		modKeys = append(modKeys, modKey)
-		if v.PersonalOrg {
```
matoszz, line 265: This isn't a "bad" way to do this but the way the code was previously structured, if you had wanted to remove creating the base module (or any module) for personal orgs all you'd have needed to do is remove / unset the bool flag in the catalog.yaml which would have then bypassed the creation. This will remove all of our available toggles + generated types, so if there's a possibility we may want to in the future do something with personal organizations on this front it's worth considering leaving some of the code in-place

pkg/entitlements/client.go
```diff
@@ -16,6 +16,12 @@ type StripeClient struct {
 	backends *stripe.Backends
 }
 
+// Enabled checks if the entitlements feature is enabled based on the status of the Stripe client settings
+// this is a helper function for more readable code
+func Enabled(s *StripeClient) bool {
```
matoszz: The `Config` struct used in the `StripeClient` already has enabled bool value - is there some reason we wouldn't use that (or a bool value on the client itself) rather than having an "enabled" check potentially bypass the actual underlying config? (e.g. you could have the config parameter set to "false" but if the client was instantiated in general we'd be returning that it is enabled when the setting itself is not)

internal/ent/privacy/rule/modules.go
```diff
@@ -16,13 +16,19 @@ import (
...
+	"github.com/theopenlane/core/pkg/entitlements"
 	"github.com/theopenlane/core/pkg/models"
 	"github.com/theopenlane/core/pkg/permissioncache"
 )
 
 // HasFeature reports whether the current organization has the given feature enabled
 func HasFeature(ctx context.Context, feature string) (bool, error) {
-	feats, err := orgFeatures(ctx)
+	// all organizations have access to the base module
+	if feature == models.CatalogBaseModule.String() {
```
matoszz, line 27: do we have checks in other locations for things like whether or not the org is active / inactive, etc.? We don't "revoke" feature access from a subscription as a rule of thumb, which means we rely on some cascading actions to occur, e.g. "when org is deleted / set to inactive, also inactivate associated subscription"

internal/ent/hooks/organization.go
```diff
@@ -623,16 +617,8 @@ func createDefaultOrgModulesProductsPrices(ctx context.Context, orgCreated *gene
 	modulesCreated := make([]string, 0)
 
 	// the catalog contains config for which things should be in a trial, or added for a personal org
```
matoszz:
```suggestion
	// the catalog contains config for which things should be in a trial
```

internal/ent/hooks/organization.go
```diff
@@ -315,7 +316,8 @@ func postOrganizationCreation(ctx context.Context, orgCreated *generated.Organiz
 		return err
 	}
 
-	if m.EntConfig.Modules.Enabled {
+	// create subscriptions if the entitlement manager is enabled
+	if entitlements.Enabled(m.EntitlementManager) {
```
matoszz: I think we should check the `Config` struct with the `enabled` flag on it rather than rely on whether or not the client is instantiated

internal/httpserve/authmanager/authmanager.go
```diff
@@ -138,14 +139,14 @@ func (a *Client) GenerateOauthAuthSession(ctx context.Context, w http.ResponseWr
 }
 
 // checkActiveSubscription checks if the organization has an active subscription
-func (a *Client) checkActiveSubscription(ctx context.Context, orgID string) (active bool, err error) {
-	// if the entitlement manager is disabled or modules not enabled, we can skip the check
-	if !a.GetDBClient().EntConfig.Modules.Enabled || a.db.EntitlementManager == nil {
+func (a *Client) checkActiveSubscription(ctx context.Context, orgID string) (active bool, err error) { //nolint:unused
+	// if the entitlement manager is disabled, we can skip the check
+	if !entitlements.Enabled(a.GetDBClient().EntitlementManager) {
```
matoszz: Repeating comments left in other locations, but I think we should be using the `Config` struct `enabled` parameter for this check

matoszz: that, or we should be checking both that the configuration is enabled and that the client is present*

internal/httpserve/authmanager/authmanager.go
```diff
@@ -286,13 +286,6 @@ func (a *Client) authCheck(ctx context.Context, user *generated.User, orgID stri
 		orgID = au.OrganizationID
 	}
 
-	active, err := a.checkActiveSubscription(ctx, orgID)
```
matoszz, line 289: is there some reason we want to remove this? It seems like we would want to ensure there's an active subscription

internal/httpserve/authmanager/authmanager.go
```diff
@@ -309,7 +302,7 @@ func (a *Client) authCheck(ctx context.Context, user *generated.User, orgID stri
 	}
 
 	// if the org is active and they are allowed, we can return the org
-	if active && allow {
+	if allow {
```
matoszz, line 304: same qq as above - why would we not want to check for an active subscription?

internal/httpserve/handlers/login_test.go
```diff
@@ -164,13 +157,13 @@ func (suite *HandlerTestSuite) TestLoginHandler() {
...
 			expectedOrgID:  invalidConfirmedUserRestrictedOrg.PersonalOrgID,
 		},
-		{
-			name:           "inactive org, switch to personal org",
-			username:       userWithInactiveDefaultOrg.UserInfo.Email,
-			password:       validPassword,
-			expectedStatus: http.StatusOK,
-			expectedOrgID:  userWithInactiveDefaultOrg.PersonalOrgID,
-		},
+		// {
```
matoszz: just delete?

internal/httpserve/handlers/seed_test.go
```diff
@@ -127,7 +127,7 @@ func (suite *HandlerTestSuite) userBuilderWithInput(ctx context.Context, input *
 
 	// add dummy subscription to the organization
 	err = suite.db.OrgSubscription.Update().Where(orgsubscription.OwnerID(testOrg.ID)).
-		SetFeatureLookupKeys(dummyFeatures).
+		SetFeatureLookupKeys([]string{"base_module"}).
```
matoszz: instead of changing this to be "base_module" I'd consider refactoring the test to be more specific to our actual use case - we previously stored the `FeatureLookupKeys` on the subscription in a string map but that (along with our "Tier" fields) would be going away in favor of organization modules, so the refactor would look more like we're running updates on the OrgModules `where` OrgSubscription` has ownerID(testOrg.ID)

pkg/catalog/Taskfile.yaml
```diff
@@ -21,3 +21,11 @@ tasks:
...
       - task: genjsonschema
       - task: genyaml
+
+  all:sandbox:
+    desc: runs all the generation tasks for the sandbox catalog
+    env:
+      OPENLANE_CATALOG_SANDBOX: "true"
+    cmds:
+      - task: genjsonschema
+      - task: genyaml
```
matoszz: It's not ideal to force-set an environment variable that should be derived from our configuration parameters; if I accidentally ran this Task command while I had our stripe production key loaded, it could create all sorts of problems. Task provides numerous methods of gating task execution such as `preconditions` or `requires` or `deps`. We want to structure our Tasks also so that we can have fully generic commands / task references and then conditionally perform actions based on environment settings rather than having fully duplicated task references with overridden parameters (lots of good docs on the many ways you can accomplish this, [ref](https://taskfile.dev/docs/guide#referencing-other-variables)). The example being, I'm always running `task catalog:all` and based on what's in my `.config.yaml` the behavior changes if I set in the `.config.yaml` to use the sandbox flag or not, rather than having multiple tasks that will run regardless of what my underlying configuration is.

I would delete the `all:sandbox` entirely and modify the `all` task to use the set environment variable, e.g.

```suggestion

  all:
    desc: runs all the generation tasks
    vars:
      OPENLANE_CATALOG_SANDBOX: "{{.OPENLANE_CATALOG_SANDBOX}}"
    cmds:
      - task: genjsonschema
      - task: genyaml
```

or if you wanted a different approach because cognitively you just want a Task command that's "specific" for sandbox (although I think this is not ideal), I would use `requires`, ex:

```
  all:sandbox:
    desc: runs all the generation tasks
    cmds:
      - task: genjsonschema
      - task: genyaml
    requires:
      vars:
        - name: OPENLANE_CATALOG_SANDBOX
          enum: [true]
```

This way you're still depending on the actual underlying configuration parameters in our `.config.yaml` rather than force-setting a value that possibly conflicts with the underlying configuration. Environment variables specified directly at the task-level will override variables defined in dotfiles / variables sourced from the environment, which is a problematic pattern.

pkg/entitlements/subscriptions.go
```diff
@@ -149,19 +147,29 @@ func (sc *StripeClient) MapStripeSubscription(ctx context.Context, subs *stripe.
...
+		}
+
 		productID = item.Price.Product.ID
 
 		product, err := sc.GetProductByID(ctx, productID)
 		if err != nil {
 			log.Warn().Err(err).Msg("failed to get product by ID")
 		}
 
+		interval := "month"
```
matoszz, line 166: why statically assign this?

matoszz (changes requested):

pkg/entitlements/client.go
golanglemonade: I just moved the existing check to a wrapped function; the current setup doesn't allow for this because this is `nil` if its not enabled; that flag is used for creation of this struct instead. I can change that though

internal/httpserve/authmanager/authmanager.go
golanglemonade: This is the code path that prevents you from logging or switching into an org with an expired subscription. You said previously thats not what we wanted because then "how can you update your billing if your sub lapsed accidentally.

internal/httpserve/handlers/login_test.go
golanglemonade: I was waiting on confirmation of this: https://github.com/theopenlane/core/pull/1188#discussion_r2324776207 so intentionally just commented out for now

internal/httpserve/handlers/seed_test.go
golanglemonade: There are tests that already cover that; This was left because this branch was started as "make sure I can even create subscriptions without module enforcement"

pkg/catalog/Taskfile.yaml
golanglemonade: This doesn't come from config.yaml though? It's not part of the actual config - it is instead just env vars?

pkg/entitlements/subscriptions.go
golanglemonade: I was chasing null pointer errors/panics in tests based on mocks not having all the data. This _shouldnt_ be hit in prod unless stripe changes the structure but was looking to prevent panics

pkg/catalog/genyaml/yamlgen.go
golanglemonade: Just here in the catalog gen? Or in the entitlements package as well?

internal/ent/privacy/rule/modules.go
golanglemonade: I'm not really sure I understand the question here based on the context of base module. We had talked about that a customer would need access to their organization if they subscription lapsed; otherwise they couldn't go back in to setup their billing info. 

If someone's subscription to the paid module lapsed; I assume (I haven't confirmed this myself yet) that the module would be removed so the only thing they could access would be things in the base module? 

The code here is _just_ assuming that any organization would have the base module so we don't go check FGA every time (and dont require a personal org to have an actual sub in stripe)

internal/httpserve/authmanager/authmanager.go
golanglemonade: the problem was that since we had the client (`EntitlementManager`) with the config under it; were werent `EntitlementManager` is nil if its not enabled. Pushing a fix to create the `EntitlementManager` so you can check the config.

pkg/catalog/catalog_sandbox.yaml
golanglemonade: added back to gencatalog

internal/httpserve/handlers/seed_test.go
golanglemonade: Dont think this is needed anymore now that I setup the entitlements in tests to be created; removing

pkg/catalog/genyaml/yamlgen.go
golanglemonade: added back to gen catalog

internal/ent/privacy/rule/modules.go
matoszz: We're calling `GetOrgFeatures` which performs an fga `list` operation for the organization and whether or not it has a mapping to the feature module. However, if a module were to be removed (or expire, cancel) as it stands, the webhook receiver is primarily using `syncOrgSubscriptionWithStripe` which does not disassociate OrgModules in our system with subscriptions (nor does it revoke or delete any feature tuple). I could be wrong, but looking at the way this is constructed, we only ever add tuples for the modules but never actually remove or delete the tuples.

pkg/catalog/Taskfile.yaml
matoszz: then what purpose does `CORE_ENTCONFIG_MODULES_USESANDBOX` have? (I used the wrong environment variable in my example - this^ is the one I was referring to, which would be the one you'd ostensibly be setting if you wanted to use sandbox configuration vs. not

golanglemonade: CORE_ENTCONFIG_MODULES_USESANDBOX is used to determine which go struct to use when running the server. This cli doesn't reference config.yml at all.

matoszz (dismissed):

pkg/catalog/Taskfile.yaml
golanglemonade: updated the taskfile to only include one command again

golanglemonade: I went with defaulting to using sandbox so you have to intentionally specify the prod setup

PR #1200, opened 2025-09-08
Fix org deletion

adelowo:
When deleting an org, it is an `OpDelete` operation so there wouldn't be an id, so use all the existing ids.

> didn't want to add a BulkDelete api to the storage client as only S3 really supports it. others will still need a loop within them

### DIFF
--- a/internal/httpserve/common/doc.go
+++ b/internal/httpserve/common/doc.go
@@ -1,2 +1,4 @@
 // Package common provides common utilities and types used across the HTTP server package
+//
+//revive:disable:var-naming
 package common


internal/ent/hooks/file.go
```diff
@@ -19,31 +19,48 @@ func HookFileDelete() ent.Hook {
...
 				v, err := next.Mutate(ctx, m)
 				if err != nil {
 					return nil, err
 				}
 
-				if storagePath != "" {
-					if err := m.ObjectManager.Storage.Delete(ctx, storagePath); err != nil {
+				for _, id := range ids {
+
+					file, err := m.Client().File.Get(ctx, id)
```
golanglemonade: I would get all the `files` at once, and then do the look to delete the storage, would cut down on the `Get` query when cascade deleting.

golanglemonade (approved):

PR #1232, opened 2025-09-11
feat: add system owned fields, @hidden and @readOnly directives

golanglemonade:
How it started:

> When creating mapped controls, I realized we have no way to reference where they came from later if we want to update the same one. Added these fields are internal only fields so we can track  an internal id and notes on sources for use later. 

> Also more fixes for stripe mocks, this time in the graphapi tests. 

How it ended:

- Updates systemowned_mixin to add additional fields:
    - systemInternalID
    - internalNotes
    
Example: 
As a regular user:
```json
        {
          "node": {
            "id": "01K5HG9THB15FQPAS1R3VW6Y2G",
            "systemInternalID": null,
            "framework": "NIST-SP-800-53-r5"
          }
        },
```

As a system-admin:
```json
        {
          "node": {
            "id": "01K5HG9THB15FQPAS1R3VW6Y2G",
            "systemInternalID": "testing",
            "framework": "NIST-SP-800-53-r5"
          }
        },
```
- Automatically adds directives for hiding the fields in query responses for non-admins with `@hidden`
- Automatically adds directives to mark fields in the mutation as `@readOnly` for non-admins
- Adds the systemowned_mixin to additional schemas
- If the systemowned mixin is there, it will automatically add the `AllowEmptyForSystemAdmin` on the object owned mixin if that is also present
- Updates logs in hooks to use the context logger
- README updated because I was trying to update startup docs in our docs, only to realize it actually pulls from here 

Follow-ups:
- Look into adding a traverse function for all of the system owned schemas to consolidate this logic:

```go
		orgIDs, err := auth.GetOrganizationIDsFromContext(ctx)
		if err != nil {
			return err
		}

		systemStandardPredicates := []predicate.Standard{
			standard.OwnerIDIsNil(),
			standard.SystemOwned(true),
		}

		admin, err := rule.CheckIsSystemAdminWithContext(ctx)
		if err != nil {
			return err
		}

		if !admin {
			// if the user is a not-system admin, restrict to only public standards
			systemStandardPredicates = append(systemStandardPredicates, standard.IsPublic(true))
		}

		// filter to return system owned standards and standards owned by the organization
		q.Where(
			standard.Or(
				standard.And(
					systemStandardPredicates...,
				),
				standard.OwnerIDIn(orgIDs...),
			),
		)
```
- Hide `@hidden` fields from the graphql apollo explorer/introspection as well


README.md
```diff
@@ -74,6 +74,20 @@ See the [README](/config/README.md) in the `config` directory.
...
    defaults inside should allow you to run the server without a problem
 
+> [!TIP]
+>   When developing locally, unless you are specifically testing subscriptions and module enforcement, ensure the following settings are set to false:
+>
+>   ```yaml
+>    entConfig:
+>      modules:
+>        enabled: false
+>        useSandbox: true
```
matoszz: wouldn't they want to set useSandbox also to false? Because that setting still triggers the calls to stripe and if you don't have the sandbox API key, you'd get errors when using it "regularly". I would assume we would want all of these to be false under the conditions "when developing locally, unless you are specifically testing subscriptions and module enforcement" - vs. "if you're developing locally and want to test the subscriptions and module enforcement, you'd need to have a stripe account and stripe API key" etc.?

internal/ent/generate/entc.go
```diff
@@ -60,18 +61,26 @@ func main() {
...
 	if err != nil {
 		log.Fatal().Err(err).Msg("creating entx extension")
 	}
+	schemaHooks = append(schemaHooks, xExt.GQLSchemaHooks()...)
+
+	dExt, err := directives.NewExtension()
+	if err != nil {
+		log.Fatal().Err(err).Msg("creating directives extension")
+	}
+	schemaHooks = append(schemaHooks, dExt.SchemaHooks()...)
```
matoszz, line 78:
```suggestion
	}
	
	schemaHooks = append(schemaHooks, dExt.SchemaHooks()...)
```

internal/ent/generate/entc.go
```diff
@@ -60,18 +61,26 @@ func main() {
 
 	// generate the history schemas
 	historyExt, entfgaExt := preRun()
+	schemaHooks := []entgql.SchemaHook{}
 
 	xExt, err := entx.NewExtension(entx.WithJSONScalar())
 	if err != nil {
 		log.Fatal().Err(err).Msg("creating entx extension")
 	}
+	schemaHooks = append(schemaHooks, xExt.GQLSchemaHooks()...)
```
matoszz, line 71:
```suggestion
	}
	
	schemaHooks = append(schemaHooks, xExt.GQLSchemaHooks()...)
```

internal/ent/hooks/program.go
```diff
@@ -130,7 +129,7 @@ func programDeleteHook(ctx context.Context, m *generated.ProgramMutation) error
 	zerolog.Ctx(ctx).Debug().Str("object", object).Msg("deleting relationship tuples")
 
 	if err := m.Authz.DeleteAllObjectRelations(ctx, object, userRoles); err != nil {
-		log.Error().Err(err).Msg("failed to delete relationship tuples")
+		zerolog.Ctx(ctx).Error().Err(err).Msg("failed to delete relationship tuples")
```
matoszz: could we maybe make this a `Msgf` with what the tuples are / the objects they are for? the generic message doesn't do much good without some additional context

internal/ent/hooks/trustcentersetting.go
```diff
@@ -167,7 +166,7 @@ func trustCenterSettingDeleteHook(ctx context.Context, m *generated.TrustCenterS
 	zerolog.Ctx(ctx).Debug().Str("object", object).Msg("deleting relationship tuples")
 
 	if err := m.Authz.DeleteAllObjectRelations(ctx, object, userRoles); err != nil {
-		log.Error().Err(err).Msg("failed to delete relationship tuples")
+		zerolog.Ctx(ctx).Error().Err(err).Msg("failed to delete relationship tuples")
```
matoszz: same comment as the other log lines with relationship tuples

internal/ent/privacy/rule/invite.go
```diff
@@ -47,15 +47,15 @@ func CanInviteUsers() privacy.InviteMutationRuleFunc {
...
+		zerolog.Ctx(ctx).Debug().Interface("tuple", ac).Msg("checking relationship tuples")
 
 		access, err := m.Authz.CheckOrgAccess(ctx, ac)
 		if err != nil {
 			return privacy.Skipf("unable to check access, %s", err.Error())
 		}
 
 		if access {
-			log.Debug().Str("relation", relation).Str("organization_id", oID).Msg("access allowed")
+			zerolog.Ctx(ctx).Debug().Str("relation", relation).Str("organization_id", oID).Msg("access allowed")
```
matoszz, line 58: We might want to make some of these `Info` so in production we have some form of logging around the checks?

internal/graphapi/directives/extension.go
```diff
@@ -0,0 +1,101 @@
...
+// SchemaHooks of the extension to seamlessly edit the final gql interface
+func (e *Extension) SchemaHooks() []entgql.SchemaHook {
+	return []entgql.SchemaHook{hook}
+}
+
+// hook is used to add the @readOnly directive to input fields that are marked as @hidden
+// this prevents hidden fields from being set in create and update mutations
+// as of today, there is no way to annotate a schema to do this automatically so we use a schema
+// hook to modify the generated schema
+var hook = func(_ *gen.Graph, s *ast.Schema) error {
```
matoszz: just a nit, but a more descriptive variable name would be nice

internal/graphapi/directives/extension.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, t := range s.Types {
+		object := s.Types[getInputObjectName(t.Name)]
+		if object == nil {
+			continue
+		}
+
+		// if the type is an input object, we need to check its fields for directives
+		if t.Kind == ast.InputObject {
```
matoszz: should this happen prior to the function above it? It seems like this is being called on most schema operations and our filter for applying them is based on `t.Kind` so just thinking from a performance perspective if it would be beneficial to evaluate this first ? It's difficult to tell whether or not they have to be in this order, so if that's the case you can ignore this comment

README.md
golanglemonade: I don't think it matters - this is just which file is used for the prices, when modules + subscriptions are off its not used at all. I think it's better to just remove this from this comment and use the defaults so it's less confusing.

internal/ent/hooks/program.go
golanglemonade: Added the `object` to the log like the debug logs

internal/ent/privacy/rule/invite.go
golanglemonade: I actually previously made these all debug because they are noisy and not generally useful to see an "allow" log in most of these cases, unless you are changing the logic + doing development work around these.

matoszz (approved):

PR #1238, opened 2025-09-13
Test sso implementation before enforcing

adelowo:


internal/httpserve/handlers/sso.go
```diff
@@ -26,6 +26,14 @@ import (
 	"github.com/theopenlane/utils/rout"
 )
 
+const (
+	ssoTestCookieName  = "is_test"
+	ssoTestCookieValue = "1"
+
+	ssoSwitchOrgCookieName  = "switch_org"
```
golanglemonade: why do we need different cookies for test + switch? Why can't they follow the same code path as regular sso?

internal/ent/hooks/orgsetting_sso.go
```diff
@@ -35,6 +45,13 @@ func ValidateIdentityProviderConfig(ctx context.Context, m *generated.Organizati
 		return nil
 	}
 
+	// sso connection must have been tested
+	// before it can be can be enforced
+	tested, _ := m.OldIdentityProviderAuthTested(ctx)
```
golanglemonade, line 50: This function is called on all operations: `hook.HasOp(ent.OpCreate|ent.OpUpdate|ent.OpUpdateOne),` but ` m.OldIdentityProviderAuthTested(ctx)` is only valid on `OpUpdateOne` ops and will error when that isn't the case (but the error is being ignored here as well). Will `ent.OpUpdate` be used here? What happens on create has `enforced=true`?

internal/ent/hooks/orgsetting_sso.go
```diff
@@ -133,3 +140,51 @@ func fallbackString(primary func() (string, bool), fallback func() (*string, err
 
 	return "", false
 }
+
+// disableEnforcementIfConfigChanged makes sure if any value changes in the idp config,
+// the enforcement is disabled. The "tested" value is turned off too so the user/org is forced
+// to test the connection before they can enable it again
+func disableEnforcementIfConfigChanged(ctx context.Context, m *generated.OrganizationSettingMutation) error {
+
+	if didSSOConfigChange(ctx, m) || m.Op() == ent.OpCreate {
```
golanglemonade: These shoudl be switched, so you can skip the more expensive check `didSSOConfigChange` when it is a createOp
```suggestion
	if  m.Op() == ent.OpCreate || didSSOConfigChange(ctx, m){
```

internal/ent/hooks/orgsetting_sso.go
```diff
@@ -133,3 +140,51 @@ func fallbackString(primary func() (string, bool), fallback func() (*string, err
...
+		m.SetIdentityProviderLoginEnforced(false)
+		m.SetIdentityProviderAuthTested(false)
+	}
+
+	return nil
+}
+
+func didSSOConfigChange(ctx context.Context, m *generated.OrganizationSettingMutation) bool {
```
golanglemonade, line 157: Same question here, this is called on OpUpdate, but all the `OldX` functions will error except on `OpUpdateOne`

internal/ent/hooks/orgsetting_sso.go
adelowo: we check if enforced first. https://github.com/theopenlane/core/pull/1238/files/b16c314baf29831e7c3363382d26dbca54f271b9#diff-d8c483519cee0ecdb6753a4dc03a114c590f874a5753ec6b823becbf57a4fca0R42-R46

if it is not enforced, the validation does not kick in and returns early. Does not get here

adelowo: we also cannot set enforced to true on create.  it is disabled

adelowo: gotten the response from slack

golanglemonade (dismissed):

PR #1246, opened 2025-09-16
Switching to an SSO org should trigger authentication

adelowo:


adelowo:
merging since I am putting into another feature branch

PR #1257, opened 2025-09-18
Trust center doc schema

acookin:
- **trust center doc schema**
- **generate**


internal/ent/hooks/trustcenterdoc.go
```diff
@@ -0,0 +1,136 @@
...
+				// delete all relationship tuples on delete, or soft delete (Update Op)
+				err = trustCenterDocDeleteHook(ctx, m)
+			}
+
+			return retValue, err
+		})
+	}
+}
+
+func trustCenterDocCreateHook(ctx context.Context, m *generated.TrustCenterDocMutation) error {
```
golanglemonade: I see that you did this on trust center settings, so maybe I missed it - but you should be able to just use object owned mixin, the which auto creates parent tuples with the `HookObjectOwnedTuples` hook. 

```
newObjectOwnedMixin[generated.TrustCenterDoc](c,
				withParents(TrustCenter{}),
				withOrganizationOwner(true),
),
```

internal/ent/hooks/trustcenterdoc.go
```diff
@@ -0,0 +1,136 @@
...
+			zerolog.Ctx(ctx).Error().Err(err).Msg("failed to create relationship tuple")
+
+			return ErrInternalServerError
+		}
+	}
+
+	return nil
+}
+
+func trustCenterDocDeleteHook(ctx context.Context, m *generated.TrustCenterDocMutation) error {
```
golanglemonade: Same thing here, the object owned mixin takes care of deletion of tuples as well

internal/graphapi/tools_test.go
```diff
@@ -467,4 +467,23 @@ func (suite *GraphTestSuite) orgSubscriptionMocks() {
 		}
 
 	}).Return(nil)
+
+	// setup mocks for subscription schedule creation
```
golanglemonade, line 471: ha sorry, had this on my branch that I just haven't finished yet. :-D thanks!

internal/graphapi/trustcenterdoc_test.go
```diff
@@ -0,0 +1,679 @@
...
+	trustCenter := (&TrustCenterBuilder{client: suite.client}).MustNew(testUser1.UserCtx, t)
+	trustCenterDoc := (&TrustCenterDocBuilder{client: suite.client, TrustCenterID: trustCenter.ID}).MustNew(testUser1.UserCtx, t)
+
+	testCases := []struct {
+		name     string
+		queryID  string
+		client   *testclient.TestClient
+		ctx      context.Context
+		errorMsg string
+	}{
```
golanglemonade, line 29: should probably have a test case for anon trust center viewer

internal/ent/hooks/trustcenterdoc.go
acookin: ah, nice. thanks

internal/graphapi/trustcenterdoc_test.go
acookin: This is intentionally not implemented in this PR, but I'll add a test case to assert the intended behavior here.

PR #1261, opened 2025-09-19
acookin/prep trust center nda template

acookin:
- **prep Templates to support trust center NDA templates**
- **updates**


fga/model/model.fga
```diff
@@ -371,13 +371,13 @@ type procedure
...
-    define can_view: [user, service] or can_edit or (viewer but not blocked)
+    define can_view: [user, service, user:*, service:*] or can_edit or (viewer but not blocked)
     define can_edit: [user, service] or (editor but not blocked)
     define can_delete: [user, service] or (editor but not blocked)
     define audit_log_viewer: ([user, service] or audit_log_viewer from parent) and can_view
-    define parent: [organization]
+    define parent: [organization, user, trust_center]
     # allow a group to be assigned to add edit/view permissions for a set of users
-    define editor: [group#member] or can_edit from parent
+    define editor: [group#member, organization#owner] or can_edit from parent
```
golanglemonade, line 380: You shouldn't need `organization#owner`, since `can_edit from parent` would cover org owners + admins already

fga/model/model.fga
```diff
@@ -371,13 +371,13 @@ type procedure
 # editors will be assigned by groups and can be excluded by users or groups
 type template
   relations
-    define can_view: [user, service] or can_edit or (viewer but not blocked)
+    define can_view: [user, service, user:*, service:*] or can_edit or (viewer but not blocked)
     define can_edit: [user, service] or (editor but not blocked)
     define can_delete: [user, service] or (editor but not blocked)
     define audit_log_viewer: ([user, service] or audit_log_viewer from parent) and can_view
-    define parent: [organization]
+    define parent: [organization, user, trust_center]
```
golanglemonade, line 378: Why do you need `user` as a parent here?

internal/ent/hooks/template.go
```diff
@@ -35,3 +38,110 @@ func HookTemplate() ent.Hook {
...
+			// do the mutation, and then create/delete the relationship
+			retValue, err := next.Mutate(ctx, m)
+			if err != nil {
+				// if we error, do not attempt to create the relationships
+				return retValue, err
+			}
+
+			if m.Op().Is(ent.OpCreate) {
+				createdTemplate := retValue.(*generated.Template)
+				if createdTemplate.TemplateType == enums.RootTemplate && createdTemplate.OwnerID == "" {
```
golanglemonade: you should be able to use `createdTemplate.systemOwned` instead

internal/ent/hooks/template.go
```diff
@@ -35,3 +38,110 @@ func HookTemplate() ent.Hook {
...
+			file.Parent.Type = m.Type()
+
+			ctx = objects.UpdateFileInContextByKey(ctx, key, file)
+		}
+	}
+
+	return ctx, nil
+}
+
+func templateCreateHook(ctx context.Context, m *generated.TemplateMutation) error {
```
golanglemonade, line 110: nothing to change here, but just a heads up: I just merged https://github.com/theopenlane/core/pull/1232 that consolidates a lot of this and so I'll probably pull this out and centralized all this next for the tuple creation. 

Similar for files, Matt has a branch to centralize all that so we don't have these duplicate functions. I"ll let you know when thats merged as well because this pattern will change.

internal/ent/interceptors/filter.go
```diff
@@ -218,6 +219,7 @@ func skipFilter(ctx context.Context, customSkipperFunc ...skipperFunc) bool {
 
 // filterIDList filters a list of object ids to only include the objects that the user has access to
 func filterIDList(ctx context.Context, ids []string, objectType string) ([]string, error) {
+	zerolog.Ctx(ctx).Debug().Msg("filterIDList")
```
golanglemonade: Might want to add more context to these debug logs? e.g something like :

```
zerolog.Ctx(ctx).Str("object", objectType).Strs("ids", ids).Debug().Msg("filterIDList")
```

internal/ent/schema/template.go
```diff
@@ -106,13 +122,27 @@ func (Template) Indexes() []ent.Index {
...
+		// Only one non-deleted NDA per trust center allowed
+		index.Fields("trust_center_id").
+			Unique().Annotations(entsql.IndexWhere(fmt.Sprintf("deleted_at is NULL and kind = '%s'", enums.TemplateKindTrustCenterNda.String()))),
+	}
+}
+
+func (Template) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		entsql.Annotation{
+			Check: fmt.Sprintf("trust_center_id IS NOT NULL OR kind != '%s'", enums.TemplateKindTrustCenterNda.String()),
```
golanglemonade, line 136: The negatives are making my brain hurt. This is just ensuring that if the kind is trust center NDA that there is a trust center id, and vice versa - only allowing a trust center id if the kind is trust center nda? Also TIL about the `Check` annotation haha

internal/graphapi/query/simple/template.graphql
```diff
@@ -1,78 +1,55 @@
...
     templates {
-      createdAt
-      createdBy
       description
       id
       jsonconfig
       name
-      ownerID
-      tags
-      templateType
```
golanglemonade, line 12: any reason to remove all these fields?

fga/model/model.fga
acookin: The `HookObjectOwnedTuples` automatically adds a `user` as an owner to the object, which fails unless this is specified here. I assumed that was intentional, so just made the user change here

golanglemonade: ah right. it gives that user admin permissions over the object automatically. I should probably add a skip for when that isn't really needed, but this works. thanks for confirming!

fga/model/model.fga
acookin: hmm, was getting an error without this:
```
{"level":"error","error":"POST validation error for Write POST with body {\"code\":\"validation_error\",\"message\":\"Invalid tuple 'template:01K5RZ7CS1BCD3HSBTAYCPE7QM#editor@organization:01K5RZ7BXGSJ09JT2QBRPRRZZX#owner'. Reason: 'organization#owner' is not an allowed type restriction for 'template#editor'\"}\n with error code validation_error error message: Invalid tuple 'template:01K5RZ7CS1BCD3HSBTAYCPE7QM#editor@organization:01K5RZ7BXGSJ09JT2QBRPRRZZX#owner'. Reason: 'organization#owner' is not an allowed type restriction for 'template#editor'","user":"organization:01K5RZ7BXGSJ09JT2QBRPRRZZX#owner","relation":"editor","object":"template:01K5RZ7CS1BCD3HSBTAYCPE7QM","time":"2025-09-22T09:03:40-06:00","message":"error creating relationship tuples"}
```

internal/graphapi/query/simple/template.graphql
acookin: nope just was lazy and copied the regular query to the simple query 😬

internal/ent/schema/template.go
acookin: Yeah, basically saying that "if the kind == trust center nda, then it must be associated with a trust center"

fga/model/model.fga
golanglemonade: oh, another one auto created

internal/ent/schema/template.go
```diff
@@ -77,14 +82,20 @@ func (Template) Fields() []ent.Field {
...
 }
 
 // Mixin of the Template
 func (t Template) Mixin() []ent.Mixin {
 	return mixinConfig{
 		additionalMixins: []ent.Mixin{
-			newOrgOwnedMixin(t),
+			newObjectOwnedMixin[generated.Template](t,
+				withParents(TrustCenter{}, Organization{}),
+				withOrganizationOwner(false),
```
golanglemonade: This will be the same I believe because the system admin mixin should auto-set this as well, but I need to confirm that works with the object owned mixin, and not just orgOwnedMixin - can you set this to true for now to be sure and also not confusing to read?
```suggestion
				withOrganizationOwner(true),
```

PR #1263, opened 2025-09-20
Trust Center NDA resolver

acookin:
Trust center NDA graphql resolver.

This creates a resolver that creates a Template object that represents an NDA Template for the trust center

### DIFF
--- a/internal/graphapi/trustcenterndatmpl.go
+++ b/internal/graphapi/trustcenterndatmpl.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"html/template"
 
-	"github.com/99designs/gqlgen/graphql"
 	"github.com/rs/zerolog/log"
 	"github.com/theopenlane/core/internal/ent/generated"
 	gentemplate "github.com/theopenlane/core/internal/ent/generated/template"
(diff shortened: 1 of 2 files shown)


internal/graphapi/schema/trustcenternda.graphql
```diff
@@ -0,0 +1,41 @@
...
+        """
+        templateFiles: [Upload!]
+    ): TrustCenterNDAUpdatePayload!
+}
+
+type TrustCenterNDACreatePayload {
+    template: Template!
+}
+
+input CreateTrustCenterNDAInput {
```
acookin, line 32: We can also add custom titles n stuff here, but I think this is sufficient for now?

golanglemonade: yeah, lets leave it simple for now. We could do some fancy stuff with branding and such as well but getting it working and then making it pretty 💯

internal/graphapi/trustcenterndatmpl.go
```diff
@@ -0,0 +1,173 @@
...
+//go:embed trustcenternda.json.tpl
+var trustCenterNDATemplate string
+
+var (
+	errOneNDAOnly          = errors.New("one NDA file is required")
+	errNDATemplateNotFound = errors.New("NDA template not found")
+	errNDATemplateRequired = errors.New("one NDA template is required")
+)
+
+func createTrustCenterNDA(ctx context.Context, input model.CreateTrustCenterNDAInput, _ []*graphql.Upload) (*model.TrustCenterNDACreatePayload, error) {
```
golanglemonade: why isn't the file input used?

internal/graphapi/trustcenterndatmpl.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, parseRequestError(err, action{action: ActionCreate, object: "trustcenternda"})
+	}
+
+	return &model.TrustCenterNDACreatePayload{
+		Template: updatedTmpl,
+	}, nil
+
+}
+
+func updateTrustCenterNDA(ctx context.Context, id string, _ []*graphql.Upload) (*model.TrustCenterNDAUpdatePayload, error) {
```
golanglemonade: same here, do we need this param in the function?

internal/graphapi/trustcenterndatmpl.go
```diff
@@ -0,0 +1,173 @@
...
+	txnCtx := withTransactionalMutation(ctx)
+	templates, err := txnCtx.Template.Query().Where(gentemplate.TrustCenterIDEQ(id)).All(ctx)
```
golanglemonade: you should be able to use `.Only(ctx)` instead of `All(ctx)` It will ensure there is only 1 returned instead of doing the checks after yourself

> Returns a *NotSingularError when more than one Template entity is found. Returns a *NotFoundError when no Template entities are found.

golanglemonade (approved):

PR #1265, opened 2025-09-22
SSO callbacks always generating sessions for the default org instead of the org being SSO'd

adelowo:


internal/httpserve/handlers/sso.go
```diff
@@ -125,12 +125,18 @@ func (h *Handler) SSOCallbackHandler(ctx echo.Context, openapi *OpenAPIContext)
...
+		return h.BadRequest(ctx, err)
+	}
+
 	// build the OAuth session request
 	oauthReq := apimodels.OauthTokenRequest{
 		Email:            tokens.IDTokenClaims.Email,
 		ExternalUserName: tokens.IDTokenClaims.Name,
 		AuthProvider:     "oidc",
 		Image:            tokens.IDTokenClaims.Picture,
+		OrgID:            orgCookie.Value,
```
golanglemonade, line 139: Did you miss committing a change to add this to the [model](https://github.com/theopenlane/core/blob/05e86a569ca82714306666f68fbb620f3637bb38/pkg/openapi/models.go#L947)?

golanglemonade (approved):

PR #1266, opened 2025-09-22
trust center NDA submissions

acookin:


internal/ent/hooks/objectownedtuples.go
```diff
@@ -38,7 +39,7 @@ func HookObjectOwnedTuples(parents []string, ownerRelation string) ent.Hook {
 			var addTuples []fgax.TupleKey
 
 			// add user permissions to the object on creation
-			if m.Op() == ent.OpCreate {
+			if _, ok := contextx.From[auth.TrustCenterNDAContextKey](ctx); !ok && m.Op() == ent.OpCreate {
```
golanglemonade: nit: this could get messy, I would pull this out into a `skipCreateUserPermssions()` func and this would be the first condition, but I've already thought of a few others where we don't really need this.

internal/ent/schema/mixin_orgowned.go
```diff
@@ -150,6 +150,10 @@ func (o ObjectOwnedMixin) setOwnerIDField(ctx context.Context, m ent.Mutation) e
 		return nil
 	}
 
+	if trustCenterNdaCtx, ok := contextx.From[auth.TrustCenterNDAContextKey](ctx); ok {
```
golanglemonade: Instead of having a conditional, this user will have an auth context, its just the Anon context I believe? Would make sense to update this function instead so it follows the same path and validation: https://github.com/theopenlane/iam/blob/a59ea23d55da21ac467c3a3359a30a623f3e6dec/auth/authuser.go#L63

internal/ent/schema/mixin_orgowned.go
```diff
@@ -164,11 +168,16 @@ func (o ObjectOwnedMixin) setOwnerIDField(ctx context.Context, m ent.Mutation) e
...
 
 // addOrganizationOwnerEditorRelation adds the organization owner as an editor to the object
-func addOrganizationOwnerEditorRelation(ctx context.Context, m ent.Mutation, id string) error {
-	// always add the organization owner relationship as an editor
-	orgID, err := auth.GetOrganizationIDFromContext(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to get organization id from context: %w", err)
+func addOrganizationOwnerEditorRelation(ctx context.Context, m ent.Mutation, id string) (err error) {
+	var orgID string
+	if trustCenterNdaCtx, ok := contextx.From[auth.TrustCenterNDAContextKey](ctx); ok {
```
golanglemonade: updating `GetOrganizationIDFromContext` would also remove this conditional

internal/graphapi/trustcenterndatmpl.go
```diff
@@ -160,3 +171,112 @@ func updateTrustCenterNDA(ctx context.Context, id string) (*model.TrustCenterNDA
...
+
+	if err = validateTrustCenterNDAJSON(trustCenterNDATemplate.Jsonconfig, input.Response, anon); err != nil {
+		return nil, parseRequestError(err, action{action: ActionCreate, object: "trustcenternda"})
+	}
+
+	allowCtx := contextx.With(privacy.DecisionContext(ctx, privacy.Allow), auth.TrustCenterNDAContextKey{
+		OrgID: anon.OrganizationID,
+	})
+
+	previousDocs, err := txnCtx.DocumentData.Query().Where(
```
golanglemonade: All this business logic should be moved to the hooks instead of being in the resolver to follow the patterns elsewhere.

internal/ent/schema/mixin_orgowned.go
```diff
@@ -281,6 +290,10 @@ func (o ObjectOwnedMixin) orgInterceptorSkipper(ctx context.Context, q intercept
 		return true
 	}
 
+	if _, trustCenterNda := contextx.From[auth.TrustCenterNDAContextKey](ctx); trustCenterNda {
```
golanglemonade, line 286: Wouldn't the above conditional cover this since the user would have an anon JWT when signing it?

golanglemonade: Ignore this; I'm confusing this context with the auth user context

go.mod
```diff
@@ -1,17 +1,9 @@
...
 
-tool (
-	github.com/dave/jennifer
-	github.com/invopop/jsonschema
-	github.com/invopop/yaml
-	github.com/vektra/mockery/v3
-	gotest.tools/gotestsum
-)
-
 require (
```
golanglemonade, line 15: uhh whatcha doing? :-D

internal/ent/hooks/documentdata.go
```diff
@@ -0,0 +1,129 @@
...
+	errMissingTemplate                      = errors.New("missing template")
+	errDocInfoDoesNotMatchAuthenticatedUser = errors.New("NDA submission does not match authenticated user")
+	errUserHasAlreadySignedNDA              = errors.New("user has already signed the NDA")
+	errValidationFailed                     = errors.New("validation failed")
+	errMustBeAnonymousUser                  = errors.New("must be an anonymous user")
+	errMissingResponse                      = errors.New("missing response")
+)
+
+// HookDocumentData runs on document data create mutations to set default values that are not provided
+func HookDocumentData() ent.Hook {
```
golanglemonade: I would rename this since its specific to the NDA type of data, not all DocumentData
```suggestion
func HookDocumentDataNDA() ent.Hook {
```

internal/ent/hooks/documentdata.go
```diff
@@ -0,0 +1,129 @@
...
+					},
+				),
+			).Count(ctx)
+			if err != nil {
+				return nil, err
+			}
+
+			if previousDocs > 0 {
+				return nil, errUserHasAlreadySignedNDA
+			}
```
golanglemonade, line 72: I would check this before the validation of the JSON so you fail fast

internal/ent/privacy/policy/checks.go
```diff
@@ -80,6 +80,9 @@ func CanCreateObjectsUnderParents(edges []string) privacy.MutationRuleFunc {
 // CheckOrgReadAccess checks if the requestor has access to read the organization
 func CheckOrgReadAccess() privacy.QueryRule {
 	return privacy.QueryRuleFunc(func(ctx context.Context, q ent.Query) error {
+		if _, hasAnon := auth.AnonymousTrustCenterUserFromContext(ctx); hasAnon {
```
golanglemonade, line 83: ohhhh, just for my understanding, is this because of the default allow on the `CheckCurrentOrgAccess`? 

e.g. 

```
	if err == nil {
		if relation == fgax.CanView {
			// if the relation is view, we can skip the check
			return privacy.Allow
		}

		return checkOrgAccess(ctx, relation, orgID)
	}
```

internal/httpserve/authmanager/authmanager.go
```diff
@@ -97,6 +97,7 @@ func (a *Client) GenerateAnonymousTrustCenterSession(ctx context.Context, w http
 		UserID:        anonUserID,
 		OrgID:         targetOrgID,
 		TrustCenterID: targetTrustCenterID,
+		Email:         "acook@theopenlane.io",
```
golanglemonade: I Assume you are still working on this, but just fyi ^

go.mod
acookin: god dammit, stupid AI. the dummy keeps using a go 1.19 binary instead of the 1.25 that is first in my path lol

internal/ent/hooks/documentdata.go
acookin: I was considering this check to be more expensive than the JSON check, but yeah either way

internal/ent/privacy/policy/checks.go
acookin: yep! and i was scared to mess with that lol

internal/httpserve/authmanager/authmanager.go
acookin: yeah whoops i got lazy

internal/ent/privacy/policy/checks.go
golanglemonade: yeah, its there so we don't have to constantly check FGA on a view check - that a user is already authorized to. Although I'll probably have to update for things like auditor access - but thats for me to figure out later :)

internal/ent/hooks/documentdata.go
```diff
@@ -0,0 +1,130 @@
...
+var (
+	errMissingTemplate                      = errors.New("missing template")
+	errDocInfoDoesNotMatchAuthenticatedUser = errors.New("NDA submission does not match authenticated user")
+	errUserHasAlreadySignedNDA              = errors.New("user has already signed the NDA")
+	errValidationFailed                     = errors.New("validation failed")
+	errMustBeAnonymousUser                  = errors.New("must be an anonymous user")
+	errMissingResponse                      = errors.New("missing response")
+)
+
+// HookDocumentData runs on document data create mutations to set default values that are not provided
```
golanglemonade:
```suggestion
// HookDocumentDataTrustCenterNDA runs on document data create mutations to ensure trust center NDA document submissions are valid
```

PR #1273, opened 2025-09-23
Organization enforced 2fa

matoszz:
In a similar fashion to how we implemented organization enforced SSO, this adds the capability and checks for organization enforced multifactor authentication. Implementing this included:

- Adding the necessary field to organization settings
- Updating the SSO, Login, and WebFinger handlers to include the enforcement status checks / lookups
- Adds tests confirming the various permutations of behavior surrounding multi factor auth org enforcement (user has 2fa set, org not enforced, org enforced + user 2fa not set, org enforced + user 2fa set, etc.)
- Updates the openapi models for SSOStatusReply and LoginReply to include 2fa setting(s)
- Adds some functionality to the CLI during login to perform 2fa checks and receive the code as input in the terminal and performs verification

Additionally, this PR includes some other misc changes:
- Adds organization settings fields in preparation for functionality around auto-joining organizations with matching domains, SAML-based configuration for SSO
- Updates pre-commit hooks to latest major and the implemented hooks to latest minor
- Output of running pre-commit hook on all files


internal/ent/schema/organizationsetting.go
```diff
@@ -119,9 +123,22 @@ func (OrganizationSetting) Fields() []ent.Field {
 		field.String("oidc_discovery_endpoint").
 			Comment("OIDC discovery URL for the SSO provider").
 			Optional(),
+		field.String("saml_signin_url").
+			Comment("the sign in URL to be used for SAML-based authentication").
+			Optional(),
```
golanglemonade, line 129: Should have a URL validator here and on the maybe on the issuer as well? e.g:
```go
Validate(validator.ValidateURL()).
```

internal/ent/schema/organizationsetting.go
```diff
@@ -119,9 +123,22 @@ func (OrganizationSetting) Fields() []ent.Field {
...
+		field.String("saml_issuer").
+			Comment("the SAML issuer").
+			Optional(),
+		field.String("saml_cert").
+			Comment("the x509 certificate used to validate SAML responses").
+			Optional(),
```
golanglemonade, line 135: this should probably be a `field.Text` instead of a string based on the potential size of a cert? Should this be encrypted? `hush.EncryptField(), `?

internal/httpserve/handlers/login.go
```diff
@@ -83,11 +84,21 @@ func (h *Handler) LoginHandler(ctx echo.Context, openapi *OpenAPIContext) error
 		return h.InternalServerError(ctx, err, openapi)
 	}
 
+	// Check if TFA is enforced for the user's organization using already-fetched status
+	tfaRequired := false
+	if orgStatus != nil && orgStatus.OrgTFAEnforced {
+		// Check if user has TFA enabled
+		if user.Edges.Setting == nil || !user.Edges.Setting.IsTfaEnabled {
+			tfaRequired = true
+		}
```
golanglemonade, line 94: why would the orgs `tfaRequired` be based on if the user has it setup or not? This would then return `false` even if it's enforced for the org, but the user never set it up. The `TFAEnabled` already sets if the users has it setup"


```
		TFAEnabled:  user.Edges.Setting.IsTfaEnabled,
```

internal/ent/schema/organizationsetting.go
matoszz: nah its a public cert, but good call on the text field

internal/httpserve/handlers/login.go
matoszz: the flag is to indicate whether to redirect the user to set up TFA, or not. If the user already has TFA setup, and the org enforces it == you don't need to proceed to setup but instead need to proceed to code entry. If the org has TFA enforced, but the user has NOT setup TFA, you would need to direct them to the setup first, then code entry. This is checking that the user's tfa is enabled or not (so !enabled) which means == direct the user to the setup flow.

internal/httpserve/handlers/login_test.go
```diff
@@ -363,3 +370,105 @@ func (suite *HandlerTestSuite) TestLoginHandlerSSOEnforcedOwnerBypass() {
...
+	rec := httptest.NewRecorder()
+	suite.e.ServeHTTP(rec, req)
+
+	// Should succeed without requiring TFA setup since user already has it
+	require.Equal(t, http.StatusOK, rec.Code)
+	var out apimodels.LoginReply
+	require.NoError(t, json.NewDecoder(rec.Body).Decode(&out))
+	assert.True(t, out.Success)
+	assert.True(t, out.TFAEnabled)   // User has TFA enabled
+	assert.False(t, out.TFARequired) // No additional TFA setup needed
```
golanglemonade: oh hmm. based on this test, I guess this field actually is saying "TFA Setup Required" ...not that they should have to go through the TFA flow?

matoszz: Correct (per my comment above)

internal/httpserve/handlers/sso.go
```diff
@@ -305,33 +305,37 @@ func (h *Handler) oidcConfig(ctx context.Context, orgID string) (rp.RelyingParty
 // ssoCallbackURL builds the callback URL for OIDC flows, ensuring a single path segment is appended
 func (h *Handler) ssoCallbackURL() string { return h.OauthProvider.RedirectURL }
 
-// ssoOrgForUser checks if the user's default org requires SSO login and the user is not an owner
-// Returns the org ID and true if SSO is enforced and the user must use SSO, otherwise false
-func (h *Handler) ssoOrgForUser(ctx context.Context, email string) (string, bool) {
+// ssoOrgForUser checks the user's default org SSO and TFA requirements
+// Returns the org settings status which includes both SSO and TFA enforcement
+func (h *Handler) ssoOrgForUser(ctx context.Context, email string) *apimodels.SSOStatusReply {
```
golanglemonade: nit: rename the function since its no longer just sso settings, but org settings `orgRequirementsForUser` or something

PR #1274, opened 2025-09-24
Trust center Document watermark configuration schema

acookin:
Create a schema to hold the trust center doc watermarking config. We will use this when applying document watermarks to the trust center documents.


internal/ent/schema/trustcenterwatermarkconfig.go
```diff
@@ -0,0 +1,169 @@
...
+			Default(defaultWatermarkRotation).
+			Min(-maxWatermarkRotation).
+			Max(maxWatermarkRotation).
+			Optional(),
+		field.String("color").
+			Comment("color of the watermark text").
+			Optional(),
+		field.String("font").
+			Comment("font of the watermark text").
+			Optional(),
```
golanglemonade, line 100: do we need validation here? is the color a hex code? Are there a certain set of fonts that can be set?

internal/ent/schema/trustcenterwatermarkconfig.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+// Mixin of the TrustCenterWatermarkConfig
+func (t TrustCenterWatermarkConfig) Mixin() []ent.Mixin {
+	return mixinConfig{
+		excludeTags: true,
+		additionalMixins: []ent.Mixin{
+			newObjectOwnedMixin[generated.TrustCenterWatermarkConfig](t,
+				withParents(TrustCenter{}),
```
golanglemonade, line 110: I think we want `withOrganizationOwner(false),`  - that allows the interceptors to auto filter on the org ID and pull less data on a list query

golanglemonade (dismissed):

golanglemonade (approved):

PR #1294, opened 2025-09-26
Existing member of an org should not be reinvited

adelowo:


internal/httpserve/handlers/invite_test.go
```diff
@@ -71,18 +72,21 @@ func (suite *HandlerTestSuite) TestOrgInviteAcceptHandler() {
...
 		errMsg   string
 	}{
-		{
-			name:     "happy path",
-			email:    groot,
-			tokenSet: true,
-		},
+		// check for missing token first
+		// else we would have accepted the invite since the
+		// first item in the test table will do that
```
golanglemonade: because we create the invite? We should just  wrap the invite creation in the `tokenSet` as well?

golanglemonade (approved):

PR #1295, opened 2025-09-27
trust center doc watermarking

acookin:


internal/ent/hooks/trustcenterdoc.go
```diff
@@ -50,41 +59,146 @@ func HookTrustCenterDoc() ent.Hook {
...
+			if trustCenterDoc.OriginalFileID == nil {
 				return nil, errMissingFileID
 			}
 
-			if m.Op() == ent.OpUpdateOne && (len(fileIDs) == 0 || mutationSetsFileID) {
-				err = updateTrustCenterDocVisibility(ctx, m, *trustCenterDoc.FileID, trustCenterDoc.ID)
+			tuples := []fgax.TupleKey{}
+
+			if trustCenterDoc.Visibility != enums.TrustCenterDocumentVisibilityNotVisible {
+				tuples = append(tuples, fgax.CreateWildcardViewerTuple(trustCenterDoc.ID, "trust_center_doc")...)
```
golanglemonade, line 79: Why is a wildcard tuple created here when it could potentially be a protected file and not necessarily public?

internal/ent/hooks/trustcenterdoc.go
```diff
@@ -50,41 +59,146 @@ func HookTrustCenterDoc() ent.Hook {
...
+				}, nil); err != nil {
+					return nil, err
+				}
+			} else {
+				zerolog.Ctx(ctx).Debug().Msg("watermarking disabled, setting file id")
+				// Use privacy allow context for internal update operation to bypass authorization checks
+				// and mark as internal operation to avoid triggering the update hook logic
+				allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
+				internalCtx := contextx.With(allowCtx, internalTrustCenterDocUpdateKey{})
+				trustCenterDoc, err = m.Client().TrustCenterDoc.UpdateOne(trustCenterDoc).SetFileID(*trustCenterDoc.OriginalFileID).Save(internalCtx)
```
golanglemonade: Is there a reason we need to do an update after the create instead of setting this before the mutation is done?

internal/ent/hooks/trustcenterdoc.go
acookin: The doc metadata is globally viewable unless it is "not visible". It's the files that are not viewable globally if its protected

internal/ent/hooks/trustcenterdoc.go
acookin: nope lol

PR #1296, opened 2025-09-27
fix: reconcile fixes, delete org sub fix

golanglemonade:
Reconciler:
- adds ability to only run for specific org ids instead of all orgs with diffs
- adds org name to output
- Excludes personal orgs from queries; these are no longer in stripe and so the diffs were showing extra orgs to fix that I wanted to bypass
- Adds additional logic to set the stripe customer id on the org schema if it wasn't already there
- Adds a new "createSubscription" that will create a trial subscription schedule when its missing completely based on the same logic that would happen on an org creation. In theory; this shouldn't be used much but needed it to reconcile a bunch of orgs in weird states
- Removes error on multiple prices, all orgs will will multiple prices now, but its no longer set on the org subscription

Events:
- fixes subscription cancellation when an org is deleted by looking at `Update` events using the `DeleteOrganization` resolver. These return a count of orgs delete in the ent.Value so we are pulling the id from the mutation.ID() for this event type
- Skips the soft delete interceptor when looking up the organization on delete because the org will have already been soft-deleted

Hooks:
- Moved the `createDefaultOrgModulesProductsPrices` to the `reconciler` package. At first I thought I could use the same function for creating completely missing subscriptions and couldn't have the functions in the hooks because of import cycles, however, I needed just the prices. Because this function and the new `GetDefaultPrices` is very similar logic I decided to just leave it in the `reconciler` package. 

Schemas:
- orgSubscription: Removed `product_price`, `features`, and `feature_lookup_keys`; these are no longer used and data was all over the place. We now use modules for features and prices are on each product, not on the subscription

Webhook:
- adds a fallback to looking up on stripe org subscription, that will look up based on our internal org sub id or organization id. This should fix a few remaining records that are in both stripe and our database, but the stripe org subscription id is missing so its not getting updated.

### DIFF
--- a/internal/entitlements/reconciler/reconciler.go
+++ b/internal/entitlements/reconciler/reconciler.go
@@ -71,6 +71,9 @@ var (
 	ErrMissingDBClient       = fmt.Errorf("missing database client")
 	ErrMissingSubscriptionID = fmt.Errorf("missing organization subscription ID")
 	ErrMultiplePrices        = fmt.Errorf("multiple prices found for customer")
+	ErrMissingPrice          = fmt.Errorf("missing price for customer")
+	ErrMultipleCustomers     = fmt.Errorf("multiple customers found for organization")
+	ErrMissingCustomer       = fmt.Errorf("missing customer for organization")
 )
 
 // New creates a new Reconciler instance with the provided options
@@ -229,8 +232,12 @@ func (r *Reconciler) createSubscription(ctx context.Context, cust *entitlements.
 		return err
 	}
 
+	if len(customers) == 0 {
+		return ErrMissingCustomer
+	}
+
 	if len(customers) != 1 {
-		return fmt.Errorf("expected 1 customer for org %s, found %d", cust.OrganizationID, len(customers))
+		return ErrMultipleCustomers
 	}
 
 	cust.Prices, err = GetDefaultPrices(ctx, r.db)
@@ -239,7 +246,7 @@ func (r *Reconciler) createSubscription(ctx context.Context, cust *entitlements.
 	}
 
 	if len(cust.Prices) == 0 {
-		return fmt.Errorf("no prices found for to add for organization %s", cust.OrganizationID)
+		return ErrMissingPrice
 	}
 
 	cust.Status = string(stripe.SubscriptionStatusTrialing)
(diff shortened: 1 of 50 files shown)


internal/entitlements/reconciler/reconciler.go
```diff
@@ -185,6 +223,48 @@ func (r *Reconciler) reconcileOrg(ctx context.Context, org *ent.Organization) er
 	return nil
 }
 
+var trialdays int64 = 30
+
+// createSubscription creates a new subscription in Stripe for the organization customer with the trial settings
+func (r *Reconciler) createSubscription(ctx context.Context, cust *entitlements.OrganizationCustomer) error {
```
matoszz, line 230: I would use CreateCustomerAndSubscription rather than duplicating the logic around searching for customers, getting the prices, etc., and re-defining the trial days. `CreateCustomerAndSubscription` wraps this up (and calls CreateSubscriptionWithPrices)

internal/entitlements/reconciler/reconciler.go
```diff
@@ -268,9 +336,163 @@ func (r *Reconciler) analyzeOrg(ctx context.Context, org *ent.Organization) (str
...
+			return nil, fmt.Errorf("failed to update OrgModule with price ID for module %s: %w", moduleName, err)
+		}
+
+		modulesCreated = append(modulesCreated, moduleName)
+	}
+
+	return modulesCreated, nil
+}
+
+func (r *Reconciler) GetDefaultPrices() ([]entitlements.Price, error) {
```
matoszz: Inside of internal/entitlements there are functions like `TrialMonthlyPriceIDs` that you should be able to import rather than duplicate

internal/entitlements/reconciler/reconciler.go
golanglemonade: I actually started this way; trying to remember the exact issue the things I remember looking at was:
1. this went through hooks, so I needed the stripe client, added that
2. then it also added the tuples, so then I also need the fga client, which then needed access to that database - which kind of pain to get all setup working because I then needed two cloud proxies running

dont remember if this was an import cycle issue too, or just other missing fields, but it ended up being easier to just create the sub and have the webhook fill in the missing data than trying to get all those clients setup and working.

internal/entitlements/reconciler/reconciler.go
golanglemonade: thats is just the price ids returned as a string; I need the full `Price` object. but let me put there there instead so its in the same location

golanglemonade: Moved: https://github.com/theopenlane/core/pull/1296/commits/20a06017afea56a6029289f86194c37919ebaeca

matoszz (approved):

PR #1298, opened 2025-09-28
watermark doc job

acookin:
Job implementation for watermarking trust center documents.

### DIFF
--- a/internal/graphapi/trustcenterdoc_test.go
+++ b/internal/graphapi/trustcenterdoc_test.go
@@ -349,7 +349,7 @@ func TestMutationCreateTrustCenterDoc(t *testing.T) {
 			file:        createPDFUpload(),
 			client:      suite.client.api,
 			ctx:         testUser2.UserCtx,
-			expectedErr: notFoundErrorMsg,
+			expectedErr: notAuthorizedErrorMsg,
 		},
 	}
 
--- a/internal/objects/store/persistence.go
+++ b/internal/objects/store/persistence.go
@@ -14,12 +14,9 @@ import (
 
 	ent "github.com/theopenlane/core/internal/ent/generated"
 	"github.com/theopenlane/core/internal/ent/generated/privacy"
-	"github.com/theopenlane/core/internal/ent/generated/trustcenter"
-	"github.com/theopenlane/core/internal/ent/generated/trustcenterdoc"
 	"github.com/theopenlane/core/pkg/middleware/transaction"
 	pkgobjects "github.com/theopenlane/core/pkg/objects"
 	"github.com/theopenlane/core/pkg/objects/storage"
-	"github.com/theopenlane/gqlgen-plugins/graphutils"
 	"github.com/theopenlane/iam/auth"
 )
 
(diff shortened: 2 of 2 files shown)


internal/middleware/objects/middleware.go
```diff
@@ -188,6 +193,37 @@ func getOrgOwnerID(ctx context.Context, f objects.FileUpload) (string, error) {
...
 
+// getOrgIDFromTrustCenterDoc gets the organization ID for a trust center doc by querying through the trust center
+func getOrgIDFromTrustCenterDoc(ctx context.Context, trustCenterDocID string) (string, error) {
+	var rows sql.Rows
+	query := `
+		SELECT tc.owner_id
+		FROM trust_center_docs tcd
+		JOIN trust_centers tc ON tcd.trust_center_id = tc.id
+		WHERE tcd.id = $1
+	`
```
golanglemonade: Any reason you are writing raw sql here instead of using the ent builders? e.g 

```
				m.Client().TrustCenter.
					Query().
					Where(
						trustcenter.HasTrustCenterDocsWith(trustcenterdoc.ID("trustCenterDocID")),
					).
					Select(trustcenter.FieldOwnerID).
					Only(ctx)
```

pkg/corejobs/watermark_doc.go
```diff
@@ -42,8 +67,245 @@ func (w *WatermarkDocWorker) WithOpenlaneClient(cl olclient.OpenlaneClient) *Wat
...
+	}
+
+	logger.Info().Msg("document watermarking completed successfully")
 	return nil
 }
+
+func watermarkPDFWithText(rs io.ReadSeeker, w io.Writer, config *generated.TrustCenterWatermarkConfig) error {
+	// Create watermark description string
+	// Format: "text:<text>, f:<font>, p:<points>, c:<color>, op:<opacity>, rot:<rotation>, pos:<position>"
+	wmDesc := fmt.Sprintf("fontname:Helvetica, points:%d, fillcolor:%s, op:%.2f, rot:%.0f",
```
golanglemonade: isnt the font also in the config?

acookin: oops! yep. forgot to write the enum -> font name translation thing

internal/middleware/objects/middleware.go
acookin: nope lol

internal/middleware/objects/middleware.go
```diff
@@ -196,32 +200,81 @@ func getOrgOwnerID(ctx context.Context, f objects.FileUpload) (string, error) {
...
+	inputBytes, err := json.Marshal(inputVar)
+	if err != nil {
+		return "", err
 	}
 
-	defer rows.Close()
+	var inputMap map[string]interface{}
+	if err := json.Unmarshal(inputBytes, &inputMap); err != nil {
+		return "", err
+	}
```
golanglemonade: You can use `graphutils.GetMapInputVariableByName(ctx, "input")` instead to simplify this, e.g. something like:

```
			requestInput := graphutils.GetMapInputVariableByName(ctx, "input")
			if requestInput != nil {
				i := *requestInput
                trustCenterID, ok := requestInput["trustCenterID"].(string)
                if ok && ok && trustCenterID != "" {
                    return trustCenterID, nil
                } 
			}
```

PR #1300, opened 2025-09-29
trust center control schema

acookin:
Allow controls to be accessible in the trust center.

Creates a mapping table between trust center and controls to allow for showing the controls in the trust center.

It does look like the control descriptions can be a lil more verbose, so maybe we'll want to add a `trust center description` to the control at some point? or maybe I just used the incorrect entity for this.


golanglemonade (changes requested):
we don't want to link the trust center to individual control objects; let's chat in slack on this one.

PR #1303, opened 2025-09-29
System managed group per org user

adelowo:
- Group should be auto-created and name should be set to Display Name of the user
- If the user update's their Display Name, the group name should also be updated
- If the user is removed from the org, the group should also be deleted

Added a new cli to be used to backfill these groups

### DIFF
--- a/internal/ent/hooks/user.go
+++ b/internal/ent/hooks/user.go
@@ -12,6 +12,7 @@ import (
 	"github.com/theopenlane/core/internal/ent/generated/group"
 	"github.com/theopenlane/core/internal/ent/generated/hook"
 	"github.com/theopenlane/core/internal/ent/generated/organization"
+	"github.com/theopenlane/core/internal/ent/generated/orgmembership"
 	"github.com/theopenlane/core/internal/ent/generated/privacy"
 	"github.com/theopenlane/core/pkg/enums"
 	"github.com/theopenlane/core/pkg/objects"
@@ -414,29 +415,53 @@ func updateSystemManagedGroupForUser(ctx context.Context, m *generated.UserMutat
 
 	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
 
-	groups, err := m.Client().Group.Query().
-		Where(
-			group.CreatedBy(user.ID),
-			group.IsManaged(true),
-			group.Name(getUserGroupName(oldDisplayName, user.ID)),
-		).All(allowCtx)
+	memberships, err := m.Client().OrgMembership.Query().
+		Where(orgmembership.UserID(user.ID)).
+		All(allowCtx)
 	if err != nil {
-		zerolog.Ctx(ctx).Error().Err(err).Msg("error querying user's system managed groups")
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error querying user's org memberships")
 		return err
 	}
 
-	for _, group := range groups {
+	for _, membership := range memberships {
+		newCtx := auth.WithAuthenticatedUser(ctx, &auth.AuthenticatedUser{
+			SubjectID:       user.ID,
+			OrganizationID:  membership.OrganizationID,
+			OrganizationIDs: []string{membership.OrganizationID},
+		})
 
-		err := m.Client().Group.UpdateOneID(group.ID).
+		groups, err := m.Client().Group.Query().
+			Where(
+				group.OwnerID(membership.OrganizationID),
+				group.CreatedBy(user.ID),
+				group.IsManaged(true),
+				group.Name(getUserGroupName(oldDisplayName, user.ID)),
+			).
+			All(privacy.DecisionContext(newCtx, privacy.Allow))
+		if err != nil {
+			zerolog.Ctx(ctx).Error().Err(err).Msg("error querying user's system managed groups for org")
+			return err
+		}
+
+		if len(groups) == 0 {
+			continue
+		}
+		groupIDs := make([]string, 0, len(groups))
+		for _, g := range groups {
+			groupIDs = append(groupIDs, g.ID)
+		}
+
+		err = m.Client().Group.Update().
+			Where(group.IDIn(groupIDs...)).
 			SetName(getUserGroupName(displayName, user.ID)).
 			SetDisplayName(displayName).
 			SetDescription(getUserGroupName(displayName, user.ID)).
-			Exec(allowCtx)
+			Exec(privacy.DecisionContext(newCtx, privacy.Allow))
 
 		if err != nil {
 			zerolog.Ctx(ctx).Error().Err(err).
-				Str("group_id", group.ID).Str("old_display_name", oldDisplayName).Str("new_display_name", displayName).
-				Msg("error updating system managed group name")
+				Str("old_display_name", oldDisplayName).Str("new_display_name", displayName).
+				Msg("error updating system managed group names in bulk")
 			return err
 		}
 	}


internal/httpserve/handlers/invite_test.go
```diff
@@ -149,7 +148,6 @@ func (suite *HandlerTestSuite) TestOrgInviteAcceptHandler() {
 
 			// test the user was added to the group
 			group, err := suite.api.GetGroupByID(recipientCtx, group.ID)
-			require.ErrorContains(t, err, interceptors.ErrFeatureNotEnabled.Error())
```
adelowo, line 152: No idea why this ever worked and build never failed. Group schema has only the base module which every org has

internal/ent/hooks/invite.go
```diff
@@ -527,17 +533,69 @@ func checkAllowedEmailDomain(email string, orgSetting *generated.OrganizationSet
...
-	if !allowed {
-		return ErrEmailDomainNotAllowed
+	// add the user to the group
+	for _, g := range groups {
+
+		var role = enums.RoleMember
+
+		if g.Name == dbUser.DisplayName {
+			role = enums.RoleAdmin
+		}
```
golanglemonade: User does not need to be an admin of their own group; they can't really change anything about the group members.

internal/ent/hooks/invite.go
```diff
@@ -208,6 +209,12 @@ func HookInviteAccepted() ent.Hook {
 				return nil, err
 			}
 
+			if err := createUserManagedGroup(allowCtx, m, userID, ownerID); err != nil {
```
golanglemonade: Can't this be on the orgMembers hook instead so there doesn't need to be a seperate path for invites vs. on org creation where the owner is auto-added?

internal/httpserve/handlers/invite_test.go
golanglemonade: I bet it was something to do with the adding of the authenticated user  on the invite code that changed this path

internal/ent/hooks/orgmembers.go
```diff
@@ -193,3 +222,85 @@ func updateOrgMemberDefaultOrgOnCreate(ctx context.Context, m *generated.OrgMemb
...
+	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
+
+	dbUser, err := m.Client().User.Get(allowCtx, member.UserID)
+	if err != nil {
+		zerolog.Ctx(allowCtx).Error().Err(err).Msg("error fetching user from the database")
+		return err
+	}
+
+	userGroups := map[string]string{
+		dbUser.DisplayName: dbUser.DisplayName,
```
golanglemonade:
```suggestion
		dbUser.ID: dbUser.DisplayName,
```

internal/ent/hooks/orgmembers.go
```diff
@@ -193,3 +222,85 @@ func updateOrgMemberDefaultOrgOnCreate(ctx context.Context, m *generated.OrgMemb
...
+	}
+
+	builders := make([]*generated.GroupCreate, 0, len(userGroups))
+
+	for name, desc := range userGroups {
```
golanglemonade:
```suggestion
	for id, name := range userGroups {
```

internal/ent/hooks/orgmembers.go
```diff
@@ -193,3 +222,85 @@ func updateOrgMemberDefaultOrgOnCreate(ctx context.Context, m *generated.OrgMemb
...
+		tags := []string{"managed", name}
+
+		groupInput := generated.CreateGroupInput{
+			Name:        name,
+			Description: &desc,
```
golanglemonade: chatted in slack but something like this
```
desc := fmt.Sprintf("Managed group for %s", name)
```
```suggestion
			Name:      fmt.Sprintf("%s (%s), id name) ,
			DisplayName: name,
			Description: &desc,
```

internal/ent/hooks/user.go
```diff
@@ -388,3 +394,51 @@ func defaultUserSettings(ctx context.Context, user *generated.UserMutation) (str
...
+		return nil
+	}
+
+	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
+
+	groups, err := m.Client().Group.Query().
+		Where(
+			group.CreatedBy(user.ID),
+			group.IsManaged(true),
+			group.Name(oldDisplayName),
```
golanglemonade: you'll need to update here too to query with the id on name

cmd/managedgroups/main.go
```diff
@@ -0,0 +1,234 @@
...
+				return err
+			}
+
+			fmt.Printf("User: %s (display name: %s)\n", user.ID, user.DisplayName)
+
+			existing, err := db.Group.Query().
+				Where(
+					group.OwnerID(org.ID),
+					group.IsManaged(true),
+					group.Name(user.DisplayName),
```
golanglemonade, line 163: does this need to be updated now to include the user id?

internal/ent/hooks/orgmembers.go
```diff
@@ -193,3 +222,89 @@ func updateOrgMemberDefaultOrgOnCreate(ctx context.Context, m *generated.OrgMemb
...
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error getting user for managed group deletion")
+		return err
+	}
+
+	_, err = m.Client().Group.Delete().
+		Where(
+			group.CreatedBy(userID),
+			group.IsManaged(true),
+			group.OwnerID(orgID),
+			group.Name(getUserGroupName(user.DisplayName, user.ID)),
```
golanglemonade, line 240: can you set the display name to just the user display name?
```suggestion
			group.Name(getUserGroupName(user.DisplayName, user.ID)),
			group.DisplayName(user.DisplayName),
```

internal/ent/hooks/orgmembers.go
```diff
@@ -193,3 +222,89 @@ func updateOrgMemberDefaultOrgOnCreate(ctx context.Context, m *generated.OrgMemb
...
+
+	dbUser, err := m.Client().User.Get(allowCtx, member.UserID)
+	if err != nil {
+		zerolog.Ctx(allowCtx).Error().Err(err).Msg("error fetching user from the database")
+		return err
+	}
+
+	userGroups := map[string]string{
+		dbUser.ID: getUserGroupName(dbUser.DisplayName, dbUser.ID),
+	}
```
golanglemonade: we only have one record; any reason we needed to set this up here that I'm missing?

internal/ent/hooks/orgmembers.go
```diff
@@ -193,3 +222,89 @@ func updateOrgMemberDefaultOrgOnCreate(ctx context.Context, m *generated.OrgMemb
...
+			Description: &desc,
+			Tags:        tags,
+		}
+
+		builders = append(builders, m.Client().Group.Create().
+			SetInput(groupInput).
+			SetIsManaged(true).
+			SetOwnerID(member.OrgID),
+		)
+	}
```
golanglemonade:
```suggestion
      tags := []string{"managed", id}
      
      desc := fmt.Sprintf("Group for %s", dbUser.DisplayName)
      
      groupInput := generated.CreateGroupInput{
	      Name:        getUserGroupName(dbUser.DisplayName, dbUser.ID),
	      DisplayName: dbUser.DisplayName
	      Description: &desc,
	      Tags:        tags,
      }
      
      builders = append(builders, m.Client().Group.Create().
	      SetInput(groupInput).
	      SetIsManaged(true).
	      SetOwnerID(member.OrgID),
      )
```

internal/ent/hooks/orgmembers.go
adelowo: bad copy/paste from moving the logic from invite to orgmembers. will fix

internal/ent/hooks/user.go
```diff
@@ -388,3 +394,51 @@ func defaultUserSettings(ctx context.Context, user *generated.UserMutation) (str
...
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error querying user's system managed groups")
+		return err
+	}
+
+	for _, group := range groups {
+
+		err := m.Client().Group.UpdateOneID(group.ID).
+			SetName(getUserGroupName(displayName, user.ID)).
+			SetDescription(getUserGroupName(displayName, user.ID)).
```
golanglemonade, line 458: Would need to update the description here if we do the one I set on the create spot
```suggestion
            SetDisplayName(displayName).
			SetDescription(getUserGroupName(displayName, user.ID)).
```

internal/graphapi/orgmembers_test.go
```diff
@@ -425,7 +425,7 @@ func TestMutationDeleteOrgMembers(t *testing.T) {
 		Role:           &om.Role,
 	})
 
-	assert.NilError(t, err)
+	assert.ErrorContains(t, err, "orgmembership already exists")
```
golanglemonade, line 428: Why did we go from no error to  expecting an error on this test?

PR #1330, opened 2025-10-03
feat: new impersonation event schema to persist the impersonation request logs

### LINKED ISSUE #1324: [Feature Request] Log impersonation requests to the database
Add an ent schema to persist impersonation events to database for audit trail

1. Create new schema: `ImpersonationEvents`
2. Include the following fields: 
```
		impersonation_type
		user_id + fk (edge) to user table 
		target_user_id + fk (edge) to user table
		action (enum, e.g. START, STOP)
		Reason
		ip_address
		user_agent
		organization_id + edge/fk to organization table
		scopes (`field.Strings`)
	}
```

4. Do not add graphql bits (e.g. use the `entgql.Skip(entgql.SkipAll),` annotation
5. Update handler to log event: https://github.com/theopenlane/core/blob/33f16f9a62fd2035a35ffbdbca5c60cad7886952/internal/httpserve/handlers/impersonation.go#L245

aritradevelops:
Closes #1324 

# Description
Added new `ImpersonationEvent` schema to store all the impersonation requests into the database.
Also update the `logImpersonationEvent` to insert the data into the database.

### DIFF
--- a/internal/ent/schema/impersonationevent.go
+++ b/internal/ent/schema/impersonationevent.go
@@ -4,13 +4,13 @@ import (
 	"entgo.io/contrib/entgql"
 	"entgo.io/ent"
 	"entgo.io/ent/schema"
-	"entgo.io/ent/schema/edge"
 	"entgo.io/ent/schema/field"
 	"github.com/gertd/go-pluralize"
 
 	"github.com/theopenlane/core/pkg/enums"
 	"github.com/theopenlane/core/pkg/models"
 	"github.com/theopenlane/entx"
+	"github.com/theopenlane/entx/history"
 )
 
 // ImpersonationEvent holds the schema definition for the ImpersonationEvent entity
@@ -76,21 +76,22 @@ func (i ImpersonationEvent) Mixin() []ent.Mixin {
 // Edges of the ImpersonationEvent
 func (i ImpersonationEvent) Edges() []ent.Edge {
 	return []ent.Edge{
-		edge.From("user", User.Type).
-			Ref("impersonation_events").
-			Field("user_id").
-			Unique().
-			Required(),
-		edge.From("target_user", User.Type).
-			Ref("targeted_impersonations").
-			Field("target_user_id").
-			Unique().
-			Required(),
-		edge.From("organization", Organization.Type).
-			Ref("organization_impersonation_events").
-			Field("organization_id").
-			Unique().
-			Required(),
+		uniqueEdgeFrom(&edgeDefinition{
+			fromSchema: i,
+			edgeSchema: ImpersonationEvent{},
+			field:      "user_id",
+		}),
+		uniqueEdgeFrom(&edgeDefinition{
+			fromSchema: i,
+			name:       "target_user",
+			t:          ImpersonationEvent.Type,
+			field:      "target_user_id",
+		}),
+		uniqueEdgeFrom(&edgeDefinition{
+			fromSchema: i,
+			edgeSchema: Organization{},
+			field:      "organization_id",
+		}),
 	}
 }
 
@@ -105,6 +106,9 @@ func (ImpersonationEvent) Annotations() []schema.Annotation {
 		entgql.Skip(entgql.SkipAll),
 		entx.SchemaGenSkip(true),
 		entx.QueryGenSkip(true),
+		history.Annotations{
+			Exclude: true,
+		},
 	}
 }
 
--- a/internal/ent/schema/organization.go
+++ b/internal/ent/schema/organization.go
@@ -429,7 +429,7 @@ func (o Organization) Edges() []ent.Edge {
 			edgeSchema:         TrustCenterWatermarkConfig{},
 			cascadeDeleteOwner: true,
 		}),
-		edge.To("organization_impersonation_events", ImpersonationEvent.Type),
+		defaultEdgeToWithPagination(o, ImpersonationEvent{}),
 	}
 }
 
--- a/internal/ent/schema/user.go
+++ b/internal/ent/schema/user.go
@@ -252,6 +252,11 @@ func (u User) Edges() []ent.Edge {
 			name:       "program_owner",
 			t:          Program.Type,
 		}),
+		edgeToWithPagination(&edgeDefinition{
+			fromSchema: u,
+			name:       "targeted_impersonations",
+			t:          ImpersonationEvent.Type,
+		}),
 	}
 }
 


aritradevelops:
Hey @golanglemonade would you please go through the impersonation event schema and check whether I'm on track. Please let me know what other modifications are required in the schema.

also the documentation says `if you do not need the graphql bits, also be certain to add an exclusion to scripts/files_to_skip.txt` but there's no such file so exactly what should i do?

Image

also about the policy I guess this action of `inserting impersonation_event into the database` does not depend on user's permission, so exactly what policy should I use?

internal/ent/schema/impersonationevent.go
```diff
@@ -0,0 +1,132 @@
...
+	return policy.NewPolicy(
+		policy.WithMutationRules(
+			// add mutation rules here, the below is the recommended default
+			policy.AllowCreate(),
+			// this needs to be commented out for the first run that had the entfga annotation
+			// the first run will generate the functions required based on the entfa annotation
+			// entfga.CheckEditAccess[*generated.ImpersonationEventMutation](),
+		),
+	)
+}
```
golanglemonade: Since there are no graphql requests here you can leave this off completely

internal/ent/schema/impersonationevent.go
```diff
@@ -0,0 +1,132 @@
...
+// PluralName returns the plural name of the schema
+func (ImpersonationEvent) PluralName() string {
+	return pluralize.NewClient().Plural(SchemaImpersonationEvent)
+}
+
+// Fields of the ImpersonationEvent
+func (ImpersonationEvent) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("impersonation_type").
+			Comment("Type of impersonation: support, admin, job, etc."),
```
golanglemonade: probably best to make this an enum too. There is a help command  that will generate a enum file for you that you can reference, e.g: 

```
task  cli:generate:enum
task: [cli:generate:enum] go run --tags cli ../../pkg/genenum/cmd/main.go
Enter enum name (e.g. TaskStatus): ImpersonationEventType
Enter enum values (comma-separated, e.g. OPEN,IN_PROGRESS): SUPPORT,ADMIN,JOB
```

internal/ent/schema/impersonationevent.go
```diff
@@ -0,0 +1,132 @@
...
+		field.String("user_agent").
+			Optional().
+			Comment("User-Agent of the impersonator"),
+		field.Strings("scopes").
+			Optional().
+			Comment("Granted scopes during impersonation"),
+		field.String("user_id").
+			Comment("Impersonator user id"),
+		field.String("organization_id").
+			Comment("Impersonator organization id"),
```
golanglemonade:
```suggestion
		field.String("organization_id").
			Comment("id of the organization that is being impersonated"),
```

internal/ent/schema/impersonationevent.go
```diff
@@ -0,0 +1,132 @@
...
+		field.String("target_user_id").
+			Comment("Target user id"),
```
golanglemonade:
```suggestion
			Comment("id of the user that is being impersonated"),
```

golanglemonade:
> Hey @golanglemonade would you please go through the impersonation event schema and check whether I'm on track. Please let me know what other modifications are required in the schema.
> 
> also the documentation says `if you do not need the graphql bits, also be certain to add an exclusion to scripts/files_to_skip.txt` but there's no such file so exactly what should i do?
> 
> Image
> also about the policy I guess this action of `inserting impersonation_event into the database` does not depend on user's permission, so exactly what policy should I use?

@aritradevelops looking good! Made a few comments but you are on the right track. 

That looks like documentation from an old commit? The latest has:

https://github.com/theopenlane/core/blob/7cf170ef8c345c869f7771474ae32cd6e8c9a78f/internal/ent/base/entinit.tmpl#L84-L102

```
        // if you do not need the graphql bits
		// entgql.Skip(entgql.SkipAll),
		// entx.SchemaGenSkip(true),
		// entx.QueryGenSkip(true)
```

Oh, I see that is in our actual docs. Let me fix that!

golanglemonade:
Fixed the docs @aritradevelops : https://docs.theopenlane.io/docs/developers/developing/structure/#ent

aritradevelops:
Hi @golanglemonade ,

I’ve implemented the requested changes. I’ll proceed to create the migration, test everything manually, and then draft the final PR.

I had a quick question — currently, I’ve set the `created_by` field to the ID of the user who is impersonating. However, I was wondering if it would be more appropriate to use the ID of a System User instead, since the data is technically being created by the system rather than the impersonator. Do we have a provision for that?

Thanks!
image

aritradevelops:
> Fixed the docs @aritradevelops : https://docs.theopenlane.io/docs/developers/developing/structure/#ent

I also wanted to mention that in the mixin template, we could add a short comment suggesting the use of
`mixinConfig{excludeAnnotations: true}.getMixins(i)` instead of `getDefaultMixins()` — for cases where someone doesn’t need the GraphQL-related code generation.

By default, `getDefaultMixins()` includes `mixin.GraphQLAnnotationMixin{}`, which generates the GraphQL bits and can sometimes lead to confusion.

Example:

```go
// Mixin of the {{ $name }}
func ({{ $firstChar }} {{ $name }}) Mixin() []ent.Mixin {
    // getDefaultMixins returns the default mixins for all entities.
    // Use mixinConfig{excludeAnnotations: true}.getMixins(i) instead
    // if GraphQL-related annotations are not required.
    // See mixinConfig{} for more configuration options.
    return getDefaultMixins()
}
```

aritradevelops:
Hi @golanglemonade the feature is working as per the requirements. Here's an attached video of the same. Please let me know if furthur modification is required. Thanks!

https://github.com/user-attachments/assets/306e2be0-3981-4b37-adaa-7fff5c30e5da

internal/ent/schema/organization.go
```diff
@@ -429,6 +429,7 @@ func (o Organization) Edges() []ent.Edge {
 			edgeSchema:         TrustCenterWatermarkConfig{},
 			cascadeDeleteOwner: true,
 		}),
+		edge.To("organization_impersonation_events", ImpersonationEvent.Type),
```
golanglemonade: We should be able to use the helpers here, and keep the default name instead of naming it `organization_impersonation_events`. 
```suggestion
                defaultEdgeToWithPagination(o, ImpersonationEvent{})
```

internal/ent/schema/impersonationevent.go
```diff
@@ -0,0 +1,124 @@
...
+			Ref("targeted_impersonations").
+			Field("target_user_id").
+			Unique().
+			Required(),
+		edge.From("organization", Organization.Type).
+			Ref("organization_impersonation_events").
+			Field("organization_id").
+			Unique().
+			Required(),
+	}
```
golanglemonade: Same here, should be able to use the helpers (sorry for the bad indentation)
```suggestion
        uniqueEdgeFrom(&edgeDefinition{
            fromSchema: i,
            edgeSchema: ImpersonationEvent{},
            field: "user_id",
        }
	uniqueEdgeFrom(&edgeDefinition{
            fromSchema: i,
            name: "target_user",
            t: ImpersonationEvent.Type,
            field: "target_user_id",
        },
        uniqueEdgeFrom(&edgeDefinition{
            fromSchema: i,
            edgeSchema: Organization{},
            field: "organization_id",
        }
```

internal/ent/schema/impersonationevent.go
```diff
@@ -0,0 +1,124 @@
...
+
+// Annotations of the ImpersonationEvent
+func (ImpersonationEvent) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		entgql.Skip(entgql.SkipAll),
+		entx.SchemaGenSkip(true),
+		entx.QueryGenSkip(true),
+	}
+}
+
```
golanglemonade, line 114: I don't think we need the history table here, because this is acting as a history table itself. This will affect your db migrations, so I would recommend reverting the `db/*` changes and then rerun `task db:create` so you have a single migration for the PR. 
```suggestion
// Annotations of the ImpersonationEvent
func (ImpersonationEvent) Annotations() []schema.Annotation {
	return []schema.Annotation{
		entgql.Skip(entgql.SkipAll),
		entx.SchemaGenSkip(true),
		entx.QueryGenSkip(true),
		history.Annotations{
			Exclude: true,
		},
	}
}
```

internal/ent/schema/user.go
```diff
@@ -247,6 +247,8 @@ func (u User) Edges() []ent.Edge {
 				entgql.MultiOrder(),
 				accessmap.EdgeNoAuthCheck(), // membership edges are handled by the parent
 			),
+		edge.To("impersonation_events", ImpersonationEvent.Type),
+		edge.To("targeted_impersonations", ImpersonationEvent.Type),
```
golanglemonade: Same here, can use helpers to keep everything in line with the above patterns: 
```suggestion
		defaultEdgeToWithPagination(u, ImpersonationEvent{})
		edgeToWithPagination(&edgeDefinition{
			fromSchema: u,
			name:       "targeted_impersonations",
			t:          ImpersonationEvent.Type,
		}),
```

aritradevelops:
Thanks @golanglemonade  for the discussion and reviews on this PR!
Since this branch has diverged quite a bit and developed multiple merge conflicts, I’ve opened a new PR based on the latest main branch with all the asked changes in #1383 
Closing this one to keep things clean.

Issue #1325, opened 2025-10-02
[Feature Request] Add impersonation metrics


golanglemonade:
The openlane API supports impersonation of users to do things like:

- run jobs on behalf of a user
- temporary access for support
- etc.

It would be nice to have metrics added to track counts of requests:

Add metrics with the type (start/stop) and increment the count

Handler:
https://github.com/theopenlane/core/blob/33f16f9a62fd2035a35ffbdbca5c60cad7886952/internal/httpserve/handlers/impersonation.go#L245

The new metric can be added to the metrics package:
https://github.com/theopenlane/core/blob/main/pkg/metrics/vars.go

With a helper added to the metrics/helpers.go:
https://github.com/theopenlane/core/blob/main/pkg/metrics/helpers.go

AAYUSH-SPIDEY-SHARMA:
hi @golanglemonade i would like to work on this issue, can you please assign this issue to me

golanglemonade:
done @AAYUSH-SPIDEY-SHARMA !

PR #1342, opened 2025-10-05
Added impersonation metrics #1325

myselfsatyam:


internal/ent/validator/ip_address.go
```diff
@@ -0,0 +1,31 @@
...
+		}
+
+		// Reject loopback and unspecified addresses
+		if ip.IsLoopback() || ip.IsUnspecified() {
+			return rout.InvalidField("ip_address")
+		}
+
+		return nil
+	}
+}
```
adelowo, line 31: there is an existing  function that does exactly this . We should probably use that https://github.com/theopenlane/core/blob/main/pkg/models/ipaddress.go#L9-L20

internal/ent/schema/jobrunner.go
```diff
@@ -62,7 +63,8 @@ func (JobRunner) Fields() []ent.Field {
 			Comment("the status of this runner"),
 		field.String("ip_address").
 			Comment("the IP address of this runner").
-			Optional(),
+			Optional().
+			Validate(validator.ValidateIPAddress()),
```
adelowo, line 67: this was removed for a reason before. It used to be validated but I can't recall at this time :)) . Will find out

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "format-graphql": "^1.5.0"
+  }
+}
```
adelowo: we don't need this either. You can install `format-graphql` globally on your machine instead

adelowo (changes requested):
the entire `node_modules` needs to be deleted too and package.json related files

internal/ent/validator/ip_address.go
golanglemonade: This is my fault, I didn't see that and its not used at all today, so we can just delete that one. Keeping it with the other validated functions makes a lot of sense.

internal/ent/schema/jobrunner.go
golanglemonade: Oh? We should be validating it - right now you could put anything you want in there.

golanglemonade:
@myselfsatyam The changes to add metrics + ip address validation are good, but as @adelowo the change to the pre-commit hook and node install stuff needs to be removed

mohit-nagaraj:
@myselfsatyam gentle ping, i am looking forward to your response as i am interested in fixing this issue as well. Thanks

golanglemonade:
> @myselfsatyam gentle ping, i am looking forward to your response as i am interested in fixing this issue as well. Thanks

@mohit-nagaraj this has been stale for a bit. Feel free to make a new PR to fix these issues 🙏

myselfsatyam:
@golanglemonade i am closing this request and making a new one

PR #1349, opened 2025-10-07
added validation function to validate filters of export_type Closes Issue #1345

### LINKED ISSUE #1345: [Feature Request] Validation that `filters` is a valid graphql query
On the `export` schema there is a `filters` string field, that has no validation, this should hit a basic grapqhl validator before the job is inserted. 

1. Add a validator to `internal/ent/validator` [package](https://github.com/theopenlane/core/tree/main/internal/ent/validator)
2. Take the string `filter` and `schema` (`export_type`)  as input and validate, I would recommend looking at the [gqlparser]("github.com/vektah/gqlparser/v2/validator) library as this is the same library gqlgen uses under the hood
3. Ensure unit tests are added for both valid and invalid data

A filter would look something like, this for a task: 

```
{
    "statusIn": [
      "OPEN",
      "IN_PROGRESS",
      "IN_REVIEW"
    ]
}
```

sathwikshetty33:
### Export Filter Validator Closes #1345 

**Function:** `ValidateExportType`

This function validates the `filters` string provided for an export request against the allowed values of a given `export_type`. If the filter is empty, it passes validation. Otherwise, it ensures that the filter conforms to the allowed export type values.

**Unit Test:** `TestValidateExportType`

The unit test verifies the `ValidateExportType` function using two cases:
1. **Valid Input:** A filter string that conforms to one of the allowed export type values passes without error.
2. **Invalid Input:** A filter string that does not match any allowed export type value returns an appropriate error.

The tests ensure both correct validation behavior and proper error handling.

### DIFF
--- a/internal/ent/validator/graphql_filter_validator.go
+++ b/internal/ent/validator/graphql_filter_validator.go
@@ -1,0 +1,26 @@
+package validator
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/theopenlane/core/pkg/enums"
+)
+
+// ValidateExportType checks if the provided export_type is valid.
+func ValidateExportType(filter string, exportType enums.ExportType) error {
+	if filter == "" {
+		return nil
+	}
+
+	// Get all allowed values for the enum
+	validValues := exportType.Values()
+
+	for _, v := range validValues {
+		if strings.ToUpper(filter) == v {
+			return nil // valid
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


golanglemonade:
@sathwikshetty33 going to close this one out since it doesn't address https://github.com/theopenlane/core/issues/1345. Feel free to reopen or create a new PR.  Appreciate the contributions!

PR #1353, opened 2025-10-07
Create trust center by default when module is enabled in stripe

adelowo:
- **improve error message for duplicated trust centers**
- **create trustcenter on webhook and make sure there can only be one at any given time**
- **fix context**

### DIFF
--- a/internal/httpserve/handlers/webhook_helpers.go
+++ b/internal/httpserve/handlers/webhook_helpers.go
@@ -143,15 +143,21 @@ func upsertOrgModule(ctx context.Context, orgSub *ent.OrgSubscription, price *en
 	moduleKey := strings.TrimSpace(productMetadata["module"])
 
 	if moduleKey == models.CatalogTrustCenterModule.String() {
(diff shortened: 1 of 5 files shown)


internal/httpserve/handlers/webhook_helpers.go
```diff
@@ -139,6 +142,27 @@ func upsertOrgModule(ctx context.Context, orgSub *ent.OrgSubscription, price *en
...
+			SubjectID:          orgSub.CreatedBy,
+			OrganizationID:     orgSub.OwnerID,
+			OrganizationIDs:    []string{orgSub.OwnerID},
+			AuthenticationType: auth.JWTAuthentication,
+		})
+
+		// add tx back to the context
+		newCtx = transaction.NewContext(newCtx, tx)
+
+		newCtx = privacy.DecisionContext(newCtx, privacy.Allow)
```
golanglemonade, line 157: @adelowo do you know offhand why you need decision allow when you are setting the user context above? Is it because the feature tuple in't added yet?

adelowo: yes correct

PR #1377, opened 2025-10-10
Add GraphQL Filter Validation for Export Schema

sathwikshetty33:
# PR Summary

This PR adds validation for the **`filters`** field on the export schema to ensure GraphQL filters are well-formed before jobs are inserted into the system. address issue #1345 

---

## Changes

- **Added new validator:** Created `ValidateFilter` function in `internal/ent/validator` package  
- **GraphQL validation:** Utilizes the `gqlparser` library (same as `gqlgen` uses internally) to validate filter syntax and schema compliance  
- **Comprehensive tests:** Added unit tests covering both valid and invalid filter scenarios  

---

## Implementation Details

The validator:

1. Loads all GraphQL schema files from the schema directory  
2. Constructs a test query using the provided filter and export type  
3. Validates the filter against the schema using `gqlparser` and AST validation  
4. Returns detailed error messages for invalid filters

### DIFF
--- a/internal/ent/validator/graphql_filter_validator.go
+++ b/internal/ent/validator/graphql_filter_validator.go
@@ -1,36 +1,43 @@
 package validator
 
 import (
+	"errors"
 	"fmt"
-	"github.com/vektah/gqlparser/v2/parser"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/rs/zerolog/log"
 	"github.com/theopenlane/core/pkg/enums"
 	"github.com/vektah/gqlparser/v2"
 	"github.com/vektah/gqlparser/v2/ast"
+	"github.com/vektah/gqlparser/v2/parser"
 	"github.com/vektah/gqlparser/v2/validator"
 )
 
+var ErrInvalidGraphQLFilter = errors.New("invalid graphql filter provided")
+
 func ValidateFilter(filter string, exportType enums.ExportType) error {
 	if filter == "" {
+		log.Debug().Msg("No filter provided, skipping validation")
 		return nil
 	}
 
-	// Directory containing schema files
 	schemaDir := "../../graphapi/schema"
 
-	// Load all .graphql files from schemaDir
+	// Load all .graphql files in the directory
 	var sources []*ast.Source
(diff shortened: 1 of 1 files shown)


internal/ent/validator/graphql_filter_validator.go
```diff
@@ -0,0 +1,83 @@
...
+			})
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to load schema files: %w", err)
+	}
+
+	if len(sources) == 0 {
+		return fmt.Errorf("no .graphql schema files found in %s", schemaDir)
```
golanglemonade: I would log an error here and just return nil, so it will not always error if something is incorrect with the schemas. I would do this for all the errors above here as well and line 52.

internal/ent/validator/graphql_filter_validator.go
```diff
@@ -0,0 +1,83 @@
...
+		return fmt.Errorf("invalid GraphQL syntax: %w", err)
+	}
+
+	errs := validator.ValidateWithRules(schema, doc, nil)
+	if len(errs) > 0 {
+		var errMsgs []string
+		for _, e := range errs {
+			errMsgs = append(errMsgs, e.Message)
+		}
+		return fmt.Errorf("invalid filter for %s: %s", exportType, strings.Join(errMsgs, "; "))
```
golanglemonade: this needs to be a defined error, exported so usage of this validator would be able to do checks on `errors.Is()`

e.g. 

```
var ErrInvalidGraphQLFilter = errors.New("invalid graphql filter provided")
```

And then you would update this line to:

```
return fmt.Errorf("%w for %s: %s", ErrInvalidGraphQLFilter, exportType, strings.Join(errMsgs, "; "))
```

golanglemonade:
thanks @sathwikshetty33! Just a few comments on errors - which should also fix the linter failure

sathwikshetty33:
@golanglemonade I have made all the changes you have mentioned, please do check and do let me know if any further changes are required.

golanglemonade:
thanks @sathwikshetty33!

sathwikshetty33:
@golanglemonade, Thank you so much!

PR #1379, opened 2025-10-10
Object Storage refactor - Storage Provider Implementation

matoszz:
This PR is the first of several which will be created to accomplish a refactor of the object storage package with a focus on enabling:
- multiple storage providers to be used concurrently
- storage providers can be initialized from credentials that come from several sources (our config, database, etc.)
- initialized storage clients can be pooled without package-specific wiring
- configuration of storage providers moves from ad-hoc to standardized

To accomplish this, the package was refactored and several functional components were re-homed. At a high level:
- `objects` – request/context helpers, validation, ent mutation adapters
- `objects/storage` – high-level service plus convenience aliases
- `objects/storage/types` – canonical provider interfaces and DTOs
- `objects/storage/providers/*` – concrete provider implementations

This PR contains primarily the last 2 package updates mentioned above - the new provider(s) structure and types. The providers were structured this way to:
- standardize every backend on a shared interface contract (in this case the `Provider` interface)
- funnel provider construction through builder helpers
- make dynamic provider resolution and credential-driven instantiation possible by having each provider accept `ProviderOptions` and `ProviderCredentials`
- move / add cross-cutting behaviors like metrics, probes, presign defaults, etc., into individual providers (rather than be shared)

The additional items outside of the individual providers that were added in this branch are mostly helpers and a precursor to a full in-memory file buffering solution which is intended to prevent the frequency at which we're writing files to disk (today we write every file to disk before re-reading it back into memory to stream it to the destination storage location).

### DIFF
--- a/pkg/objects/storage/providers/database/builder.go
+++ b/pkg/objects/storage/providers/database/builder.go
@@ -49,7 +49,7 @@ func (b *Builder) WithOptions(opts ...Option) *Builder {
 }
 
 // Build implements eddy.Builder.
-func (b *Builder) Build(credentials storage.ProviderCredentials, config *storage.ProviderOptions) (storagetypes.Provider, error) {
+func (b *Builder) Build(config *storage.ProviderOptions) (storagetypes.Provider, error) {
 	if config == nil {
 		config = storage.NewProviderOptions()
 	}


pkg/objects/storage/providers/database/provider.go
```diff
@@ -0,0 +1,286 @@
...
+	"io"
+	"net/url"
+	"strings"
+	"time"
+
+	jwt "github.com/golang-jwt/jwt/v5"
+
+	ent "github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/file"
+	"github.com/theopenlane/core/internal/ent/generated/privacy"
```
golanglemonade, line 15: This is in `pkg` but relies on an `internal` package, which is not ideal because it is no longer able to be imported elsewhere which is usually the reason to use `pkg` over `internal`, but if its not easily able to use an interface instead I would just leave it today and we can consider a better way later. Just wanted to call it out.

pkg/objects/storage/providers/r2/provider.go
```diff
@@ -0,0 +1,260 @@
...
+
+// GetPresignedURL implements storagetypes.Provider
+func (p *Provider) GetPresignedURL(_ context.Context, file *storagetypes.File, opts *storagetypes.PresignedURLOptions) (string, error) {
+	expires := opts.Duration
+	if expires == 0 {
+		expires = DefaultPresignedURLExpiry
+	}
+
+	presignURL, err := p.presignClient.PresignGetObject(context.Background(), &s3.GetObjectInput{
+		Bucket:                     aws.String(p.options.Bucket),
```
golanglemonade:
```suggestion
func (p *Provider) GetPresignedURL(ctx context.Context, file *storagetypes.File, opts *storagetypes.PresignedURLOptions) (string, error) {
	expires := opts.Duration
	if expires == 0 {
		expires = DefaultPresignedURLExpiry
	}

	presignURL, err := p.presignClient.PresignGetObject(ctx, &s3.GetObjectInput{
		Bucket:                     aws.String(p.options.Bucket),
```

pkg/objects/storage/providers/disk/builder.go
```diff
@@ -0,0 +1,62 @@
...
+// Builder creates disk providers for the client pool
+type Builder struct{}
+
+// NewDiskBuilder creates a new Builder
+func NewDiskBuilder() *Builder {
+	return &Builder{}
+}
+
+// Build implements eddy.Builder
+func (b *Builder) Build(_ context.Context, credentials storage.ProviderCredentials, config *storage.ProviderOptions) (storagetypes.Provider, error) {
```
golanglemonade: Do any of the `Build` functions use context? If not we should just remove as a parameter unless you think it would be needed in the future for any reason?

pkg/objects/storage/providers/s3/provider.go
```diff
@@ -0,0 +1,430 @@
...
+	return nil
+}
+
+// GetPresignedURL implements storagetypes.Provider
+func (p *Provider) GetPresignedURL(_ context.Context, file *storagetypes.File, opts *storagetypes.PresignedURLOptions) (string, error) {
+	if opts.Duration == 0 {
+		opts.Duration = DefaultPresignedURLExpiry
+	}
+
+	presignURL, err := p.presignClient.PresignGetObject(context.Background(), &s3.GetObjectInput{
```
golanglemonade:
```suggestion
// GetPresignedURL implements storagetypes.Provider
func (p *Provider) GetPresignedURL(ctx context.Context, file *storagetypes.File, opts *storagetypes.PresignedURLOptions) (string, error) {
	if opts.Duration == 0 {
		opts.Duration = DefaultPresignedURLExpiry
	}

	presignURL, err := p.presignClient.PresignGetObject(ctx, &s3.GetObjectInput{
```

pkg/objects/storage/types/provider.go
```diff
@@ -0,0 +1,180 @@
...
+	Delete(ctx context.Context, file *File, opts *DeleteFileOptions) error
+	// GetPresignedURL generates a presigned URL for file access
+	GetPresignedURL(ctx context.Context, file *File, opts *PresignedURLOptions) (string, error)
+	// Exists checks if a file exists in the storage provider
+	Exists(ctx context.Context, file *File) (bool, error)
+	// GetScheme returns the URL scheme for this provider
+	GetScheme() *string
+	// ListBuckets is used to list the buckets in the storage backend
+	ListBuckets() ([]string, error)
+	ProviderType() ProviderType
```
golanglemonade, line 29: missing comment on `ProviderType`

pkg/objects/storage/types/provider.go
```diff
@@ -0,0 +1,180 @@
...
+	UpdatedAt time.Time `json:"updated_at"`
+	// Parent is the parent object of the file, if any
+	Parent ParentObject `json:"parent"`
+	// Metadata contains additional file metadata
+	Metadata map[string]string `json:"metadata,omitempty"`
+	// CorrelatedObjectID is the ID of the object this file belongs to
+	CorrelatedObjectID string
+	// CorrelatedObjectType is the type of object this file belongs to
+	CorrelatedObjectType string
+	ProviderType         ProviderType `json:"provider_type,omitempty"`
```
golanglemonade: missing comment on `ProviderType`

pkg/objects/storage/types/provider.go
```diff
@@ -0,0 +1,180 @@
...
+	// Metadata contains additional hints for provider selection
+	Metadata map[string]string `json:"metadata,omitempty"`
+}
+
+type PresignedURLOptions struct {
+	// Duration is the duration the presigned URL should be valid for
+	Duration time.Duration `json:"duration"`
+}
+
+type DeleteFileOptions struct {
```
golanglemonade, line 190: comment exported structs 🙏

PR #1394, opened 2025-10-14
Object Storage Refactor + Client Pooling Implementation

matoszz:
This is a relatively large refactor (apologies in advance!) but whose intention is to accomplish, generally speaking, the ability to run concurrent storage providers (rather than a single one configured globally) based on our own configuration, as well as concurrent to that provide the ability to instantiate new storage clients should we choose to allow our tenants to bring their own storage credentials / wish us to store their content in their storage provider.

This PR relies heavily on a new package introduced [here](https://github.com/theopenlane/eddy). I elected to have this package be external given the generics / nature of what it's doing as well as trying to create the ability for broader consumption outside the `core` repo. 

## Overview

- Introduced a dedicated object-storage orchestration service that fronts provider resolution, validation, pre-sign tokens, and existence checks via eddy client pooling
  - `internal/objects/service.go`
- Added a configurable resolver stack with provider hints, module-aware routing, and dev-mode overrides so storage backends can be swapped per tenant or feature at runtime
  - internal/objects/hints.go
  - internal/objects/resolver/rules.go
  - internal/objects/resolver/factory.go
- Replaced the legacy middleware/handler flow with a cohesive upload pipeline that normalizes files in context, enriches provider hints, buffers streams, and assembles `UploadOptions` for downstream persistence
  - `internal/graphapi/helpers.go`
  - `internal/objects/upload/handler.go`
- Rebuilt the public `pkg/objects` surface: streamlined upload parsing/helpers, refreshed documentation, and retired the old middleware adapters in favor of a slimmer API with new buffering, MIME detection, and validator utilities 
- Stood up a full persistence layer for ent-backed metadata updates, including provider-specific storage details, org ownership derivation, and resolver integration
  - `internal/objects/store/persistence.go`
  - `internal/objects/store/permissions.go`
- Updated configuration, CLI commands, and HTTP bootstrapping to load the new objectStorage provider map and to source uploads through the revamped service

## Behavioral / Safety changes

- Each individual storage provider has a configuration parameter that can be enabled `ensureAvailable` which not only resolves the credentials provided for the provider, but lists the buckets and ensures a writeable location exists. If enabled, this would block startup of the service if the credentials were invalid for the configured provider
- Added readiness checks which re-wrap the same logic mentioned above but for the running server instance to ensure the storage provider backend remains accessible. The `eddy` integration allows for rule definitions that enable "fallback" logic in case you wanted to not block on the availability of the provider but instead use a secondary provider, set of credentials, etc.
- A `waitGroup` is incremented when performing uploads of files and a call to `WaitForUploads` is issued before allowing the server to fully shut down, so if there is an upload in progress the server would wait for the upload to finish (or for the context to time out)
- When `devMode` is selected, the `ValidateAvailabilityByProvider` is bypassed; `handleDevMode` in `internal/objects/resolver/rules.go` installs a single unconditional rule that always resolves to the disk builder using the default dev bucket (`objects.DefaultDevStorageBucket`). No other provider rules run, however, the goal would be to use this setting infrequently and instead take advantage of the ability to configure the s3 provider with a local minio instance so that full s3 compatibility is frequently tested locally

## Staged, but not included

- CLI code which wholesale uses the packages exposed as well as the openlane clients to perform operations similar to how an external customer / tenant might (omitted from this PR for sake of brevity)
- Pre-sign endpoint validation which masks our bucket configuration with the individual providers and uses `TokenManager` to perform JWT signing + URL encoding capable of buffering the content stream from our bucket directly into the pre-sign URL response (avoiding exposing our bucket name, provider, etc., entirely)
- `CredentialSync` service which on server startup, parses our configuration files and creates `integration` and `hush` records; the intent of the service was to make how we fetch + init storage clients the same regardless of whether or not they were ours or provided by our tenants, but for simplicity this was omitted from this PR
- Docker compose with minio configuration, credential / user creation, load + benchmark testing

### DIFF
--- a/internal/graphapi/search_test.go
+++ b/internal/graphapi/search_test.go
@@ -19,20 +19,22 @@ func TestGlobalSearch(t *testing.T) {
 	testViewOnlyUser := suite.userBuilder(context.Background(), t)
 	suite.addUserToOrganization(testSearchUser.UserCtx, t, &testViewOnlyUser, enums.RoleMember, testSearchUser.OrganizationID)
 
+	testAnotherUser := suite.userBuilder(context.Background(), t)
+
 	// create multiple objects to be searched by testSearchUser
 	// dont use objects that might be created by the system, that could be returned such as system owned controls, policies, etc
 	// also don't use objects that are created automatically, like groups
 	numContacts := 10
 	contactIDs := []string{}
 	for i := range numContacts {
-		contact := (&ContactBuilder{client: suite.client, Name: fmt.Sprintf("Test Contact %d", i)}).MustNew(testSearchUser.UserCtx, t)
+		contact := (&ContactBuilder{client: suite.client, Name: fmt.Sprintf("Test A1CD2D Contact %d", i)}).MustNew(testSearchUser.UserCtx, t)
 		contactIDs = append(contactIDs, contact.ID)
 	}
 
 	numPrograms := 3
 	programIDs := []string{}
 	for i := range numPrograms {
-		program := (&ProgramBuilder{client: suite.client, Name: fmt.Sprintf("Test Program %d", i)}).MustNew(testSearchUser.UserCtx, t)
+		program := (&ProgramBuilder{client: suite.client, Name: fmt.Sprintf("Test A1CD2D Program %d", i)}).MustNew(testSearchUser.UserCtx, t)
 		programIDs = append(programIDs, program.ID)
 	}
 
@@ -50,7 +52,7 @@ func TestGlobalSearch(t *testing.T) {
 			name:             "happy path",
 			client:           suite.client.api,
 			ctx:              testSearchUser.UserCtx,
-			query:            "Test",
+			query:            "A1CD2D",
 			expectedResults:  13, // this is total count of all objects searched
 			expectedContacts: 10,
 			expectedPrograms: 3,
@@ -59,7 +61,7 @@ func TestGlobalSearch(t *testing.T) {
 			name:             "happy path, case insensitive",
 			client:           suite.client.api,
 			ctx:              testSearchUser.UserCtx,
-			query:            "TEST",
+			query:            "a1cd2d",
 			expectedResults:  13, // this is total count of all objects searched
 			expectedContacts: 10,
 			expectedPrograms: 3,
@@ -77,7 +79,7 @@ func TestGlobalSearch(t *testing.T) {
 			name:             "happy path, view only user",
 			client:           suite.client.api,
 			ctx:              testViewOnlyUser.UserCtx,
-			query:            "Test",
+			query:            "A1CD2D",
 			expectedResults:  10, // this is total count of all objects searched
 			expectedContacts: 10,
 			expectedPrograms: 0, // no access to the programs by the view only user
@@ -92,8 +94,8 @@ func TestGlobalSearch(t *testing.T) {
 		{
 			name:            "no results, another user",
 			client:          suite.client.api,
-			ctx:             testUser2.UserCtx,
-			query:           "Test",
+			ctx:             testAnotherUser.UserCtx,
+			query:           "A1CD2D",
 			expectedResults: 0,
 		},
 		{


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/339

### Changes Preview:

config/config.example.yaml
```diff
@@ -187,20 +187,80 @@ keywatcher:
     keyDir: ./keys
     secretManager: ""
 objectStorage:
```
golanglemonade, line 189: The dev example should be updated as well: https://github.com/theopenlane/core/blob/main/config/config-dev.example.yaml#L148

internal/ent/hooks/file.go
```diff
@@ -48,17 +52,70 @@ func HookFileDelete() ent.Hook {
...
+						file.FieldID,
+						file.FieldStoragePath,
+						file.FieldStorageProvider,
+						file.FieldDetectedContentType,
+						file.FieldPersistedFileSize,
+						file.FieldMetadata,
+						file.FieldStorageVolume,
+					).
+					WithIntegrations().
+					WithSecrets().
```
golanglemonade: Do we have test cases that cover this from a non-admin users (e.g. member/view only users)? I'm worried that a user that can create files, is going to hit a permission denied error when trying to delete because they don't have access to secrets. 

Also, the `Select` doesn't include any fields from integrations/or secrets? I'm not confident I know what happens here if you use `With` but also don't have the fields from the join tables? Feels like you wouldn't actually get those fields back?

internal/ent/hooks/file.go
```diff
@@ -48,17 +52,70 @@ func HookFileDelete() ent.Hook {
...
+						// These ID's arent yet used but we should set them now so that when we do use them, they are available
+						// Set integration ID from edges
+						for _, integration := range f.Edges.Integrations {
+							storageFile.ProviderHints.IntegrationID = integration.ID
+						}
+
+						// Set hush ID from edges
+						for _, secret := range f.Edges.Secrets {
+							storageFile.ProviderHints.HushID = secret.ID
+						}
```
golanglemonade: What happens if there is more than one? we just get the last one set?

internal/ent/hooks/file.go
```diff
@@ -35,6 +38,7 @@ func HookFileDelete() ent.Hook {
 					ids = append(ids, dbIDs...)
 
 				case ent.OpDeleteOne:
```
golanglemonade: You didn't change these, but these probably need 

```
	case ent.OpDelete, ent.OpUpdate:
...
    case ent.OpDeleteOne, ent.OpUpdateOne:
```

to cover soft deletes

internal/ent/hooks/file.go
```diff
@@ -48,17 +52,70 @@ func HookFileDelete() ent.Hook {
...
+						}
+
+						// Set provider-specific fields if available
+						if f.StorageProvider != "" {
+							storageFile.ProviderType = storagetypes.ProviderType(f.StorageProvider)
+						}
+
+						if err := m.ObjectManager.Delete(ctx, storageFile, nil); err != nil {
+							log.Error().Err(err).Str("file_id", f.ID).Msg("failed to delete file from storage")
+							// Continue with other files rather than failing the entire operation
```
golanglemonade: Any reason not to fail? This could leave dangling files in storage

internal/ent/hooks/procedure_test.go
```diff
@@ -1,42 +0,0 @@
-package hooks
-
-import (
-	"testing"
-)
-
-func TestDetectMimeTypeFromContent(t *testing.T) {
```
golanglemonade, line 7: This test should still exist, just the file renamed to document_test.go. The function it's testing, `detectMimeTypeFromContent` still exists.

internal/ent/hooks/trustcenterdoc.go
```diff
@@ -174,6 +208,21 @@ func HookUpdateTrustCenterDoc() ent.Hook {
 				}
 			}
 
+			// Ensure parent tuple exists linking the document to the current file for protected visibility flows
```
golanglemonade: This looks like new code - but I'm not sure why - did you find a bug here? Isn't this handled by `AddFilePermissions`?

internal/httpserve/server/server.go
```diff
@@ -158,9 +134,9 @@ func (s *Server) StartEchoServer(ctx context.Context) error {
 
 	// Set the local file path if the object storage provider is disk
 	// this allows us to serve up the files during testing
-	if s.config.Settings.ObjectStorage.Provider == storage.ProviderDisk {
-		s.Router.LocalFilePath = s.config.Settings.ObjectStorage.DefaultBucket
-	}
+	//	if s.config.Settings.ObjectStorage.Provider == storagetypes.DiskProvider {
+	//		s.Router.LocalFilePath = s.config.Settings.ObjectStorage.DefaultBucket
+	//	}
```
golanglemonade: Should this be removed instead of commented out?

internal/objects/resolver/rules.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// handleDevMode adds a dev-only disk rule when configured and returns true if handled
+func (rc *ruleCoordinator) handleDevMode() bool {
+	if !rc.config.DevMode {
+		return false
+	}
+
+	options := storage.NewProviderOptions(
+		storage.WithBucket(objects.DefaultDevStorageBucket),
```
golanglemonade: Does this mean it won't use the config setting? The `DefaultDevStorageBucket` is `/tmp` which doesn't exist windows. We should be sure to respect the `bucket` setting if its in the config, rather than using the default. This already happens here: 

https://github.com/theopenlane/core/pull/1394/files?diff=unified&w=1#diff-7b60098fd097559342f7f357bab5bf5367118df3eae0f0ba4a77eba270a0b948R93-R97

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+	isAdmin := false
+	if au, err := auth.GetAuthenticatedUserFromContext(ctx); err == nil && au.IsSystemAdmin {
+		isAdmin = true
+	}
+
+	orgID, _ := auth.GetOrganizationIDFromContext(ctx)
+	if isAdmin {
+		orgID = ""
+	}
+
```
golanglemonade, line 135: nit: you can simplify this - org ID is in the authenticated user context already and also return early: 
```suggestion
	// If the actor is a system admin, prefer deriving the organization from the
	// correlated object rather than using the admin's org from context
	au, err := auth.GetAuthenticatedUserFromContext(ctx)
	if err != nil {
	    return "", err
	}
	    
    if !au.IsSystemAdmin && au.OrganizationID != "" {
        return au.OrganizationID
    }
```

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+		// derive table name from correlated object type using snake_case to match DB naming
+		objectType := lo.SnakeCase(f.CorrelatedObjectType)
+		objectTable := pluralize.NewClient().Plural(objectType)
+
+		// For schemas that do not include owner_id (e.g., trust_center_docs), resolve owner through the parent
+		if objectType == "trust_center_doc" {
+			// first get the trust_center_id from the doc
+			var tcRows sql.Rows
+
+			tcQuery := "SELECT trust_center_id FROM trust_center_docs WHERE id = $1"
```
golanglemonade: This should be able to be written with one query, using a join, instead of two:

```
SELECT tc.owner_id FROM trust_center as tc JOIN trust_center_docs as tcd ON tc.id = tcd. trust_center_id WHERE tcd.id = $1
```

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+	if au, err := auth.GetAuthenticatedUserFromContext(ctx); err == nil && au.IsSystemAdmin {
+		isAdmin = true
+	}
+
+	orgID, _ := auth.GetOrganizationIDFromContext(ctx)
+	if isAdmin {
+		orgID = ""
+	}
+
+	if orgID == "" {
```
golanglemonade: With the above suggestion, you can also drop this if `if orgID == "" {`

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+			if trustCenterID == "" {
+				return "", ErrMissingOrganizationID
+			}
+
+			// now resolve owner_id from trust_centers
+			var ownerRows sql.Rows
+			ownerQuery := "SELECT owner_id FROM trust_centers WHERE id = $1"
+			if err := txClientFromContext(ctx).Driver().Query(ctx, ownerQuery, []any{trustCenterID}, &ownerRows); err != nil {
+				return "", err
+			}
```
golanglemonade: Instead of doing the multiple `ownerRows.Close()`, just do a `defer ownerRows.Close()` so you don't have to worry about it

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+			if rows.Err() != nil {
+				return "", rows.Err()
+			}
+			defer rows.Close()
+			if rows.Next() {
+				var ownerID string
+				if err := rows.Scan(&ownerID); err != nil {
+					return "", err
+				}
+				orgID = ownerID
```
golanglemonade: same here
```suggestion
				return ownerID, nil
```

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+				_ = ownerRows.Close()
+				return "", ownerRows.Err()
+			}
+			if ownerRows.Next() {
+				var ownerID string
+				if err := ownerRows.Scan(&ownerID); err != nil {
+					_ = ownerRows.Close()
+					return "", err
+				}
+				orgID = ownerID
```
golanglemonade: We can just return if we have it, nothing happens below once we have the org ID
```suggestion
				return ownerID, nil
```

internal/objects/store/persistence.go
```diff
@@ -0,0 +1,224 @@
...
+			}
+		}
+	}
+
+	if orgID == "" {
+		return "", ErrMissingOrganizationID
+	}
+
+	return orgID, nil
```
golanglemonade: Now if we get here, we can return the  error because org id wasn't found:

```suggestion
	return "", ErrMissingOrganizationID
```

internal/objects/upload/handler.go
```diff
@@ -0,0 +1,163 @@
...
+		pkgobjects.AddUpload()
+		metrics.StartFileUpload()
+		startTime := time.Now()
+
+		finish := func(status string) {
+			metrics.FinishFileUpload(status, time.Since(startTime).Seconds())
+			pkgobjects.DoneUpload()
+		}
+
+		orgID, _ := auth.GetOrganizationIDFromContext(ctx)
```
golanglemonade, line 40: We shouldn't ignore the error here 
```suggestion
		orgID, err := auth.GetOrganizationIDFromContext(ctx)
		if err != nil {
		     return ctx, nil, err
		} 
```

internal/objects/upload/handler.go
```diff
@@ -0,0 +1,163 @@
...
+					contentType = detected
+				}
+				// Replace the original reader so the upload pipeline can still stream the contents.
+				f.RawFile = buffered
+			}
+		}
+
+		if contentType != "" {
+			f.ContentType = contentType
+		}
```
golanglemonade, line 142: Why do we need this conditional to only set it when its not empty?

internal/objects/validators/mime.go
```diff
@@ -39,18 +36,18 @@ var validMimeTypes = map[string][]string{
 	"noteFiles":          sharedMimeTypes,
 	"exportFiles":        {"text/csv", "text/plain; charset=utf-8", "text/plain", "application/json", "application/json; charset=utf-8"},
 	"procedureFile":      importSchemaMimeTypes,
-	"internalPolicyFile": importSchemaMimeTypes,
-	"actionPlanFile":     importSchemaMimeTypes,
+	"policyFile":         importSchemaMimeTypes,
```
golanglemonade: Why was internalPolicy name changed? And actionPlanFile dropped?

pkg/objects/adapter.go
```diff
@@ -0,0 +1,38 @@
+package objects
+
+// Mutation represents any ent mutation that can provide ID and Type
+type Mutation interface {
+	ID() (string, error)
+	Type() string
+}
```
golanglemonade, line 7: Do you need `IDs(ctx context.Context) ([]string, error)` here as well? `ID()` only works on single mutation, e.g. `UpdateOne`, `DeleteOne` and not `Update` or `Delete`

pkg/objects/storage/utils.go
```diff
@@ -62,12 +64,43 @@ func ParseDocument(reader io.Reader, mimeType string) (any, error) {
...
+		}
 
+		return text, nil
 	default:
 		return data, nil
 	}
 }
 
+// parseDocx extracts and returns the text content from a DOCX file
+func parseDocx(content []byte) (string, error) {
```
golanglemonade, line 80: Can we export this function and use it in `internal/ent/hooks/documents.go` instead of having two functions doing the same thing?

Just a note: this is being updated in another PR to not use `unioffice` but want to make sure we aren't writing this multiple times.

golanglemonade: Or is this like this because of this comment in hooks? 

```
// TODO: Reconcile with pkg/objects/storage/utils.go - I'm leaving this as-is for now for sanity
```

pkg/objects/upload.go
```diff
@@ -1,34 +1,332 @@
...
+				continue
+			}
+
+			files = append(files, File{
+				RawFile:      file,
+				OriginalName: header.Filename,
+				FieldName:    key,
+				FileMetadata: FileMetadata{
+					Size:        header.Size,
+					ContentType: header.Header.Get("Content-Type"),
```
golanglemonade: I dont think this will ever be the content type of the actual file when in a graphl request?

pkg/objects/upload.go
```diff
@@ -1,34 +1,332 @@
...
 
-	pr.BytesRead += int64(n)
+		var files []File
+		for _, header := range fileHeaders {
+			file, err := header.Open()
+			if err != nil {
+				log.Error().Err(err).Str("file", header.Filename).Msg("failed to open file")
 
-	pr.Print()
+				continue
```
golanglemonade: What happens if this gets hit? For a user does that mean a the request could be succesful even if the file can't be read? Should the error be returned?

internal/ent/hooks/file.go
matoszz: A failure would still leave dangling files in this case, though, right ?

internal/ent/hooks/trustcenterdoc.go
matoszz: I think I added it when debugging the trust center graphapi tests, but I don't believe it's needed so removing

internal/objects/upload/handler.go
matoszz: it was easier to do it a single time in this location than it was to set it earlier up in the `if` check being run after the content type detection is re-attempted after the original reader is re-set

internal/objects/validators/mime.go
matoszz: inadvertent change from a copy / paste out of an older branch

pkg/objects/adapter.go
matoszz: For full compatibility we could add it but none of the existing schema hooks performed more than the `One` operation types

pkg/objects/upload.go
matoszz: its just an opportunistic population

pkg/objects/upload.go
matoszz: added TODO to the code around this

config/config-dev.example.yaml
```diff
@@ -146,16 +146,35 @@ entConfig:
...
-  # if switching to s3, be sure to use the below defaultBucket instead
-  defaultBucket:
-    "file_uploads"
-    # uncomment and update to enable s3, otherwise it will default to local storage
-    # accessKey: "REDACTED"
-    # secretKey: "REDACTED"
-    # region: "us-west-2"
-    # defaultBucket: "openlane-development"
-    # provider: "s3"
+    devMode: true
```
golanglemonade: indentation is over too far here looks like

internal/ent/hooks/file.go
```diff
@@ -48,16 +52,54 @@ func HookFileDelete() ent.Hook {
 					return nil, err
 				}
 
+				// This query is intentionally being modified to return WithIntegrations and WithSecrets so that
```
golanglemonade: It looks like you removed the `With...()` from the query but need to update the comment

internal/ent/hooks/file.go
golanglemonade: the failure would rollback the transaction so the file object isnt actually deleted in our db

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/343

### Changes Preview:

PR #1395, opened 2025-10-16
Added impersonationMiddleware in the auth middleware and added a skipper func to skip bearer authentication for impersonated requests.

### LINKED ISSUE #1361: [Bug] Impersonation Route should include impersonation middleware

We can do `v1/impersonation/start` request, but the `v1/impersonation/end` errors on parsing the token from the headers. 

1. Start an impersonation request:
```
curl -H 'Content-Type: application/json' -H "Authorization: Bearer tolp_REDACTED" http://localhost:17608/v1/impersonation/start -X POST -d '{"target_user_id":"01K713329YR7GT7VKXAV0ES9EN", "type": "job", "reason": "test mitb impersonation", "organization_id":"01K71333QN5VYD3HRDEJPMNEHW"}' |jq

{
  "success": true,
  "token": "REDACTED",
  "expires_at": "2025-10-08T13:35:22.248792-06:00",
  "session_id": "01K72HPGR7MG1TZX45HWD960MV",
  "message": "Impersonation session started successfully"
}
```

This will give you a `token` and `session_id`.

2. Now try to `end` the session using `v1/impersonation/end`, and you will get an error

```
curl -H 'Content-Type: application/json' -H "Authorization: Impersonation REDACTED_JWT" 'http://localhost:17608/v1/impersonation/end' -d '{"session_id":"01K72HPGR7MG1TZX45HWD960MV", "reason":"we done here"}'
{"success":false,"error":"could not parse bearer token from authorization header"}
```

You should be able to use the `Impersonation` token to end a impersonation request. 

It's of note that if you instead pass it as a `Bearer` token, you get a different error, because of the different setup of an impersonation token vs. a regular JWT. 
```
{"success":false,"error":"token is malformed: could not JSON decode claim: json: cannot unmarshal array into Go struct field Claims.scopes of type tokens.PermissionScopes"}
```

**Relevant Code**

This endpoint currently has `authenticated` endpoint middleware:

https://github.com/theopenlane/core/blob/main/internal/httpserve/route/impersonation.go#L39

Impersonation Middleware: 
https://github.com/theopenlane/core/blob/403a2fd2d47e4cc33cd5a3322439dc3eaa4ddb35/pkg/middleware/impersonation/impersonation.go#L26

I think we should just add this middleware _before_ the auth middleware - since this just continues if there isn't an impersonation token, it would go to the regular auth middleware after:

https://github.com/theopenlane/core/blob/403a2fd2d47e4cc33cd5a3322439dc3eaa4ddb35/internal/httpserve/route/router.go#L684-L691

I believe the tests are passing because its adding the auth context so the request doesn't need to go through the auth middleware so this wasn't caught: 

https://github.com/theopenlane/core/blob/403a2fd2d47e4cc33cd5a3322439dc3eaa4ddb35/internal/httpserve/handlers/impersonation_test.go#L301-L310

aritradevelops:
Closes #1361

# Description

This PR adds impersonationMiddleware within Auth Middleware to identify and check impersonated requests. And adds a skipper function to skip bearer token checking in case of impersonated requests.

# Final Result

https://github.com/user-attachments/assets/e76af060-fba9-40a6-b176-7486162c0ab8


aritradevelops:
Hi @golanglemonade I have made the changes as suggested by you in #1361 . Please let me know if any further modification is required. Thanks!

golanglemonade (approved):
thanks @aritradevelops!

PR #1400, opened 2025-10-16
make a more generic function to get worker api creds and host

golanglemonade:
embed a `OpenlaneConfig` into all worker configs instead of using the CustomDomain one for all of them

### DIFF
--- a/internal/graphapi/controlimplementation_test.go
+++ b/internal/graphapi/controlimplementation_test.go
@@ -175,8 +175,10 @@ func TestQueryControlImplementations(t *testing.T) {
 
 	// create multiple controlImplementations to be queried using testUser
 	numCIs := 5
+	ciIDs := []string{}
 	for range numCIs {
-		(&ControlImplementationBuilder{client: suite.client}).MustNew(testUser.UserCtx, t)
+		ci := (&ControlImplementationBuilder{client: suite.client}).MustNew(testUser.UserCtx, t)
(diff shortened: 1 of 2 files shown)


pkg/corejobs/openlane.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"net/url"
+	"strings"
+
+	"github.com/theopenlane/core/pkg/corejobs/internal/olclient"
+	"github.com/theopenlane/core/pkg/openlaneclient"
+)
+
+// OpenlaneConfig contains the configuration for connecting to the Openlane API.
+type OpenlaneConfig struct {
```
matoszz, line 12: lol coincidentally in my branch for some object storage updates I also just made a struct with this name

pkg/corejobs/openlane.go
```diff
@@ -0,0 +1,77 @@
...
+		if strings.HasPrefix(token, prefix) {
+			return true
+		}
+	}
+	return false
+}
+
+// getOpenlaneClient creates and returns a new Openlane client using the provided configuration.
+// It configures the client with the appropriate base URL and authentication credentials.
+func (c *OpenlaneConfig) getOpenlaneClient() (olclient.OpenlaneClient, error) {
```
matoszz, line 57: is there a reason to make the struct for the config exported but this function non-exported?

PR #1411, opened 2025-10-20
fix(ci): Update buildkite pipeline to only run steps based on files

### LINKED ISSUE #289: [Bug] Update buildkite pipeline to only run steps based on files changed in the repository
We pay for minutes on buildkite's hosted runners, so should not run builds unnecessarily. For example, when opening this [PR](https://github.com/theopenlane/core/pull/288) which added a new file to our .github/ directory, we ran an entire build / build pipeline (https://buildkite.com/theopenlane/core/builds/1568) unnecessarily. 

Pipeline configuration should execute steps for pre-check, etc., based on modified files in the repository so things like README.md changes, .github/ changes, etc., should not kick off any steps (or at least be constrained to the steps requisite the change)

kawpii:
Fixes #289 

Updated `pipeline.yaml` according [comment](https://github.com/theopenlane/core/issues/289#issuecomment-3368765711) except the `docker-build` step, [since it depends on `go-builds` step](https://github.com/theopenlane/core/blob/2e096b55dfcff0c922055a39c1c27afa99730766/.buildkite/pipeline.yaml#L445)

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -16,7 +16,6 @@ env:
   GOLANGCI_LINT_CACHE: $HOME/.cache/golangci-lint
   HELM_CHART_REPO: https://github.com/theopenlane/openlane-infra.git
   HELM_CHART_PATH: charts/openlane
-  BUILDKITE_PULL_REQUEST_BASE_BRANCH: main
 
 steps:
   - group: ":knife: Pre-check"
@@ -29,7 +28,10 @@ steps:
           size: $RUNNER_LARGE
           location: "NYC"
         cancel_on_build_failing: true
-        if_changed: "{go.{mod,sum},internal/**,pkg/**}"
+        if_changed:
+          - go.{mod,sum}
+          - internal/**
+          - pkg/**
         plugins:
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
@@ -46,7 +48,10 @@ steps:
         key: "generate_config"
         cancel_on_build_failing: true
         if_changed:
-          "{go.{mod,sum},internal/**,pkg/**,config/**}"
+          - go.{mod,sum}
+          - internal/**
+          - pkg/**
+          - config/**
         plugins:
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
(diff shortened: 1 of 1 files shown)


matoszz:
@kaushalyap this is the failure output from the build, FYI

```

Unrecoverable error, skipping retries
--
  | buildkite-agent: fatal: Failed to upload and process pipeline: Pipeline upload rejected: One of the steps you provided was invalid: Error parsing `if` expression: Unexpected `diff`: expected comparison operator, `&&`, `\|\|`, `?`, or EOF (line 1, column 5-8)
  | 🚨 Error: The command exited with status 1

```

kawpii:
I found [`if_then` agent attribute](https://buildkite.com/docs/pipelines/configure/step-types/trigger-step#agent-applied-attributes) while looking through mono repo plugin, I think this is exactly what we need.

golanglemonade:
> I found [`if_then` agent attribute](https://buildkite.com/docs/pipelines/configure/step-types/trigger-step#agent-applied-attributes) while looking through mono repo plugin, I think this is exactly what we need.

Oh, that's new I think 👀

golanglemonade:
Nice find @kaushalyap! Looks like it needs to be added to a few more steps; right now the docker build + upload pr reports are being run but their dependent steps are not. These should have the same `if_changed` as the go tests job. 

The create draft config pr should only run if `config/*` is changed.

kawpii:
> right now the docker build + upload pr reports are being run but their dependent steps are not.

@golanglemonade Actually, if this PR is correct there should be no builds at all since it only change `pipeline.yaml`. I'm actually surprised how `image-build` group runs when it depends on `go-builds`.

For now, let's try adding `if_changed` to `image-build` group and see how it behaves

kawpii:
@golanglemonade I reverted the changes, now it does not identify `if_changed` as a valid command at all.

golanglemonade:
@kaushalyap can you try putting it back; looks like some of the runners are on an older bk agent version so I'm upgrading them now. I saw it work on one agent on a rerun

kawpii:
@golanglemonade It seems now everything works fine, but I'll clean things up a bit.

kawpii:
@golanglemonade Now things look more readable with list pattern.

golanglemonade (approved):

PR #1447, opened 2025-10-25
feat: implement organization ownership transfer (#69)

### LINKED ISSUE #69: [Feature Request] Allow transfer of org owner
Implement resolver to transfer ownership of organization 
     - if done by the current owner, request should go through if the new user exists in the org
     - current user should be updated from owner -> admin
     - new owner should be updated as owner
     - any related tuples in FGA should be added/deleted as required
     - If the user is not in the organization, they should be sent an invitation to the organization, once accepted the transfer of ownership should happen
     - if done by another user in the org that is not the owner, request should be denied

manab-pr:
feat: implement organization ownership transfer (#69)

  PR Description

  ## Summary

  Implements organization ownership transfer functionality as requested in #69. Organization owners can now transfer ownership to another user
  via the `transferOrganizationOwnership` GraphQL mutation.

  **Two transfer paths:**
  - **Existing member**: Direct role swap - new owner gets OWNER role, previous owner demoted to ADMIN
  - **Non-member**: Ownership transfer invitation sent - transfer completes when invitation is accepted

  ## Changes

  ### GraphQL API
  - Added `transferOrganizationOwnership` mutation in `orgextended.graphql`
  - Returns `OrganizationTransferOwnershipPayload` with organization and invitation status

  ### Database Schema
  - Extended `Invite` schema with `ownership_transfer` boolean field
  - Tracks invitations intended for ownership transfer vs regular membership

  ### Business Logic
  - **Resolver** (`orgextended.resolvers.go`):
    - Verifies caller is current owner
    - Checks if target user is already a member
    - Direct transfer: Updates both roles atomically
    - Invitation path: Creates invite with `ownership_transfer=true`

  - **Invite Hook** (`hooks/invite.go`):
    - Detects `ownership_transfer` flag on invitation acceptance
    - Automatically demotes all current owners to ADMIN role
    - Ensures only one owner exists after transfer

  ### Bug Fixes
  - Fixed duplicate group membership constraint violation in `managedgroups.go`
  - Added existence check before creating group memberships
  - Removed duplicate OWNER enum value in `orgmembership.go`
  - Added OWNER to valid invite role values in `role.go`

  ## Test Plan

  **Unit Tests** (`orgtransfer_test.go`):
  - ✅ `TestTransferOrganizationOwnership_ExistingMember`: Verifies direct transfer
    - Creates owner and member users
    - Transfers ownership
    - Asserts new owner has OWNER role
    - Asserts old owner demoted to ADMIN role

  - ✅ `TestTransferOwnershipViaInvitation`: Verifies invitation path
    - Creates owner and non-member users
    - Sends ownership transfer invitation
    - Accepts invitation
    - Asserts ownership transferred correctly
    - Asserts old owner demoted to ADMIN

  **Manual Testing**:
  - [ ] Transfer to existing admin - verify role change
  - [ ] Transfer to existing member - verify role change
  - [ ] Transfer to non-member - verify invitation sent
  - [ ] Accept ownership invitation - verify transfer completes
  - [ ] Attempt transfer as non-owner - verify rejection
  - [ ] Verify FGA tuples updated correctly
  - [ ] Verify managed groups updated (Admins/Viewers/All Members)

  ## Security Considerations

  - Authorization check: Only current OWNER can initiate transfer
  - Privacy contexts used appropriately for authorized operations
  - FGA tuples automatically updated via existing hooks
  - No breaking changes to existing invite flow

  ## Files Modified

  - `internal/graphapi/schema/orgextended.graphql` - GraphQL schema
  - `internal/graphapi/orgextended.resolvers.go` - Transfer logic
  - `internal/graphapi/query/organization.graphql` - Test queries
  - `internal/ent/schema/invite.go` - Added ownership_transfer field
  - `internal/ent/hooks/invite.go` - Transfer on acceptance
  - `internal/ent/hooks/managedgroups.go` - Fixed duplicate bug
  - `internal/ent/schema/orgmembership.go` - Removed duplicate enum
  - `pkg/enums/role.go` - Added OWNER to valid values
  - `internal/graphapi/orgtransfer_test.go` - Comprehensive tests

  ## Closes

  Closes #69


internal/ent/schema/orgmembership.go
```diff
@@ -51,7 +51,6 @@ func (OrgMembership) Fields() []ent.Field {
 			Annotations(
 				entgql.OrderField("ROLE"),
 			).
-			Values(string(enums.RoleOwner)). // adds owner to the allowed values
```
golanglemonade, line 54: Groups shouldn't have owners, any reason this needed to be removed here and updated for the groups?

pkg/enums/role.go
```diff
@@ -17,9 +17,9 @@ var (
 )
 
 // Values returns a slice of strings that represents all the possible values of the Role enum.
-// Possible default values are "ADMIN", "MEMBER"
+// Possible default values are "OWNER", "ADMIN", "MEMBER"
 func (Role) Values() (kinds []string) {
-	for _, s := range []Role{RoleAdmin, RoleMember} {
+	for _, s := range []Role{RoleOwner, RoleAdmin, RoleMember} {
```
golanglemonade: Ah, I think I see you did the change for invites, however, this shouldn't be changed here because it affects groups as well. Instead you should add the same line that was deleted in orgmembership to invites to add the `Owner` role there.

golanglemonade:
@manab-pr Looks like there are some failing graphapi tests as well; you should be able to see the buildkite failure and run the same tests locally.

internal/ent/schema/orgmembership.go
manab-pr: The whole thought behind this was :  
keeping OWNER for organizations (doesn't remove it or add it to groups). Here's what happened:

  The .Values(string(enums.RoleOwner)) line was removed from orgmembership.go because it was redundant. The Role enum in pkg/enums/role.go
  already includes OWNER in its Values() method:

  // pkg/enums/role.go
  func (Role) Values() (kinds []string) {
        for _, s := range []Role{RoleOwner, RoleAdmin, RoleMember} {
                kinds = append(kinds, string(s))
        }
        return
  }

  So organizations still support OWNER role - it's just defined at the enum level instead of being explicitly added in the schema.

  Groups remain unchanged - they only use ADMIN and MEMBER (OWNER is never used for groups). (here's where i am wrong , user can actually use api and use exposed owner role in here as well )

pkg/enums/role.go
manab-pr: You're right - I did a mistake . By adding OWNER to the global Role.Values() method, it becomes available in the GraphQL schema for all entities using the Role enum, including GroupMembership where owners don't make sense.

  The better approach is to:
  1. Revert the change to role.go (remove RoleOwner from the global Values() method)
  2. Keep the .Values(string(enums.RoleOwner)) constraint in orgmembership.go
  3. Add .Values(string(enums.RoleOwner)) to the invite.go schema as well
  This way, OWNER is only exposed in the GraphQL API for organizations and invites, not for groups.

manab-pr:
@golanglemonade  All tests are now passing. Ready for re-review, Please review and let me know if everything looks correct!

internal/graphapi/schema/orgextended.graphql
```diff
@@ -40,4 +40,31 @@ extend type Mutation{
...
         members: [OrgMembersInput!]
     ): OrganizationCreatePayload!
+    """
+    Transfer ownership of an organization to another user
+    """
+    transferOrganizationOwnership(
+        """
+        ID of the organization
+        """
+        organizationID: ID!
```
golanglemonade: I missed this before - you shouldn't need this. When a user an authenticated user will have an organization ID in their JWT ->  which we add to the AuthenticatedUserContext and we don't need this as input.

internal/graphapi/orgextended.resolvers.go
```diff
@@ -56,6 +59,124 @@ func (r *mutationResolver) CreateOrganizationWithMembers(ctx context.Context, or
...
+	}
+
+	c := withTransactionalMutation(ctx)
+
+	// Step 2: Verify organization exists
+	org, err := c.Organization.Get(ctx, organizationID)
+	if err != nil {
+		return nil, parseRequestError(err, action{action: ActionGet, object: "organization"})
+	}
+
```
golanglemonade, line 86: this should be updated after the schema is so that you do 
```suggestion
	// Step 1: Get current user from context
	au, err := auth.GetAuthenticatedUserFromContext(ctx
	if err != nil {
		return nil, parseRequestError(err, action{action: ActionCreate, object: "invite"})
	}
	currentUserID = au.SubjectID
	organizationID. := au.OrganizationID
	
	if organizationID == "" {
	     log.Error().Msg("unable to determine organization for ownership transfer")
	     
		return parseRequestError(err, action{action: ActionCreate, object: "invite"})
	}
```

internal/graphapi/orgextended.resolvers.go
```diff
@@ -56,6 +59,124 @@ func (r *mutationResolver) CreateOrganizationWithMembers(ctx context.Context, or
...
+		return nil, newNotFoundError("user")
+	}
+
+	// Step 5: Check if new owner is already a member
+	newOwnerMembership, err := c.OrgMembership.Query().
+		Where(
+			orgmembership.OrganizationID(organizationID),
+			orgmembership.UserID(newOwnerID),
+		).
+		First(ctx)
```
golanglemonade: I think `Only` is safer, there shouldn't ever be duplicates if DeletedAtIsNil is added to the `Where`
```suggestion
		Only(ctx)
```

internal/graphapi/orgextended.resolvers.go
```diff
@@ -56,6 +59,124 @@ func (r *mutationResolver) CreateOrganizationWithMembers(ctx context.Context, or
...
+		return nil, newPermissionDeniedError()
+	}
+
+	// Step 4: Verify new owner user exists
+	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
+	newOwnerUser, err := c.User.Get(allowCtx, newOwnerID)
+	if err != nil {
+		log.Error().Err(err).Str("new_owner_id", newOwnerID).Msg("new owner user not found")
+		return nil, newNotFoundError("user")
+	}
```
golanglemonade: This should be removed, if the new owner isn't an existing user we should still send them the invitation.

internal/graphapi/schema/orgextended.graphql
```diff
@@ -40,4 +40,31 @@ extend type Mutation{
...
+        """
+        ID of the new owner
+        """
+        newOwnerID: ID!
```
golanglemonade: Instead of an `ID`, this should be an email. This way the user can be invited if they aren't already a user in the system. There would be no easy way to get this ID either if the user wasn't already in the organizatino. 
```suggestion
        newOwnerEmail: String!
```

internal/graphapi/orgtransfer_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/orgmembership"
+	"github.com/theopenlane/core/internal/ent/generated/privacy"
+	"github.com/theopenlane/core/pkg/enums"
+	"github.com/theopenlane/iam/auth"
+	"gotest.tools/v3/assert"
+)
+
+// TestTransferOrganizationOwnership_ExistingMember tests transferring ownership to an existing org member
+func TestTransferOrganizationOwnership_ExistingMember(t *testing.T) {
```
golanglemonade: This doesn't really test the resolvers, instead its just writing the same queries that are in there - which becomes an issue if that code is changed and the paths aren't checked but it looks like you have test coverage.

You should be able to follow a similar test pattern to `TestMutationCreateInvite` to test this new resolver instead.

manab-pr:
@golanglemonade   Thanks for the detailed review.
  I needed a bit of  time to get to all of it .
  I think i understand the issues :
  1. Will remove organizationID from schema and get it from authenticated user context
  2. Will change newOwnerID to newOwnerEmail to support inviting users not yet in the system  
  3. Will remove the existing user check and allow sending invitations
  4. Will change First() to Only() for the membership query
  5. Will refactor the test to follow TestMutationCreateInvite pattern

  Working on these changes now.

manab-pr:
Thanks for the thorough review! I've addressed all the feedback:

  **Schema Changes:**
  -  Removed `organizationID` parameter - now automatically derived from the authenticated user's JWT context
  -  Changed `newOwnerID: ID!` to `newOwnerEmail: String!` to support inviting users who aren't in the system yet

  **Resolver Updates:**
  -  Now using `GetAuthenticatedUserFromContext()` instead of `GetSubjectIDFromContext()` to get both user ID and organization ID from the
  context
  - Added validation to ensure `organizationID` is present
  -  Removed the user existence check (lines 100-106) - now allows sending invitations regardless of whether the user exists
  -  Changed `First()` to `Only()` for the membership query since there shouldn't be duplicates

  **Test Improvements:**
  -  Completely refactored the test to follow the `TestMutationCreateInvite` pattern
  -  Now uses table-driven tests that actually call the resolver through the GraphQL client
  -  Includes test cases for existing members, non-members, new users, permission denied, and invalid input

@golanglemonade  Please review and let me know if everything looks correct!

golanglemonade:
@manab-pr looks like some test failures but the code changes look good:

```

❌  internal/graphapi (3m39.789s) (coverage: 36.8% of statements)
--
  |  
  | DONE 1895 tests, 3 failures in 219.789s
  |  
  | ❌  internal/graphapi (9.894s) (coverage: 1.5% of statements)
  |  
  | DONE 2 runs, 1902 tests, 6 failures in 241.606s
  |  
  | ❌  internal/graphapi (10.309s) (coverage: 1.5% of statements)
  |  
  | DONE 3 runs, 1909 tests, 9 failures in 263.101s
  |  
  | ❌  internal/graphapi (9.86s) (coverage: 1.5% of statements)
  |  
  | === Failed
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_not_owner,_permission_denied (0.02s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_different_org_owner,_no_access (0.07s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership (1.98s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_not_owner,_permission_denied (re-run 1) (0.02s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_different_org_owner,_no_access (re-run 1) (0.06s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership (re-run 1) (1.25s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_not_owner,_permission_denied (re-run 2) (0.02s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_different_org_owner,_no_access (re-run 2) (0.06s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership (re-run 2) (1.35s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_not_owner,_permission_denied (re-run 3) (0.02s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership/Transfer_different_org_owner,_no_access (re-run 3) (0.06s)
  | === FAIL: internal/graphapi TestMutationTransferOrganizationOwnership (re-run 3) (1.20s)


```

manab-pr:
@golanglemonade  All tests are finally passing.

Here’s the summary of the fix:

The non-owner test was still checking the old “permission denied” text, but since newPermissionDeniedError() now returns notAuthorizedErrorMsg (“you are not authorized…”), it needed an update.

The “other org owner” test was running with that user’s own org context, which mistakenly let the mutation create an ownership-transfer invite instead of returning notFoundErrorMsg.

Fix: Updated internal/graphapi/orgtransfer_test.go to check for notAuthorizedErrorMsg and build the “different owner” auth context using testUser1’s org ID, so the resolver correctly triggers the privacy guard and returns notFoundErrorMsg.

Everything’s green now .
Please review when you get a chance. Thanks!

golanglemonade (approved):

golanglemonade:
nice, thanks @manab-pr!

PR #1454, opened 2025-10-28
SCIM 2.0 Protocol support

matoszz:
Implements SCIM 2.0 protocol support for automated user and group provisioning from external identity providers.

- Added SCIM-specific fields to users and groups tables (`scim_external_id`, `scim_username`, `scim_active`, etc.) with corresponding database migrations
- Implemented SCIM 2.0 REST API endpoints:
  - User endpoints: GET/POST `/scim/v2/Users`, GET/PUT/PATCH/DELETE `/scim/v2/Users/:id`
  - Group endpoints: GET/POST `/scim/v2/Groups`, GET/PUT/PATCH/DELETE `/scim/v2/Groups/:id`
  - Support for filtering, pagination, partial updates, and member management
- Added attribute transformation layer between SCIM schema and internal data models with comprehensive error handling following RFC 7644
- Added code wrappers for OpenAPI specifications

NOTE: this feature requires:
- an organization API token
- SSO to be enforced on the organization
- Allowed domains to be configured properly matching domain patterns for provisioned users and groups

### DIFF
--- a/internal/httpserve/handlers/scim/group_handler.go
+++ b/internal/httpserve/handlers/scim/group_handler.go
@@ -20,7 +20,6 @@ import (
 	"github.com/theopenlane/core/internal/ent/generated/organization"
 	"github.com/theopenlane/core/internal/ent/generated/orgmembership"
 	"github.com/theopenlane/core/internal/ent/hooks"
-	"github.com/theopenlane/core/internal/ent/validator"
 	"github.com/theopenlane/core/pkg/middleware/transaction"
 	"github.com/theopenlane/utils/contextx"
 )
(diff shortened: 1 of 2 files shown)


internal/httpserve/handlers/scim/group_handler.go
```diff
@@ -0,0 +1,545 @@
...
+	} else {
+		switch pathStr {
+		case "externalid":
+			if strVal, ok := op.Value.(string); ok && strVal != "" {
+				update.SetScimExternalID(strVal)
+				*modified = true
+			}
+		case "displayname":
+			if strVal, ok := op.Value.(string); ok && strVal != "" {
+				name := strings.ReplaceAll(strings.ToLower(strVal), " ", "-")
```
golanglemonade, line 367: We have group name restrictions we might need to handle: https://github.com/theopenlane/core/blob/main/internal/ent/validator/specialcharacter.go#L12

https://github.com/theopenlane/core/blob/main/internal/ent/hooks/group.go#L294-L302

internal/httpserve/handlers/scim/validation.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	if org.Edges.Setting == nil {
+		return fmt.Errorf("%w for organization %s", ErrOrgSettingsNotFound, orgID)
+	}
+
+	if !org.Edges.Setting.IdentityProviderLoginEnforced {
+		return ErrSSONotEnforced
+	}
+
```
golanglemonade: It doesn't look like you need the organization object here so you can drop the join and just do:

```suggestion
	orgSetting, err := client.OrganizationSetting.Query().
		Where(organizationsetting.OrganizationID(orgID)).
		Only(ctx)
	if err != nil {
		if generated.IsNotFound(err) {
			return fmt.Errorf("%w: organization %s", ErrOrgNotFound, orgID)
		}
		
		return fmt.Errorf("failed to query organization: %w", err)
	}

	if !orgSetting.IdentityProviderLoginEnforced {
		return ErrSSONotEnforced
	}
```

golanglemonade (approved):

aaguiarz:
Are you synchronizing SCIM data to OpenFGA? :)

PR #1456, opened 2025-10-28
Feature/Update API key generation to use argon2id

### LINKED ISSUE #284: [Feature Request] Update API key generation to use argon2id
Similar to the way we persist a hash of the user's inputted password [here](https://github.com/theopenlane/core/blob/main/internal/ent/hooks/user.go#L49), we should update the API token and Personal Access Tokens schema and token creation to leverage the same hashing.

You can use `utils.CreateDerivedKey`, the same way the user password is done today:

https://github.com/theopenlane/core/blob/b6d81f23003699f6fd9efc5869cecba082b06606/internal/ent/hooks/user.go#L49

Instead of a hook here, you would update the default function on both the api and personal access token schemas, e.g. here on api tokens

https://github.com/theopenlane/core/blob/b6d81f23003699f6fd9efc5869cecba082b06606/internal/ent/schema/apitoken.go#L63

Additionally, the auth middleware needs to be updated to use derived key instead of the token in the request:

https://github.com/theopenlane/core/blob/b6d81f23003699f6fd9efc5869cecba082b06606/pkg/middleware/auth/auth.go#L289

Example from user password login: https://github.com/theopenlane/core/blob/b6d81f23003699f6fd9efc5869cecba082b06606/internal/httpserve/handlers/login.go#L59-L63

Achiket123:
resolves #284


Achiket123:
Hey @golanglemonade .

internal/ent/schema/jobrunner_registration_token.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/theopenlane/core/pkg/models"
 	"github.com/theopenlane/entx/accessmap"
 	"github.com/theopenlane/entx/history"
-	"github.com/theopenlane/utils/keygen"
+	"github.com/theopenlane/magiclinks/pkg/keygen"
```
golanglemonade: This looks like an incorrect change
```suggestion
	"github.com/theopenlane/utils/keygen"
```

internal/ent/schema/jobrunner_token.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/theopenlane/core/pkg/models"
 	"github.com/theopenlane/entx/accessmap"
 	"github.com/theopenlane/entx/history"
-	"github.com/theopenlane/utils/keygen"
+	"github.com/theopenlane/magiclinks/pkg/keygen"
```
golanglemonade: Same here
```suggestion
	"github.com/theopenlane/utils/keygen"
```

internal/ent/schema/organizationsetting.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/theopenlane/core/pkg/enums"
 	"github.com/theopenlane/core/pkg/models"
 	"github.com/theopenlane/iam/entfga"
-	"github.com/theopenlane/utils/keygen"
+	"github.com/theopenlane/magiclinks/pkg/keygen"
```
golanglemonade: Same here
```suggestion
	"github.com/theopenlane/utils/keygen"
```

Achiket123:
Hey @golanglemonade this repo is large and slowed my laptop (too slow) Restarted multiple times and still not running properly. I am using vim through WSL and causing some errors. Please review and tell the errors 😢

pkg/middleware/auth/auth.go
```diff
@@ -581,14 +580,49 @@ var (
...
+		// we store a derived key in the DB, so we must compare the provided token
+		// against the stored derived keys. This requires iterating candidate tokens
+		// and verifying the derived key. We load owner and organizations for downstream
+		// checks.
+		pats, err := db.PersonalAccessToken.Query().WithOwner().WithOrganizations().All(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, p := range pats {
```
golanglemonade, line 592: This is going to be really inefficient as the number of tokens grows. Lets try a small update to the schema + generation so this is better on the lookup. 

Instead of the format `prefx_secret` (e.g. `tolp_0W8b9HD0ZHgawJbJddx8q9oeK1xvYpP9`) lets try:

`<prefix>_<public_id>_<secret>` 

the token would then look something like 
```
tolp_zf47B2xk9Y4yJtqH_0W8b9HD0ZHgawJbJddx8q9oeK1xvYpP9
```

In the database 
- we store the the `<secret>` - which is the argon2 hash of the secret, probably makes sense to make a new column here called `token_hash` to maintain backwards compatibility and we will eventually drop the `token` column (no need to worry about that on this PR)
- store the `<public_id>` in a `token_fp` column

Now when a request comes in with something like `tolp_zf47B2xk9Y4yJtqH_0W8b9HD0ZHgawJbJddx8q9oeK1xvYpP9`

we can compute the fingerprint 
```
fingerprint = keygen.GenerateSHA256Hmac(secret, raw_token)

at, err := db.APIToken.
    Query().
    Where(apitoken.TokenFP(fingerprint)).
    Only(ctx)
```
and then verify just the secret 
```
valid, err := passwd.VerifyDerivedKey(p.Token, token)
if err != nil {
    return nil, err
}
```

If we add an index for `token_fp` it should speed up the lookup. 
We'd update the hooks to look something like this: 

```
			// generate raw token
			secret := keygen.Secret()
            publicID := keygen.AlphaNumeric(16)
            prefix := "tolp"
            
            rawToken := fmt.Sprintf("%s_%s_%s", prefix, publicID, secret)

			// hash the token for storage
			hash, err := passwd.CreateDerivedKey(secret)
			if err != nil {
				return nil, err
			}

			// set the hashed token for storage
			m.SetTokenHash(hash)
			
			tokenFP := keygen.GenerateSHA256Hmac(publicID)
			m.SetTokenFingerprint(tokenFP)
			
			....
			
			// return the full token to the user
			pat.Token = rawToken
```

This allows us to keep an O(1) lookup, doesnt store the id or the secret in plain text in the database. 

Let me know if this makes sense and if you want to take a shot at updating this

Achiket123:
Hey @golanglemonade  I am facing this error while trying to generate 

internal\ent\hooks\personalaccesstoken.go
image

internal\ent\hooks\apitoken.go
image



image

golanglemonade:
> Hey @golanglemonade I am facing this error while trying to generate
> 
> internal\ent\hooks\personalaccesstoken.go image
> 
> internal\ent\hooks\apitoken.go image
> 
> image

Did you run generate before updating that hook? I would updated the schema with the new fields, comment out all the code inside the hook, run generate, and then uncomment - generate is going create those functions for you.

Achiket123:
Hey @golanglemonade  I tried everything but i am stuck at this error 
{"level":"fatal","error":"template: pattern matches no files: `templates\\invite.html`","template":"invite.html","time":"2025-10-30T03:09:47+05:30","message":"could not parse template"}
exit status 1

Can you help with this one

golanglemonade:
> Hey @golanglemonade I tried everything but i am stuck at this error {"level":"fatal","error":"template: pattern matches no files: `templates\\invite.html`","template":"invite.html","time":"2025-10-30T03:09:47+05:30","message":"could not parse template"} exit status 1
> 
> Can you help with this one

Very weird, I've never seen this before. Was that when you were running `task generate` or trying to run the server with something like `task run-dev`?

I grabbed your branch, commented out the hooks, and then ran generate and everything worked, and then uncommented the hooks. If you want to grab this branch and work from there: https://github.com/theopenlane/core/pull/new/feat-argon2-tokens or cherry-pick the commits onto your branch to see if its all working as expected

Achiket123:
Hey @golanglemonade  Thankyou for your help , It was frustrating , every time i run `task generate` or `task generate:ci` i received the same error 😭 😭 . I guess its  completed now, will you please review it.

golanglemonade:
@Achiket123 when reviewing this we determined that the argon2 hash performance will be too slow for the use case so we need to move to a slightly different pattern and use HMAC for the secret instead, and keep the public_id as plain text. I'm going to mark this as `accepted` since you accomplished the goal of the original issue issue, but don't want to put different requirements on you because we changed our minds :) really appreciate the work here!

Achiket123:
> @Achiket123 when reviewing this we determined that the argon2 hash performance will be too slow for the use case so we need to move to a slightly different pattern and use HMAC for the secret instead, and keep the public_id as plain text. I'm going to mark this as `accepted` since you accomplished the goal of the original issue issue, but don't want to put different requirements on you because we changed our minds :) really appreciate the work here!

my hactoberfest shirt 😢 !!!

Its ok, Loved working under you. 
BTW @golanglemonade  will you be participating in Google summer of code ??
I will be contributing more in the organization in future because i got comfortable with the codebase :hehe:

Also I would like to implement the use of HMAC for the secret , can you please specify this requirement.

golanglemonade:
Ah I misunderstood; I thought I could also tag them as hacktoberfest-accepted and be counted without merging. Let me update the requirements and can help you get this in before tomorrow if you want. It should be just a few changes to change the public secret to not be hashed, and the secret to use the hmac hash instead of argon2. You'll also need to update the auth middleware to lookup by the public key instead of all tokens (and fix failing tests). 

I'll get the specific requirements shortly though!  @matoszz is going to make some updates to the keygen library that should enable this to be done. 

Yeah, looking at applying to Google summer of code for 2026. 🎉

golanglemonade:
@Achiket123 also if you want to join our discord, I'm sure I can find you another simple issue to do to make sure you get your hacktoberfest badge on time: https://discord.gg/4fq2sxDk7D - just tag me in #contributing (@sfunk) if you join

golanglemonade:
@Achiket123  closing this one out in favor of https://github.com/theopenlane/core/issues/1475. I would recommend adding the new fields listed in that ticket first, running generate, and then adding the hook + middleware logic so all the functions are in place. It's not worth trying to take this branch as is and update because removing a field from the schema can be kind of a pain if you don't have to.

PR #1474, opened 2025-11-03
Questionnaire schemas

adelowo:
templates -> assesments ( created from a template, can be internal or external, then shared ) -> assessment responses -> document_data


internal/ent/privacy/rule/assesment.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/theopenlane/iam/auth"
+
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/assessment"
+	"github.com/theopenlane/core/internal/ent/generated/privacy"
+)
+
+// AllowIfAssessmentCreatedBy allows users to edit assessments they created
+func AllowIfAssessmentCreatedBy() privacy.AssessmentMutationRuleFunc {
```
golanglemonade: This shouldn't be needed and we should be using FGA to determine access vs. writing a separate rule here. When the assessment is created a tuple should be added allowing the required access.

internal/ent/privacy/rule/assesment.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		return privacy.Skipf("user is not the creator of this assessment")
+	})
+}
+
+// AllowIfAssessmentQueryCreatedBy allows users to query only assessments they created
+func AllowIfAssessmentQueryCreatedBy() privacy.AssessmentQueryRuleFunc {
+	return privacy.AssessmentQueryRuleFunc(func(ctx context.Context, q *generated.AssessmentQuery) error {
+		userID, err := auth.GetSubjectIDFromContext(ctx)
```
golanglemonade: Same here

internal/ent/privacy/token/assesment.go
```diff
@@ -0,0 +1,61 @@
+package token
+
+import (
+	"context"
+
+	"github.com/theopenlane/utils/contextx"
+)
+
+// AssesmentResponseToken that implements the PrivacyToken interface
+type AssesmentResponseToken struct {
```
golanglemonade: We had talked about using anon JWTS for this instead of tokens which i how some of the trust center stuff works to retrieve private docuements? Just wondering if you tried that way out?

internal/ent/schema/assesment.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (a Assessment) Edges() []ent.Edge {
+	return []ent.Edge{
+		uniqueEdgeTo(&edgeDefinition{
+			fromSchema: a,
+			edgeSchema: Template{},
+			field:      "template_id",
+			required:   true,
+		}),
+		defaultEdgeToWithPagination(a, User{}),
```
golanglemonade: Why do we need an edge to user here?

internal/ent/schema/assesment.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func (Assessment) Policy() ent.Policy {
+	return policy.NewPolicy(
+		policy.WithQueryRules(
+			rule.AllowIfAssessmentQueryCreatedBy(),
+		),
+		policy.WithMutationRules(
+			policy.CheckCreateAccess(),
```
golanglemonade, line 86: The fga model needs to be updated to include the `can_create_assessment` type

internal/ent/schema/assesment.go
```diff
@@ -0,0 +1,126 @@
...
+			rule.AllowIfAssessmentCreatedBy(),
+			policy.CheckOrgWriteAccess(),
+		),
+	)
+}
+
+// Annotations of the Assessment
+func (Assessment) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		entfga.SelfAccessChecks(),
```
golanglemonade, line 95: Same here, this won't work without adding it to the fga model

internal/ent/schema/assesment_response.go
```diff
@@ -0,0 +1,208 @@
...
+
+		field.String("token").
+			Immutable().
+			Unique().
+			Comment("for external assessments, we want to be able to track the user").
+			Annotations(
+				entgql.Skip(entgql.SkipAll),
+			).
+			DefaultFunc(func() string {
+				token := keygen.PrefixedSecret("questionnaire")
```
golanglemonade: We'd want these prefixed similar to `tola` and `tolp` I think 
```suggestion
				token := keygen.PrefixedSecret("tolq")
```

internal/ent/schema/assesment_response.go
```diff
@@ -0,0 +1,208 @@
...
+			},
+		}),
+	}
+}
+
+func (AssessmentResponse) Policy() ent.Policy {
+	return policy.NewPolicy(
+		policy.WithQueryRules(
+			rule.AllowIfAssessmentResponseQueryOwner(),
+			privacy.AlwaysAllowRule(),
```
golanglemonade: this is the default so its not needed
```suggestion
```

internal/ent/schema/assesment_response.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+func (ar AssessmentResponse) Mixin() []ent.Mixin {
+	return mixinConfig{
+		excludeTags: true,
+		additionalMixins: []ent.Mixin{
+			newOrgOwnedMixin(ar,
+				withSkipTokenTypesObjects(&token.AssesmentResponseToken{}),
+			),
```
golanglemonade: this should probably be an object owned mixin so it's parent is the `assesment` vs. anyone in the organization? 
```suggestion
			   newObjectOwnedMixin[generated.AssesmentResponse](e,
				withParents(Assessment{}),
				withOrganizationOwner(true),
				withSkipTokenTypesObjects(&token.AssesmentResponseToken{}),
			),
```

internal/ent/schema/template.go
```diff
@@ -53,7 +53,7 @@ func (Template) Fields() []ent.Field {
 			Comment("the name of the template").
 			NotEmpty().
 			Annotations(
-				entgql.OrderField("name"),
+				entgql.OrderField("NAME"),
```
golanglemonade: Capital is only needed on enums AFAIK
```suggestion
				entgql.OrderField("name"),
```

internal/ent/schema/assesment.go
```diff
@@ -0,0 +1,126 @@
...
+func (Assessment) Hooks() []ent.Hook {
+	return []ent.Hook{
+		hook.On(
+			hooks.HookRelationTuples(map[string]string{
+				"assessment_owner": "group",
+			}, "owner"),
+			ent.OpCreate|ent.OpUpdateOne|ent.OpUpdateOne,
+		),
+	}
+}
```
golanglemonade, line 123: Looks like this schema is missing modules

```
func (Assessment) Modules() []models.OrgModule {
	return []models.OrgModule{
	}
}
```

internal/graphapi/assessmentresponse.resolvers.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := setOrganizationInAuthContextBulkRequest(ctx, input); err != nil {
+		log.Error().Err(err).Msg("failed to set organization in auth context")
+
+		return nil, rout.NewMissingRequiredFieldError("owner_id")
+	}
+
+	return r.bulkCreateAssessmentResponse(ctx, input)
+}
+
+// CreateBulkCSVAssessmentResponse is the resolver for the createBulkCSVAssessmentResponse field.
```
golanglemonade: I don't see a reason we'd want bulk assessment responses, can you remove these?

internal/ent/privacy/token/assesment.go
adelowo: yes, this is to just get the basic schema out for now.  Actually will do both, this token and the one like the trust center jwt .

But I can just remove this

internal/ent/schema/assesment.go
adelowo: Maybe in the future, _Internal_ questionnaires can be extended to users in the org and they can see in the dashboard directly as ___Todos___  

But I can remove it

internal/ent/schema/assesment.go
adelowo: yes I am aware. Just wanted to get it out because of the potential conflicts. I have the tuples on the other branch where I am working on logic.

internal/ent/schema/assesment.go
golanglemonade: Good idea; I think we should use auto-created tasks for this though. We can talk about that after I get some task updates done on my pr :)

PR #1487, opened 2025-11-04
Feature/api token fields

Achiket123:


Achiket123:
Hey @golanglemonade  i did some changes can you review it please.

internal/ent/schema/apitoken.go
```diff
@@ -63,6 +63,20 @@ func (APIToken) Fields() []ent.Field {
 				token := keygen.PrefixedSecret("tola") // api token prefix
 				return token
 			}),
+		field.String("token_public_id").
+			Unique().
+			Immutable().
+			Optional().
+			Nillable().
```
golanglemonade: I dont think you need `Nillable()` here as this value shouldn't be cleared and should be not empty
```suggestion
			Optional().
			NotEmpty().
```

internal/ent/schema/apitoken.go
```diff
@@ -63,6 +63,20 @@ func (APIToken) Fields() []ent.Field {
...
+			Comment("public identifier for the token (used to lookup token record)"),
+		field.String("token_secret").
+			Immutable().
+			Optional().
+			Nillable().
```
golanglemonade: This needs to be `Unique()`
```suggestion
			Optional().
			Unique().
			NotEmpty().
```

internal/ent/schema/personalaccesstoken.go
```diff
@@ -63,6 +63,20 @@ func (PersonalAccessToken) Fields() []ent.Field {
 				token := keygen.PrefixedSecret("tolp") // token prefix
 				return token
 			}),
+		field.String("token_public_id").
+			Unique().
+			Immutable().
+			Optional().
+			Nillable().
```
golanglemonade: Same here

```suggestion
			Optional().
			NotEmpty().
```

internal/ent/schema/personalaccesstoken.go
```diff
@@ -63,6 +63,20 @@ func (PersonalAccessToken) Fields() []ent.Field {
...
+			Comment("public identifier for the token (used to lookup token record)"),
+		field.String("token_secret").
+			Immutable().
+			Optional().
+			Nillable().
```
golanglemonade:
```suggestion
			Optional().
			NotEmpty().
			Unique().
```

internal/ent/schema/apitoken.go
```diff
@@ -63,6 +63,20 @@ func (APIToken) Fields() []ent.Field {
 				token := keygen.PrefixedSecret("tola") // api token prefix
```
golanglemonade, line 63: Can you mark this field as deprecated and remove the `DefaultFunc` so we don't keep setting that field. You'll also need to update as `Optional()` to avoid errors with it not being set anymore. 

```
entgql.Directives(
	entgql.Deprecated("Use `token_secret` instead."),
),
```

internal/ent/schema/personalaccesstoken.go
```diff
@@ -63,6 +63,20 @@ func (PersonalAccessToken) Fields() []ent.Field {
 				token := keygen.PrefixedSecret("tolp") // token prefix
 				return token
```
golanglemonade, line 64: this one as well, mark deprecated, optional , and remove default func 

```
entgql.Directives(
	entgql.Deprecated("Use `token_secret` instead."),
),
```

internal/ent/hooks/apitoken.go
```diff
@@ -44,6 +45,16 @@ func HookCreateAPIToken() ent.Hook {
...
+			// generate new key material and store public id + secret
+
+			if v, s, err := tokens.GenerateAPITokenKeyMaterial(); err == nil {
+				// store the pieces; keep legacy `token` field for compatibility
+				m.SetTokenPublicID(v)
+				m.SetTokenSecret(string(s))
+			} else {
+				return nil, err
+			}
+
```
golanglemonade: We need to be able to return the token to the user on create as a single field, the way it is now they will have the secret, id, and no idea what the prefix is. 

To accomplish this you'll use a similar pattern to `PresignedURL`:
1. Add to additional_fields, I would call it `TokenValue`
https://github.com/theopenlane/core/blob/main/internal/ent/generate/templates/ent/additional_fields.tmpl
2. Add an interceptor to return the field similar to this:
https://github.com/theopenlane/core/blob/main/internal/ent/interceptors/file.go#L87

The thing I'm not sure of offhand here is that we only want this returned on `Create` mutations, so we need to ensure that is respected here and returned as the redacted string if its ever requested on updated/get queries.

golanglemonade (commented):
I haven't reviewed the auth middleware part yet, but few things we need to fix on the schema/hook side first. nice first go!

golanglemonade:
@Achiket123 were you planning on picking this up still?

Achiket123:
> @Achiket123 were you planning on picking this up still?

Sorry for delayed submission, I will try to complete this as soon as possible, my semester exams are on. I will complete it shortly. Sorry for this. :bow:  :bow:

golanglemonade:
> > @Achiket123 were you planning on picking this up still?
> 
> 
> 
> Sorry for delayed submission, I will try to complete this as soon as possible, my semester exams are on. I will complete it shortly. Sorry for this. :bow:  :bow: 

No worries at all, just checking in! Good luck with exams!

PR #1489, opened 2025-11-04
Refactor soiree for typed topics and declarative listeners; wholesale update to ent/hooks/event.go

matoszz:
- Refactored `Listener` registration to be declarative; `Eventer.AddMutationListener` stores bindings in a map of entity name → `listeners. emitEventOn` inspects that map to decide whether a mutation should emit an event
- Payload delivery is typed; Each entity gets a `TypedTopic[*MutationPayload]` so listeners can rely on strongly typed payloads instead of re‑casting raw maps
- Mutation emission happens on successful commits. The hook still defers to the ent transaction machinery, but we now ensure that the payload includes both the generated mutation fields and the resolved entity ID so downstream listeners do not need to fetch them manually

This is a `soiree` you won't want to miss :) 

---

> [!NOTE]
> Introduce typed topics and declarative listener bindings in soiree, rewire ent hooks to emit typed mutation events, and update entitlement/Slack listeners, metrics, and tests accordingly.
> 
> - **Soiree (events)**:
>   - **Typed Topics & Listeners**: Add `TypedTopic[T]`, `TypedListener[T]`, `ListenerBinding`, and `EventContext`; wrap/unwrap strongly-typed payloads; support pre/post hooks.
>   - **Declarative Registration**: `EventPool.RegisterListeners` and bindings replace ad‑hoc `On` usage; add `InterestedIn` to query live subscriptions.
>   - **Retry & Execution**: Centralize listener invocation via `listenerItem.call`; propagate context; persist handler results with typed context.
>   - **Worker Pool Counters**: Fix `Submitted/Successful/Failed/CompletedTasks` to return correct values; add helper `uintToInt`.
>   - **Metrics**: Consolidate Redis metrics into `metrics.go`; expose default collectors; examples updated.
>   - **API Changes**: `Listener` now `func(*EventContext) error`; priority/ordering preserved; topic exposes `HasListeners`.
> - **Ent Hooks**:
>   - **Eventer**: Maintain `listeners` map (entity → bindings); `AddMutationListener` and `RegisterListeners` added.
>   - **Emission Predicate**: `emitEventOn` checks live emitter and registered listeners before emitting.
>   - **Typed Mutation Events**: Build typed topic per entity; payload includes `Mutation`, `Operation`, `EntityID`, optional ent client; emit after commit.
>   - **Utilities**: Soft delete ID extraction via `privacy/utils`; include mutation fields as event properties.
> - **Listeners**:
>   - **Entitlements**: Split, route by operation (create/delete/update); add reconciliation helpers and context elevation; robust ID/property extraction.
>   - **Slack**: Rework to use `EventContext` and template loader; config setter preserved.
> - **Reconciler**:
>   - Seed default trial prices when absent; only update org Stripe ID when changed; ensure prices before subscription creation.
> - **Docs/Tests**:
>   - Rewrite README with typed topics and EventContext; update/expand unit tests across pool, topics, options, Redis store; minor test typo fix ("behavior").
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 01287553eaeeef10f80ee2b34f148dedd599849c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -104,7 +104,7 @@ require (
 	github.com/tmc/langchaingo v0.1.14
 	github.com/urfave/cli/v3 v3.5.0
 	github.com/vektah/gqlparser/v2 v2.5.31
-	github.com/windmill-labs/windmill-go-client v1.573.1
+	github.com/windmill-labs/windmill-go-client v1.574.0
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/zitadel/oidc/v3 v3.45.0
 	gocloud.dev v0.43.0
--- a/go.sum
+++ b/go.sum
@@ -738,8 +738,8 @@ github.com/vmihailenco/msgpack/v5 v5.4.1 h1:cQriyiUvjTwOHg8QZaPihLWeRAAVoCpE00IU
 github.com/vmihailenco/msgpack/v5 v5.4.1/go.mod h1:GaZTsDaehaPpQVyxrf5mtQlH+pc21PIudVV/E3rRQok=
 github.com/vmihailenco/tagparser/v2 v2.0.0 h1:y09buUbR+b5aycVFQs/g70pqKVZNBmxwAhO7/IwNM9g=
 github.com/vmihailenco/tagparser/v2 v2.0.0/go.mod h1:Wri+At7QHww0WTrCBeu4J6bNtoV6mEfg5OIWRZA9qds=
-github.com/windmill-labs/windmill-go-client v1.573.1 h1:IUft+CcLTBj79igkkB4p36KqxXtxKxQbcf6JzuWrDnM=
-github.com/windmill-labs/windmill-go-client v1.573.1/go.mod h1:RjnzKr3lc7/Gr72zXDmmqtnUm6PzU+UVHhd3alNKxoM=
+github.com/windmill-labs/windmill-go-client v1.574.0 h1:GUJrre/KV11jLPUkhc1PqxgKH0a8N98HC7Qiyt56yxo=
+github.com/windmill-labs/windmill-go-client v1.574.0/go.mod h1:RjnzKr3lc7/Gr72zXDmmqtnUm6PzU+UVHhd3alNKxoM=
 github.com/wk8/go-ordered-map/v2 v2.1.8 h1:5h/BUHu93oj4gIdvHHHGsScSTMijfx5PeYkE/fJgbpc=
 github.com/wk8/go-ordered-map/v2 v2.1.8/go.mod h1:5nJHM5DyteebpVlHnWMV0rPz6Zp7+xBAnxjb1X5vnTw=
 github.com/woodsbury/decimal128 v1.4.0 h1:xJATj7lLu4f2oObouMt2tgGiElE5gO6mSWUjQsBgUlc=


internal/ent/hooks/event.go
```diff
@@ -160,8 +33,7 @@ func EmitEventHook(e *Eventer) ent.Hook {
 			// Delete operations return an int of the number of rows deleted
 			// so we do not want to skip emitting events for those operations
 			if op != SoftDeleteOne && reflect.TypeOf(retVal).Kind() == reflect.Int {
-				logx.FromContext(ctx).Debug().Interface("value", retVal).Msgf("mutation of type %s returned an int, skipping event emission", op)
-				// TODO: determine if we need to emit events for mutations that return an int
+				zerolog.Ctx(ctx).Debug().Interface("value", retVal).Msgf("mutation of type %s returned an int, skipping event emission", op)
```
golanglemonade: looks like you might have unintentionally reverted you changes around switching to `logx.FromContext` here and other places further down

internal/ent/hooks/event.go
```diff
@@ -226,11 +116,48 @@ func EmitEventHook(e *Eventer) ent.Hook {
...
+		log.Err(err).Msg("Failed to unmarshal return value")
+		return nil, err
+	}
+
+	return &event, nil
+}
+
+// parseSoftDeleteEventID extracts the EventID from a soft delete mutation
+func parseSoftDeleteEventID(mutation ent.Mutation) (*EventID, error) {
+	m, ok := mutation.(*entgen.OrganizationMutation)
```
golanglemonade: why is this specific to the organization mutation? Should this use `utils.GenericMutation` so it would work for any mutation kind?

internal/ent/hooks/event.go
```diff
@@ -251,484 +178,79 @@ func getOperation(ctx context.Context, mutation ent.Mutation) string {
...
+			return false
 		}
 
-		return false
+		entity := m.Type()
+		if entity == "" {
+			return false
+		}
+
+		// Prefer the live pool state so dynamically registered listeners are honoured even when they bypass
```
golanglemonade, line 199:
```suggestion
		// Prefer the live pool state so dynamically registered listeners are honored even when they bypass
```

internal/ent/hooks/event.go
```diff
@@ -251,484 +178,79 @@ func getOperation(ctx context.Context, mutation ent.Mutation) string {
...
+			return true
+		}
+
+		if e.listeners == nil {
+			return false
+		}
+
+		// Listener registration drives emission: if no subscribers, we avoid creating events altogether
+		// This mirrors the old static allowlist, but removes the need to keep two separate sources of truth in sync
+		_ = ctx
```
golanglemonade: what is this for?

matoszz: you

internal/ent/hooks/event.go
matoszz: - updated so it casts the mutation to utils.GenericMutation, calls getMutationIDs, and emits the first ID (warns if multiple). That replaces the hard-coded `entgen.OrganizationMutation` path, so any mutation implementing `GenericMutation` works
- updated to pass the request context into `parseSoftDeleteEventID`; the emit check no longer relies on organization-specific state
- updated the hook signature to ignore the context parameter cleanly

internal/entitlements/reconciler/reconciler.go
```diff
@@ -191,8 +191,13 @@ func (r *Reconciler) reconcileOrg(ctx context.Context, org *ent.Organization) er
 
 	if cust.Prices == nil && r.db != nil && r.db.EntConfig != nil {
 		// Mutations that create orgs no longer populate prices inline; seed the defaults here so
-		// the reconciler mirrors the legacy PopulatePricesForOrganizationCustomer behaviour.
-		cust.Prices = internalentitlements.TrialMonthlyPrices(r.db.EntConfig.Modules.UseSandbox)
+		// the reconciler mirrors the legacy PopulatePricesForOrganizationCustomer behavior.
+		useSandbox := false
+
+		modules := &r.db.EntConfig.Modules
+		useSandbox = modules.UseSandbox
```
golanglemonade: there's no reason to set a default `useSandbox` here because it is immediately updated:

```suggestion
		// the reconciler mirrors the legacy PopulatePricesForOrganizationCustomer behavior.
		modules := &r.db.EntConfig.Modules
		useSandbox := modules.UseSandbox
```

PR #1536, opened 2025-11-09
feat: add policy + procedure update with files

golanglemonade:
- optionally allows a file for policy + procedure update, which will enable of syncing documents from other systems. Adds the query to the cli
- on update, it will only set the details and not the name of the document
- changes revision bump log to `debug` instead of `error`


internal/ent/hooks/documents.go
```diff
@@ -108,7 +108,8 @@ func HookImportDocument() ent.Hook {
 				}
 
 				// Parse the uploaded file and write values into the mutation
-				if err := importFileToSchema(ctx, mut); err != nil {
+				isUpdate := mut.Op() != ent.OpCreate
```
adelowo, line 111: Maybe we should do the check for soft delete since it's ent.OpUpdate

adelowo (approved):

PR #1540, opened 2025-11-09
river job to create pirsch domain

acookin:
Create a job that can create a pirsch domain and associate it with a trust center.

This implementation requires there to be a custom domain to create the pirsch domain.

I'll be hooking this in to the trust center create/update hooks.


internal/graphapi/resolver.go
```diff
@@ -67,18 +67,19 @@ type Resolver struct {
 // NewResolver returns a resolver configured with the given ent client
 func NewResolver(db *ent.Client, u *objects.Service) *Resolver {
 	return &Resolver{
-		db:                  db,
```
acookin, line 70: i think this was just my go formatter

golanglemonade: yeah, I think it was more that last commit on this didn't use the right go fmt and yours fixed it :-D

PR #1549, opened 2025-11-11
Add Internal Notification Model and Migrations

manab-pr:
```
Introduce the Notification ent schema with validation, mixins, and org/user edges while keeping it out of GraphQL exposure.
Define Notification channel/type enums 
add the notifications table .
```


manab-pr:
@golanglemonade   i have added clean_notification.go which runs right after Ent codegen. Rather than forking Ent, this go:generate helper scans graphapi/schema/ent.graphql and strips every notification type, connection, input field, and query piece that Ent just wrote:
 - It watches for docstrings + type definitions like type NotificationConnection, NotificationEdge, etc., and skips those blocks entirely.
  - It removes mutation inputs (addNotificationIDs, etc.) and query fields (notifications(...)) so clients never see them.
  - It rewrites the file in-place once scanning finishes. lemme know if it's the right approach .

golanglemonade:
> @golanglemonade i have added clean_notification.go which runs right after Ent codegen. Rather than forking Ent, this go:generate helper scans graphapi/schema/ent.graphql and strips every notification type, connection, input field, and query piece that Ent just wrote:
> 
> * It watches for docstrings + type definitions like type NotificationConnection, NotificationEdge, etc., and skips those blocks entirely.
> * It removes mutation inputs (addNotificationIDs, etc.) and query fields (notifications(...)) so clients never see them.
> * It rewrites the file in-place once scanning finishes. lemme know if it's the right approach .

Hm, that shouldn't be needed at all with the skip code. I had run generation on my local with the schema changes I suggested and didn't get anything that wasn't needed and had a clean generate. Can you remove that and we can see what is generated?

db/migrations-goose-postgres/20251110154531_notifications-schema.sql
```diff
@@ -0,0 +1,27 @@
+-- +goose Up
+-- modify "invites" table
+ALTER TABLE "invites" ADD COLUMN "ownership_transfer" boolean NULL DEFAULT false;
+-- modify "trust_center_watermark_config_history" table
+ALTER TABLE "trust_center_watermark_config_history" ALTER COLUMN "font" SET DEFAULT 'HELVETICA';
+-- modify "trust_center_watermark_configs" table
+ALTER TABLE "trust_center_watermark_configs" ALTER COLUMN "font" SET DEFAULT 'HELVETICA';
```
golanglemonade, line 7: it looks like you might be not rebased off main. These are already in a migration from 1030

```
-- Modify "invites" table
ALTER TABLE "invites" ADD COLUMN "ownership_transfer" boolean NULL DEFAULT false;

```

internal/ent/schema/user.go
```diff
@@ -257,6 +257,14 @@ func (u User) Edges() []ent.Edge {
 		defaultEdgeToWithPagination(u, Event{}),
 		defaultEdgeToWithPagination(u, ActionPlan{}),
 		defaultEdgeToWithPagination(u, Subcontrol{}),
+		edgeToWithPagination(&edgeDefinition{
+			fromSchema: u,
+			edgeSchema: Notification{},
```
golanglemonade, line 262: Can you add `cascadeDelete: "User",`
```suggestion
			edgeSchema: Notification{},
			cascadeDelete: "User",
```

manab-pr:
> > @golanglemonade i have added clean_notification.go which runs right after Ent codegen. Rather than forking Ent, this go:generate helper scans graphapi/schema/ent.graphql and strips every notification type, connection, input field, and query piece that Ent just wrote:
> > 
> > * It watches for docstrings + type definitions like type NotificationConnection, NotificationEdge, etc., and skips those blocks entirely.
> > * It removes mutation inputs (addNotificationIDs, etc.) and query fields (notifications(...)) so clients never see them.
> > * It rewrites the file in-place once scanning finishes. lemme know if it's the right approach .
> 
> Hm, that shouldn't be needed at all with the skip code. I had run generation on my local with the schema changes I suggested and didn't get anything that wasn't needed and had a clean generate. Can you remove that and we can see what is generated?

actually i was kind of middle of it , where i was already trying different ways to bypass the generation , and it got messy , lemme just try with a clean branch ,and as it seems this branch got mixed up with previous commits as well , so lemme just create a fresh PR tomorrow.

golanglemonade:
sounds good @manab-pr. One strategy I use when making a new schema is commit the actual schema locally before running generate so I can easily back out/revert the generated code when it's inevitably wrong the first pass :)

Issue #1526, opened 2025-11-08
[Feature Request] New notifications schema


golanglemonade:
In order to best support generic notifications to the UI that will enabled both offline and online updates, we want to create a new table called `notifications`. 

It should have the following fields:
```
- user_id:  fk to `user` table (users can have many notifications, notifications can only have one user) - optional 
- notificationType: organization, user (enum field - see `pkg/enums` for other enum examples
- objectType: this should be strcase.LowerCamelCase(m.Type()) + strcase.UpperCamelCase(m.Op()) (e.g. taskUpdate, internalPolicyUpdate, etc)
- title:  string field
- body: text field
- data: JSON (structured payload (IDs, links))
- read_at: Time, `GoType(models.DateTime{}).`
- channels: []string{} with enum values - `IN_APP`, `SLACK`, `EMAIL`
```

`newOrgOwnedMixin`:

```
- owner_id: fk to `organization` table (orgs can have many notifications, notifications can only have 1 org) required - this should use the `newOrgOwnedMixin`  - this field doesn't need to be added manually to the `Fields` as the mixin adds it automatically
```

This issue is just to create the schema so it can be merged in without a lot of merge conflicts and then this can be used with the new `subscriptions`

Notes:
- use `task db:newschema -- Notification` - to create the base templated schema
- add the fields + edges first to do the first round of generation, the id, created at, updated at fields will be automatically added 
- use `task cli:generate:enum` to create the enum base files
- use `task generate` (or `task regenerate`, task generate:ent (can be used to just start with the ent stide), `task generate:gql` (to be used to just generate the graphql side)
- use `task db:create` after you have everything in place - this will create the db migrations but usually best to do last so you only have 1 and easy to fix merge conflicts 
- We don't want history tables here so go ahead and add this annotation _before_ you first generate, along with the skip for query + schema gen - I don't think we want this for subscriptions but we can always add them later if so 

```
// Annotations of the Notification
func (Notification) Annotations() []schema.Annotation {
	return []schema.Annotation{
		history.Annotations{
			Exclude:   true,
		},
                 // skip schema gen, this is used for subscriptions only
		entx.SchemaGenSkip(true),
                // skip query gen, this is used for subscriptions only
		entx.QueryGenSkip(true),
	}
}
```

- On the policies, the only rule we want is `rule.AllowIfContextAllowRule(),` - this will allow requests from inside the server to add notifications but not other users
- We want an index on `user_id`, `read_at` , `owner_id`
[...]

PR #1560, opened 2025-11-12
Add internal notifications system foundation with Notification entity schema and database migrations.

manab-pr:
This PR implements the core database schema for an internal notifications system that will allow users to receive in-app notifications about various events and actions within the platform.

resolves: https://github.com/theopenlane/core/issues/1526

### DIFF
--- a/internal/ent/hooks/errors.go
+++ b/internal/ent/hooks/errors.go
@@ -89,6 +89,8 @@ var (
 	ErrNotSingularTrustCenter = errors.New("you can only create/manage one trust center at a time")
 	// ErrStatusApprovedNotAllowed is returned when a user attempts to set status to APPROVED without being in the approver or delegate group
 	ErrStatusApprovedNotAllowed = errors.New("you must be in the approver group to mark as approved")
+	// ErrInvalidChannel is returned when an invalid notification channel is provided
+	ErrInvalidChannel = errors.New("invalid channel")
 )
 
 // IsUniqueConstraintError reports if the error resulted from a DB uniqueness constraint violation.
--- a/internal/ent/hooks/notification.go
+++ b/internal/ent/hooks/notification.go
@@ -56,7 +56,7 @@ func isValidChannels(channels []enums.Channel) error {
 
 	for _, ch := range channels {
 		if !validMap[string(ch)] {
-			return fmt.Errorf("invalid channel: %s", ch)
+			return fmt.Errorf("%w: %s", ErrInvalidChannel, ch)
 		}
 	}
 


.task/checksum/generate-catalog-smart
golanglemonade, line 1: @manab-pr can you restore these files?

internal/ent/schema/invite.go
```diff
@@ -88,7 +88,6 @@ func (Invite) Fields() []ent.Field {
 			Default(string(enums.InvitationSent)),
 		field.Enum("role").
 			GoType(enums.Role("")).
-			Values(string(enums.RoleOwner)).
```
golanglemonade, line 91: This should not be removed either

internal/ent/schema/notification.go
```diff
@@ -0,0 +1,138 @@
...
+			Optional(),
+		field.Time("read_at").
+			Comment("the time the notification was read").
+			GoType(models.DateTime{}).
+			Optional().
+			Nillable(),
+		field.JSON("channels", []enums.Channel{}).
+			Comment("the channels this notification should be sent to (IN_APP, SLACK, EMAIL)").
+			Optional(),
+	}
```
golanglemonade, line 72: Your previous branch had a validator here, can you add that back?

internal/ent/schema/invite.go
manab-pr: @golanglemonade  okay, actually there's an issue in here , Values(string(enums.RoleOwner)). it has this and in default it has Default(string(enums.RoleMember)), how that was a problem , and it's not right as well , so what do i do then ? in values i have to add all three roles then. otherwise it'll just throw an error in `task generate`

.task/checksum/generate-catalog-smart
manab-pr: okay.

internal/ent/schema/notification.go
manab-pr: okay i'll add that back in here.

internal/ent/schema/orgmembership.go
```diff
@@ -48,7 +48,6 @@ func (OrgMembership) Fields() []ent.Field {
 	return []ent.Field{
 		field.Enum("role").
 			GoType(enums.Role("")).
-			Values(string(enums.RoleOwner)).
```
golanglemonade, line 51: you don't want this remove either; this adds the role owner to the accepted values

internal/ent/schema/invite.go
golanglemonade: I'm not sure I understand how this schema is related to your change at all? This was added by you on a previous PR: https://github.com/theopenlane/core/blame/main/internal/ent/schema/invite.go#L91 and is required for the ownership transfer logic. Values adds it the allowed values; which is different from the default of owner. 

I'm wondering if your fork is out of date and causing you issues? You should be able to:
1. rebase your fork off main
2. grab your schema, enum changes
3. run generate - generate ent updates 
4. run task db:create -  generate migrations

internal/ent/schema/notification.go
```diff
@@ -0,0 +1,169 @@
...
+func (Notification) PluralName() string {
+	return pluralize.NewClient().Plural(SchemaNotification)
+}
+
+// Fields of the Notification
+func (Notification) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("user_id").
+			Comment("the user this notification is for").
+			NotEmpty(),
```
golanglemonade: We want this to be Optional so a notification could be org level, instead of user specific 

```suggestion
		field.String("user_id").
			Comment("the user this notification is for").
			Optional(),
```

internal/ent/schema/notification.go
```diff
@@ -0,0 +1,169 @@
...
+						for _, ch := range chSlice {
+							if !validMap[string(ch)] {
+								return nil, fmt.Errorf("invalid channel: %s", ch)
+							}
+						}
+					}
+				}
+				return next.Mutate(ctx, m)
+			})
+		},
```
golanglemonade: Curious why this was a validator on the field before and now a hook? 

If it's needs hook. can you pull this out into a function in the `hooks/` directory; add a `notification.go` and add it there; most hooks are kept there to keep the logic out of here.  You can also then use the generated fields and do `m.Channels()`, `m.AppendChannels()`, to check instead of using `m.Field()`. 

I'm leaning towards this being a validator though which would be simpler? e.g. `func IsValidChannels(channels []string) bool` and just call from the schema; validators are in `internal/ent/validator` and would look something like:

```go
func IsValidChannels(channels []string) bool {
    // base case, nothing here
    if len(channels) == 0 {
        return true
    }
    
    ... rest of logic ...
}
```

And then the validator on the schema would just be:
```
validator.IsValidChannels()
```


These functions already exist in the `privacy/utils` directory but don't believe they could be imported into validator because of import cylces, but can be used as a reference: 

```
// SliceToMap converts a slice of strings to a map for faster lookups
func SliceToMap(s []string) map[string]any {
	m := make(map[string]any)
	for _, t := range s {
		m[t] = struct{}{}
	}

	return m
}

// CheckContains checks if any of the elements in `e` are contained in `s`
func CheckContains(s []string, e []string) bool {
	m := SliceToMap(s)

	for _, item := range e {
		if _, exists := m[item]; exists {
			return true
		}
	}

	return false
}
```

internal/ent/schema/notification.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Edges of the Notification
+func (n Notification) Edges() []ent.Edge {
+	return []ent.Edge{
+		uniqueEdgeFrom(&edgeDefinition{
+			fromSchema: n,
+			edgeSchema: User{},
+			field:      "user_id",
+			required:   true,
```
golanglemonade: Don't think this needs to be required, same as comment on the field 
```suggestion
```

internal/ent/schema/notification.go
manab-pr: Honestly as there was no JSON().validation() i used a reference , i can't find it now and i have cleared all the previous branches as well , i couldn't find it , that's y i took this Hook validator path which is used by most of the schema .
and as for the validator  `internal/ent/validator` , how do i use it ? like  : 
```
field.JSON("channels", []enums.Channel{}).
			Comment("the channels this notification should be sent to (IN_APP, SLACK, EMAIL)").
			Optional().Validator(),
``` 
that validator is undefined .
and i guess in here Validator will be better approach though cause it validates the schema directly and as the notification is internal only , so i guess this would be a better approach but i can't find anything to use it in the schema.
and as the hooks is better for logging and mutation , but as per out usecase in here  `internal/ent/validator` this would be better.

golanglemonade: oh interesting, I figured since `strings` is really just a JSON field, `JSON` would work the same but I don't see that function existing. This works.

golanglemonade (approved):

PR #1583, opened 2025-11-15
Notifications schema does not need soft delete

### LINKED ISSUE #1572: Notifications schema does not need soft deletes
1. Update the notifications schema to not include soft deletes; there is not reason to retain notifications that were deleted. Add the following the mixin config: 
```go
excludeSoftDelete: true,
```
1. Run `task generate`
2. run `task db:create`

manab-pr:
Added excludeSoftDelete: true, in mixin config.

Closes #1572

### DIFF
--- a/db/migrations-goose-postgres/20251115142555_notifications-schema.sql
+++ b/db/migrations-goose-postgres/20251115142555_notifications-schema.sql
@@ -1,7 +1,0 @@
--- +goose Up
--- modify "notifications" table
-ALTER TABLE "notifications" DROP CONSTRAINT "notifications_users_notifications", ALTER COLUMN "user_id" DROP NOT NULL, ADD CONSTRAINT "notifications_users_notifications" FOREIGN KEY ("user_id") REFERENCES "users" ("id") ON UPDATE NO ACTION ON DELETE SET NULL;
-
--- +goose Down
--- reverse: modify "notifications" table
-ALTER TABLE "notifications" DROP CONSTRAINT "notifications_users_notifications", ALTER COLUMN "user_id" SET NOT NULL, ADD CONSTRAINT "notifications_users_notifications" FOREIGN KEY ("user_id") REFERENCES "users" ("id") ON UPDATE NO ACTION ON DELETE NO ACTION;
--- a/db/migrations-goose-postgres/20251117155928_notifications-schema.sql
+++ b/db/migrations-goose-postgres/20251117155928_notifications-schema.sql
@@ -4,7 +4,7 @@ DROP INDEX "notification_owner_id";
 -- drop index "notification_user_id_read_at_owner_id" from table: "notifications"
 DROP INDEX "notification_user_id_read_at_owner_id";
 -- modify "notifications" table
-ALTER TABLE "notifications" DROP COLUMN "deleted_at", DROP COLUMN "deleted_by";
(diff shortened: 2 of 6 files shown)


manab-pr (commented):
checked

golanglemonade:
@manab-pr looks like you need to run gofmt on the schema file for the linter

manab-pr:
@golanglemonade  Ahh , i totally forgot to run it beforehand.

internal/ent/schema/notification.go
```diff
@@ -21,8 +21,6 @@ import (
 
 // Notification holds the schema definition for the Notification entity
 type Notification struct {
-	SchemaFuncs
```
golanglemonade, line 24: I don't think you want this removed here

manab-pr: What about 
```
SchemaFuncs is an interface that defines methods (Name(), PluralName(), GetType())
  The Notification struct already implements these methods (lines 29-41), so it doesn't need to embed the interface
```

golanglemonade: Where are you seeing that?

manab-pr: `golangci-lint run --config=.golangci.yaml internal/ent/schema/notification.go `  showed this is an issue , so i searched through and got that , That's y  i removed that.

manab-pr: the **test panics** is resolved with `task generate` .

golanglemonade: Hmmm every other schema has this, so if it's falling CI it's because of a new linter rule 😔 

If it's all passing with generate than that works

manab-pr: okay

manab-pr: and one more thing there's this mockery.yml , which is basically throwing it runs on go1.24  i mean the configuration but the project runs on go1.25 , should i touch that file at all if needed ? or like leave it as it is ?

golanglemonade: Feel free to adjust, if needed. I'm not at a computer but can look tomorrow if you need any help on it

golanglemonade:
I see the issue, its an index that is autocreated; 

```
	{
		Name:    "notification_owner_id",
		Unique:  false,
		Columns: []*schema.Column{NotificationsColumns[13]},
		Annotation: &entsql.IndexAnnotation{
			Where: "deleted_at is NULL",
		},
	},
```

Let me see the best way to fix this

golanglemonade:
> I see the issue, its an index that is autocreated;
> 
> ```
> 	{
> 		Name:    "notification_owner_id",
> 		Unique:  false,
> 		Columns: []*schema.Column{NotificationsColumns[13]},
> 		Annotation: &entsql.IndexAnnotation{
> 			Where: "deleted_at is NULL",
> 		},
> 	},
> ```
> 
> Let me see the best way to fix this

I tested this change against your branch and it removed this bad field: https://github.com/theopenlane/core/pull/1587 so once its merged, you should be able to rebase + generate again

manab-pr:
@golanglemonade  thanks .

manab-pr:
@golanglemonade  I have added those changes that you have made @ #1587 .

db/migrations-goose-postgres/20251115142555_notifications-schema.sql
```diff
@@ -0,0 +1,7 @@
+-- +goose Up
```
golanglemonade: @manab-pr can you combine the two migrations into one; you should be able to just revert all changes in the `db` directory and then run `task db:create` again

manab-pr: okay , on it.

manab-pr: done.

PR #1587, opened 2025-11-16
skip index on soft delete for org owner when no deleted_at

golanglemonade:
when excluding soft deletes on a schema with org owner, it would fail because there was an index auto-created with `deleted_at` but the column doesn't exists. This will auto-set a flag to skip the index when the combination exists on the schema


golanglemonade:
pulled into https://github.com/theopenlane/core/pull/1583 instead

PR #1588, opened 2025-11-16
cache trust center data

acookin:
New river job to cache the trust center data in storage.

Initializing the worker https://github.com/theopenlane/riverboat/pull/176

importing the internal `resolvers` package was causing a cycle in the dependency graph, so we defer the initialization of the object storage to the caller. Moving the resolver package to a public place so that the riverboat repo can use it.


golanglemonade:
@acookin I think we should close this out after our discussion on Monday; the UI will handle the caching; the only thing we might want in a job here is to clear the cache when changes are published for the trust center.

PR #1594, opened 2025-11-18
feat : Update notifications schema fields, add topic field

### LINKED ISSUE #1574: Update notifications schema fields, add topic field
The notifications that happen should be defined in the format:
`object_type.action`

Few examples of this could be:

`task.assigned`
`task.completed`
`internal_policy.needs_approval`
`control.needs_approval`
`control.changes_requested`
`export.ready`
`export.failed`
`comment.tagged`
`comment.reply`
...etc...

Add a field to the notification schema called `topic` that we can store these values, and eventually allow users to subscribe/unsubscribe to certain types. 

Also update the comment on the `object_type` field to be the following so we can use the `topic` from the event here directly: 
```
("the event type this notification is related to (e.g., task.created, control.updated)").
```

manab-pr:
Updated notifications schema fields, add topic field
and replaced comment in the object_type field.
Closes #1574

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -104,7 +104,7 @@ require (
 	github.com/tmc/langchaingo v0.1.14
 	github.com/urfave/cli/v3 v3.6.1
 	github.com/vektah/gqlparser/v2 v2.5.31
-	github.com/windmill-labs/windmill-go-client v1.576.3
+	github.com/windmill-labs/windmill-go-client v1.578.0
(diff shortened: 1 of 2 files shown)


golanglemonade (commented):
@manab-pr you need to generate the migrations using `task db:create`

manab-pr:
@golanglemonade  done

PR #1600, opened 2025-11-19
feat: trust center preview schema changes

golanglemonade:
- adds environment enum to settings
- adds preview domain 
- allows trust centers to have 1 preview domain and one custom domain
- when publishing settings it will copy settings from preview -> live


acookin (dismissed):
well, that's a lot easier if you only need to support one haha

PR #1601, opened 2025-11-19
feat: added handlers for notifications events

### LINKED ISSUE #1573: Handlers for Notifications Events
The general idea is that objects might be created/updated that a user might want to know about or take action, so we can create notifications based on these 

Add a package `internal/ent/notifications` where we can store the code for the logic, this will keep the packages small and it can be called from `internal/ent/hooks` for the event. 

The scope of this ticket is to just create a few of the types to ensure it works and we can come back later to define all the different paths

The emitted event, will have all the fields as properties:

https://github.com/theopenlane/core/blob/daaeb5b788bf47ba3a5f26f5eb3bcd2a7c276ecb/internal/ent/hooks/event.go#L251-L262

Walking through an example of tasks with psuedo code:

```
// handleTaskAssigned processes tasks mutations and creates a notification record based on the task details
func handleTaskMutation(ctx *soiree.EventContext, payload *MutationPayload) error {
	// ignore deleted operations for tasks
	if payload.Operation == ent.OpDelete.String() {
		return nil
	}

	if assigneeID, ok := ctx.Properties().GetKey("assignee_id") {
		if err := addTaskAssigneeNotification(ctx, payload, assigneeID); err != nil {
			// log error 
			return err
		}
        }

       // other types 
       
       return nil
}

type notificationData struct {
      title string
      body string
      channels []string
      userIDs []string
      data string
      ownerID string
}

const (
      topicTaskAssigned = "task.assigned"
      topicExportCompleted = "export.completed"
...
}



func addTaskAssigneeNotification(ctx *soiree.EventContext, payload *MutationPayload, userID string) error {
      // create notificationData based on fields + payload
      data := notificationData{
            topic: topicTaskAssigned,
            channels: ...
            // in this case, the assignee is the userID to send, in other cases where the object is a group,
            we would need to get all users in the group and add a notification for each
            userIDs: []string{assigneeID}
      }


      return newNotificationCreation(ctx, payload, data)
}

func newNotificationCreation(ctx *soiree.EventContext, payload *MutationPayload, data notificationData) error {
	client := mutationClient(event, payload)
	if client == nil  {
		return nil, false
	}

	// set allow context
	allowCtx := allow(ctx.Context())

	mut := client.Notification.Create()

	// always set the ownerID, body, title, topic, object type
	// organization id the object was created under, on most mutations this the `OwnerID` field 
	if data.OwnerID == "" {
		// if the specific mutation didn't set it, we should attempt to grab it from the object, checking for the OwnerID on the payload, falling back to querying the database for the object to get the ownerID from the object itself
		mut.SetOwnerID() 
	} 

	mut.SetBody(data.body) 
	mut.SetTitle(data.title) 
	mut.SetTopic(data.topic)
	mut.SetObjectType(ctx.Event().Topic())

 	if data.Data != "" {
	 	mut.SetData(data.data) // optional, depending on notification
 	}

 	if len(data.Channels) > 0 {
	 	mut.SetChannels(data.channels) 
 	}

 	//  create notification per user that it should be sent to
 	for _, userID := range userIDs {
 	 	mut.SetUserID(userID)

  	 	if err :=  mut.Exec(allowCtx); err != nil {
  		 	// log error
  		 	return err
  	 	}
   	}

	return nil
```

manab-pr:
--added internal/ent/notifications package
-- called the handlers from internal/ent/hooks/eventer.go 
--lint checked
-- generated
--tested

Closes  #1573


internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    // Get assignee_id from properties
+    assigneeID := props.GetKey("assignee_id")
+    if assigneeID == nil {
+        return nil
+    }
+
+    newAssigneeID := fmt.Sprintf("%v", assigneeID)
+    if !taskAssigneeChanged(ctx, newAssigneeID) {
```
golanglemonade: I don't think you need this, the event only adds fields that were change: 

https://github.com/theopenlane/core/blob/f6797cbe768fcc2571d7093fd79cd6b04a7ef559/internal/ent/hooks/event.go#L254-L264

`Fields()`
> 	// Fields returns all fields that were changed during
	// this mutation. Note that, in order to get all numeric
	// fields that were in/decremented, call AddedFields().

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+        return nil
+    }
+
+    title := props.GetKey("title")
+    entityID := props.GetKey("ID")
+    ownerID := props.GetKey("owner_id")
+
+    // Get assigner info (the user who made the change) from context
+    assignerID := "system"
```
golanglemonade: `assignerID` must be a userID so this would fail, leave the `assigner` empty and it will be set to the system user as a fallback

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    // Get status from properties
+    status := props.GetKey("status")
+    if status == nil {
+        return nil
+    }
+
+    // Check if status is NEEDS_APPROVAL
+    if fmt.Sprintf("%v", status) != "NEEDS_APPROVAL" {
```
golanglemonade: use the named enum, https://github.com/theopenlane/core/blob/main/pkg/enums/documentstatus.go#L18, e.g. something like: 

```
statusEnum := enums.ToDocumentStatus(status.(string))

if status != enums.DocumentNeedsApproval {
...
```

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+        return nil
+    }
+
+    // TODO: Check if status was changed (compare with old value)
```
golanglemonade: this isn't needed since the event provides only changed fields

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+    approverID := props.GetKey("approver_id")
+    if approverID == nil {
+        logx.FromContext(ctx.Context()).Warn().Msg("approver_id not set for internal policy with NEEDS_APPROVAL status")
+        return nil
+    }
+
+    name := props.GetKey("name")
+    entityID := props.GetKey("ID")
+    ownerID := props.GetKey("owner_id")
+
```
golanglemonade: Instead of using fmt.Sprintf all over, I would just cast to a string:
e.g.
```
    name := props.GetKey("name").(string)
```

Additionally, to ensure if fields change this will get updated we can use the typed field names
```
    name := props.GetKey(internalpolicy.FieldName).(string)
```

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+        return err
+    }
+
+    return nil
+}
+
+func addTaskAssigneeNotification(ctx *soiree.EventContext, assigneeID, taskTitle, taskID, ownerID, assignerID string) error {
+    // create the data map with the URL
+    dataMap := map[string]string{
+        "url": fmt.Sprintf("https://console.theopenlane.io/tasks?id=%s", taskID),
```
golanglemonade: the domain needs to be a config variable instead of hard coded. I would add `Notifications` config to the exisiting entconfig: https://github.com/theopenlane/core/blob/main/internal/ent/entconfig/config.go 

// Notifications settings for notifications sent to users based on events
type Notifications struct {
    // ConsoleURL for ui links used in notifications
	ConsoleURL string `koanf:"consoleurl" json:"consoleurl" default:"http://localhost:3001" domain:"inherit" domainPrefix:"https://console"`
}

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    dataJSON, err := json.Marshal(dataMap)
+    if err != nil {
+        logx.FromContext(ctx.Context()).Error().Err(err).Msg("failed to marshal notification data")
+        return err
+    }
+
+    data := notificationData{
+        notificationType: "USER",
```
golanglemonade: you should use the defined enums here

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+        logx.FromContext(ctx.Context()).Error().Err(err).Str("group_id", approverID).Msg("failed to get approver group")
+        return err
+    }
+
+    // get all users in the approver group
+    users, err := group.QueryUsers().All(allowCtx)
+    if err != nil {
+        logx.FromContext(ctx.Context()).Error().Err(err).Str("group_id", approverID).Msg("failed to query users in approver group")
+        return err
+    }
```
golanglemonade: You can query group members driectly
```suggestion
    users, err := client.GroupMembership.Query().Where(groupmembership.GroupID(approverID)).All(allowCtx)
    if err != nil {
        logx.FromContext(ctx.Context()).Error().Err(err).Str("group_id", approverID).Msg("failed to get approver group m embers")
        
        return err
    }
```

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+    if iface == nil {
+        return nil
+    }
+
+    taskMutation, _ := iface.(*generated.TaskMutation)
+    return taskMutation
+}
+
+// HandleInternalPolicyMutation processes internal policy mutations and creates notifications when status = NEEDS_APPROVAL
+func HandleInternalPolicyMutation(ctx *soiree.EventContext) error {
```
golanglemonade: I dont think the `HandleX` functions need to be exported

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+        "url": fmt.Sprintf("https://console.theopenlane.io/policies/%s/view", policyID),
+    }
+
+    dataJSON, err := json.Marshal(dataMap)
+    if err != nil {
+        logx.FromContext(ctx.Context()).Error().Err(err).Msg("failed to marshal notification data")
+        return err
+    }
+
+    data := notificationData{
```
golanglemonade: We need to set the `topic` and `objectType` here still I think

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,326 @@
...
+        title:            "Policy approval required",
+        body:             fmt.Sprintf("%s needs approval, internalPolicy", policyName),
+        data:             string(dataJSON),
+        ownerID:          ownerID,
+    }
+
+    return newNotificationCreation(ctx, data, "InternalPolicy")
+}
+
+func newNotificationCreation(ctx *soiree.EventContext, data notificationData, objectType string) error {
```
golanglemonade: I think instead of using a custom `notificationData` and then having to set each field, we should get the `Input` type generated. I found the issue we ran into with some of that generation last night - so leave this as is for now and I can get that fix in and we can use the generated type here instead which would simply the double setting of data.

internal/ent/notifications/events.go
manab-pr: Done

internal/ent/notifications/events.go
manab-pr: handled.

internal/ent/notifications/events.go
manab-pr: done

internal/ent/notifications/events.go
manab-pr: done

internal/ent/notifications/events.go
manab-pr: yes, added.

internal/ent/notifications/events.go
manab-pr: yes , that would be appropriate .

internal/ent/entconfig/config.go
```diff
@@ -23,6 +23,8 @@ type Config struct {
 	EmailValidation validator.EmailVerificationConfig `json:"emailvalidation" koanf:"emailvalidation"`
 	// Billing contains configuration for billing related features
 	Billing Billing `json:"billing" koanf:"billing"`
+	// Notifications contains configuration for notifications sent to users based on events
```
golanglemonade, line 26: fyi - you will need to run `task config:generate` for these changes

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,268 @@
...
+
+	statusEnum := enums.ToDocumentStatus(status.(string))
+
+	// Check if status is NEEDS_APPROVAL
+	if statusEnum != &enums.DocumentNeedsApproval {
+		return nil
+	}
+
+	// Get approver_id from properties
+	approverID := props.GetKey(internalpolicy.FieldApproverID)
```
golanglemonade: It's possible the approver isn't updated so in the same mutation the status is set. you want to 

1. check the `MutationPayload` first, and if its not there, 
2. query the database for the internal policy

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,268 @@
...
+	if approverID == nil {
+		logx.FromContext(ctx.Context()).Warn().Msg("approver_id not set for internal policy with NEEDS_APPROVAL status")
+		return nil
+	}
+
+	name := props.GetKey(internalpolicy.FieldName).(string)
+	entityID := props.GetKey(internalpolicy.FieldID).(string)
+	ownerID := props.GetKey(internalpolicy.FieldOwnerID).(string)
```
golanglemonade: Same with all these, the only thing you want to check on the props is "if that field changed, triggering the notification". Everything else should be payload -> still missing data -> query database to be sure

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,268 @@
...
+	// Get assignee_id from properties
+	assigneeID := props.GetKey(task.FieldAssigneeID)
+	if assigneeID == nil {
+		return nil
+	}
+
+	title := props.GetKey(task.FieldTitle)
+	entityID := props.GetKey(task.FieldID)
+	ownerID := props.GetKey(task.FieldOwnerID)
+
```
golanglemonade: these should all come from the payload -> fall back to database query since the props are only _changed_ fields.

manab-pr:
@golanglemonade  okay lemme know .

golanglemonade:
> @golanglemonade okay lemme know .

You'll want to update the ent schema:

```
-		entgql.Skip(entgql.SkipAll),
+		entgql.Skip(entgql.SkipMutationUpdateInput | entgql.SkipWhereInput | entgql.SkipOrderField),
+		entgql.Mutations(entgql.MutationCreate()), // only allow creation via mutations, updates are not supported
```

And add the following to `internal/graphapi/schema/scalars.graphql`

```
"""
Channel notifications will be sent to including in-app, slack, etc
"""
scalar Channel
```

You'll want to leave the reset of the skips on the ent schema as-is. This will generate the `CreateNotificationInput` which you can use in the mutation instead of doing a `Set<Field>` for each. 

I put it on a branch [here](https://github.com/theopenlane/core/pull/1632/files) - you can just cherry-pick the commit on your branch if you want.

internal/ent/notifications/events.go
manab-pr: done

internal/ent/notifications/events.go
manab-pr: done

manab-pr:
@golanglemonade  The problem is that we added Channel as a scalar in scalars.graphql, but channels field in the notification schema is already defined as []enums.Channel (an array of enum values), not a custom scalar. This is causing type incompatibility. what do i do in here then ?

golanglemonade:
> @golanglemonade The problem is that we added Channel as a scalar in scalars.graphql, but channels field in the notification schema is already defined as []enums.Channel (an array of enum values), not a custom scalar. This is causing type incompatibility. what do i do in here then ?

Where are you seeing the issue? On my branch I defined the resolvers: https://github.com/theopenlane/core/pull/1632/files#diff-137f35f58e3dad10b0645a05bd5c79be9f11a0d6a63a81c2aae205f9d06d53a4R17-R4750

manab-pr:
> > @golanglemonade The problem is that we added Channel as a scalar in scalars.graphql, but channels field in the notification schema is already defined as []enums.Channel (an array of enum values), not a custom scalar. This is causing type incompatibility. what do i do in here then ?
> 
> Where are you seeing the issue? On my branch I defined the resolvers: https://github.com/theopenlane/core/pull/1632/files#diff-137f35f58e3dad10b0645a05bd5c79be9f11a0d6a63a81c2aae205f9d06d53a4R17-R4750

The issue was i was cherry-picking your commit , but your branch was created with an extra commit in the main branch , i mean before you created that branch , so although generated files had all the necessary codes , but when i cheery picked that particular commit was missing that was creating issue.

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,399 @@
...
+	if err := addTaskAssigneeNotification(ctx, assigneeID, title, entityID, ownerID); err != nil {
+		logx.FromContext(ctx.Context()).Error().Err(err).Msg("failed to add task assignee notification")
+		return err
+	}
+
+	return nil
+}
+
+// getTaskFields retrieves task fields from props or queries database if missing
+func getTaskFields(ctx *soiree.EventContext, props soiree.Properties) (title, entityID, ownerID string, err error) {
```
golanglemonade: on both the `get` functions, I would recommend create a struct and returning that - that keeps the function signature a lot cleaner + easier to read, e.g. for this one: 
```go
type taskFields struct {
    title string
    entityID string
    ownerID string
}

func getTaskFields(ctx *soiree.EventContext, props soiree.Properties) (*taskFields, error) {
```

internal/ent/hooks/eventer.go
```diff
@@ -52,81 +53,89 @@ type MutationHandler func(*soiree.EventContext, *MutationPayload) error
...
+    e.AddMutationListener(entgen.TypeOrganizationSetting, handleOrganizationSettingMutation)
+    e.AddMutationListener(entgen.TypeSubscriber, handleSubscriberMutation)
+    e.AddMutationListener(entgen.TypeUser, handleUserMutation)
+
+    // Register notification listeners from notifications package
+    notifications.RegisterListeners(func(entityType string, handler func(*soiree.EventContext) error) {
+        // Wrap the soiree-based handler to match the MutationHandler signature expected by AddMutationListener
+        e.AddMutationListener(entityType, func(ctx *soiree.EventContext, _ *MutationPayload) error {
+            return handler(ctx)
+        })
```
golanglemonade: You can check the payload before checking the db, you just need to pass it in herE: 
```suggestion
        e.AddMutationListener(entityType, func(ctx *soiree.EventContext, payload *MutationPayload) error {
            return handler(ctx, payload)
        })
```

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,399 @@
...
+	title            string
+	body             string
+	channels         []string
+	userIDs          []string
+	data             string
+	ownerID          string
+	notificationType string
+	topic            string
+	objectType       string
+}
```
golanglemonade: We should be able to remove this type now and use the `generated.CreateNotificationInput` struct

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,399 @@
...
+		}
+	}
+
+	if ownerVal := props.GetKey(task.FieldOwnerID); ownerVal != nil {
+		if o, ok := ownerVal.(string); ok {
+			ownerID = o
+		}
+	}
+
+	// If any field is missing, query the database
```
golanglemonade: check the payload, and then fall back to the db

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,399 @@
...
+		data:             string(dataJSON),
+		ownerID:          ownerID,
+		topic:            "policy_approval",
+		objectType:       "InternalPolicy",
+	}
+
+	return newNotificationCreation(ctx, data)
+}
+
+func newNotificationCreation(ctx *soiree.EventContext, data notificationData) error {
```
golanglemonade: We can use this instead:

```suggestion
func newNotificationCreation(ctx *soiree.EventContext, data *generated.CreateNotificationInput) error {
```

internal/ent/notifications/events.go
```diff
@@ -0,0 +1,399 @@
...
+			channels := make([]enums.Channel, 0, len(data.channels))
+			for _, ch := range data.channels {
+				if c := enums.ToChannel(ch); c != nil && *c != enums.ChannelInvalid {
+					channels = append(channels, *c)
+				}
+			}
+			if len(channels) > 0 {
+				mut.SetChannels(channels)
+			}
+		}
```
golanglemonade: Switching to the input struct, we can now set these once in the other functions, and now just do: 
```suggestion
		mut := client.Notification.Create().SetInput(data)
```

manab-pr:
@golanglemonade  i have modified things as you have asked and there are certain places like in here : mut := client.Notification.Create().SetInput(*input).SetUserID(userID) i have added extra userID as it was not in createNotification struct and well there are certain places where i have changed accordingly , review it , and let me know . Thank you.

golanglemonade:
> @golanglemonade i have modified things as you have asked and there are certain places like in here : mut := client.Notification.Create().SetInput(*input).SetUserID(userID) i have added extra userID as it was not in createNotification struct and well there are certain places where i have changed accordingly , review it , and let me know . Thank you.

Yep, thats exactly right; changes look good

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/507

### Changes Preview:

manab-pr:
@golanglemonade  sonarqube , lint checked .

golanglemonade (approved):

PR #1602, opened 2025-11-19
Workflow and Directory schemas

matoszz:
Purely the schemas and generated output, plus needed enums and models used for the type definitions


internal/ent/schema/directoryaccount.go
```diff
@@ -0,0 +1,228 @@
...
+
+// Mixin composes shared behaviors for DirectoryAccount.
+func (d DirectoryAccount) Mixin() []ent.Mixin {
+	return mixinConfig{
+		prefix:            "DAC",
+		excludeSoftDelete: true,
+		additionalMixins: []ent.Mixin{
+			newOrgOwnedMixin(d),
+			// TODO: uncomment after first generation
+			// newObjectOwnedMixin[generated.DirectoryAccount](d, withParents(Integration{})),
```
golanglemonade: Do you want this as object owned?

internal/ent/schema/directorygroup.go
```diff
@@ -0,0 +1,212 @@
...
+
+// Mixin composes shared behaviors.
+func (g DirectoryGroup) Mixin() []ent.Mixin {
+	return mixinConfig{
+		prefix:            "DRG",
+		excludeSoftDelete: true,
+		additionalMixins: []ent.Mixin{
+			newOrgOwnedMixin(g),
+			// TODO: uncomment after first generation
+			// newObjectOwnedMixin[generated.DirectoryGroup](g, withParents(Integration{})),
```
golanglemonade: Same here - did you want to uncomment this one?

internal/ent/schema/directorysyncrun.go
```diff
@@ -0,0 +1,144 @@
...
+			Comment("additional provider-specific stats for the run").
+			Optional(),
+	}
+}
+
+// Mixin composes shared behavior.
+func (r DirectorySyncRun) Mixin() []ent.Mixin {
+	return mixinConfig{
+		prefix:      "DSR",
+		excludeTags: true,
```
golanglemonade: does this schema need soft delete? seems like more of a history of runs so we don't really need soft deletes here?

internal/ent/schema/organization.go
```diff
@@ -479,6 +479,56 @@ func (o Organization) Edges() []ent.Edge {
...
+		edgeToWithPagination(&edgeDefinition{
+			fromSchema: o,
+			edgeSchema: WorkflowAssignmentTarget{},
+			comment:    "workflow assignment targets configured for this organization",
+		}),
+		edgeToWithPagination(&edgeDefinition{
+			fromSchema: o,
+			edgeSchema: WorkflowObjectRef{},
+			comment:    "workflow object references for this organization",
+		}),
```
golanglemonade, line 511: any reason these don't have `cascadeDeleteOwner` and the others do?

internal/ent/schema/task.go
```diff
@@ -239,6 +239,12 @@ func (t Task) Edges() []ent.Edge {
 				accessmap.EdgeViewCheck(Evidence{}.Name()),
 			},
 		}),
+		edgeFromWithPagination(&edgeDefinition{
+			fromSchema: t,
+			edgeSchema: WorkflowObjectRef{},
+			name:       "workflow_object_refs",
+			ref:        "task",
```
golanglemonade: I dont think you need these since it would be the default already? 
```suggestion
```

internal/ent/schema/workflowobjectref.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+}
+
+// Policy of the WorkflowObjectRef
+func (WorkflowObjectRef) Policy() ent.Policy {
+	return policy.NewPolicy(
+	// policy.WithMutationRules(
+	// 	policy.CheckOrgWriteAccess(),
+	// ),
+	)
```
golanglemonade: this would mean no one can write to this table; did you mean to comment this out?

internal/ent/schema/directoryaccount.go
matoszz: I'm not 100% until getting further into the dev; i'll remove the commented sections for now

PR #1617, opened 2025-11-20
Establish Object Upload via token; minor "Job" uplift

matoszz:
Working on finalizing some basic external agent / job capabilities in conjunction with recent Integrations + Workflow work. Most functionality is already present, but a few problems when building + testing, mostly surrounding creation of evidence as a result of checks, ability to establish relationship, and some highly specific field names that need generic alternatives or more lax requirements on creation. Additionally, the "registration" flow feels pretty broken (and largely lacking tangible outcomes vs. complexity) so removing / bypassing as a part of this work. 

- API tokens/PATs can upload files without session-derived org context as long as they specify ownerID (PAT) or have a single org (API/JWT).
- File persistence now prefers an already selected org or a single authorized org before querying correlated objects
- Updated CLI command to allow generation of the end-state token without the need for registration to allow for easier agent initialization
- Updated several fields for `Job*` and `ScheduledJob*` to allow for easier creation of adhoc checks / jobs without creating an entirely new schema (wanted to just keep the scheduledJob and not worry so much that it was called "scheduled" even though it could be run ad hoc)
- Added edges from `JobResult` and `JobTemplate` to schemas such as Finding, Evidence, Controls, Standards, etc., to allow for external agent / jobs to create + associate additional objects as a part of the "agent" / job configuration
- Removed / cleaned up several (now) out of date comments, made some field requirements more lax for various permutations of checks / jobs
- Updated FGA model to include more direct roles / path to creating a job runner token, effectively bypassing the "registration construct" (the rationale: if I'm a user, deploying an "agent", i'm crafting my local config it will use and then going to Openlane to somehow generate a token I will feed into the agent, assuming this is the static token and that it will source properly from its environment, similar to get this value. If the value we're feeding is a "registration token" that makes the agent itself responsible then for finding persistence for this actual secret rather than the now destroyed registration token. It largely defeats some of the setup a user would have to do to provide the proper token, and the user shouldn't be expected to take an extra step -> they go to generate a token, just like a PAT / API token, they should be shown that token 1 time in response and then expected to persist it)


internal/objects/store/persistence.go
```diff
@@ -123,13 +123,22 @@ func getOrgOwnerID(ctx context.Context, f pkgobjects.File) (string, error) {
...
 		return "", err
 	}
 
+	if orgIDs, err := auth.GetOrganizationIDsFromContext(ctx); err == nil && len(orgIDs) == 1 {
+		return orgIDs[0], nil
+	}
+
 	if !au.IsSystemAdmin && au.OrganizationID != "" {
 		return au.OrganizationID, nil
 	}
```
golanglemonade: This changes the behavior system admins which I don't we want here. 
```suggestion
	// If the actor is a system admin, prefer deriving the organization from the
	// correlated object rather than using the admin's org from context
	au, err := auth.GetAuthenticatedUserFromContext(ctx)
	if err != nil {
		return "", err
	}

	if !au.IsSystemAdmin {
		if au.OrganizationID != "" {
			return au.OrganizationID, nil
		}
		
		if len(u.OrganizationIDs) == 1 {
			return au.OrganizationIDs[0]
		}
	}
```

fga/model/model.fga
```diff
@@ -116,6 +116,12 @@ type organization
     define evidence_creator: [group#member]
     define can_create_evidence: can_edit or evidence_creator
```
golanglemonade, line 118: these types need to be added here: (yes I know this is annoying and manual still 😅 )

https://github.com/theopenlane/core/blob/main/internal/ent/schema/mixin_createacess.go#L21-L36

matoszz:
I'm going to go a different direction with this and closing for now. Plan is to wait until Trust Center schema changes are all in, and re-establish some new schemas that are integrated with / work in conjunction with the "workflow" capability recently added rather than trying to get the existing "job" + "scheduledjob" constructs to refactor into it.

PR #1637, opened 2025-11-24
updates to tags + enums

adelowo:
- [x] Add icon to enums
- [x] Deleting an enum should check if in use by another schema/object before proceeding
- [x] Deleting a tagdefintion should check if in use by another schema/object before proceeding
- [x] Name on enums and tagdefinition should be immutable 
- [x] before creating a tag, lookup the aliases to see if the value is there, if so, use that instead
- [x] Sluggify name for both tag definitions and enums
- [x] Update the permissions to only allow users with org access to create + update tags 
- [x] removed regex for converting org name to slug for trustcenters


internal/ent/hooks/customenums.go
```diff
@@ -79,3 +93,159 @@ func HookCustomEnums(in CustomEnumFilter) ent.Hook {
 		})
 	}, hook.HasOp(ent.OpCreate|ent.OpUpdate|ent.OpUpdateOne))
 }
+
+// HookCustomTypeEnumDelete checks if the enum(s) being deleted is in use by any other object.
+// If in use, the deletion cannot proceed
+func HookCustomTypeEnumDelete() ent.Hook {
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.CustomTypeEnumFunc(func(ctx context.Context, m *generated.CustomTypeEnumMutation) (generated.Value, error) {
```
golanglemonade, line 98: Did you look into using `entsql.OnDelete(entsql.Restrict),` at all - not sure if that will work with soft deletes

internal/ent/hooks/customenums.go
```diff
@@ -79,3 +93,159 @@ func HookCustomEnums(in CustomEnumFilter) ent.Hook {
...
+func isEnumInUse(ctx context.Context, client *generated.Client, enumID, edgeName, objectType, name string, allErrors *[]string, mu *sync.Mutex) []func() {
+	type countTask struct {
+		queryFunc func(context.Context) (int, error)
+		label     string
+	}
+
+	var operations []countTask
+
+	switch edgeName {
+	case "task":
```
golanglemonade: Instead of having to update this for every edge we go add this to, can we use a query like:

```
table := strcase.SnakeCase(objectType)
field := fmt.Sprintf("%s_%s_id, table, fieldName)
query := "SELECT count(id) FROM " + table + " WHERE " + field + " = $1 and deleted_at is NULL"

	var rows sql.Rows
	if err := generated.FromContext(ctx).Driver().Query(ctx, query, []any{enumID}, &rows); err != nil {
		logx.FromContext(ctx).Error().Err(err).Msg("failed to query enum edges")

		return err
	}

	defer rows.Close()

	if rows.Next() {
....
```

internal/ent/hooks/tagdefinition.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/hook"
+	"github.com/theopenlane/core/internal/ent/generated/tagdefinition"
+	"github.com/theopenlane/core/internal/ent/generated/workflowdefinition"
+	"github.com/theopenlane/core/pkg/logx"
+)
+
+func sluggify(s string) string {
+	return strings.ReplaceAll(
+		strcase.SnakeCase(
```
golanglemonade: You should be able to use `KebabCase` instead which does both the `ToLower` and `-`

internal/ent/hooks/tagdefinition.go
```diff
@@ -0,0 +1,164 @@
...
+			name, ok := m.Name()
+			if !ok {
+				return next.Mutate(ctx, m)
+			}
+
+			tagDefs, err := m.Client().TagDefinition.Query().
+				All(ctx)
+			if err != nil {
+				logx.FromContext(ctx).Warn().Err(err).Msg("error fetching all org tags")
+				return nil, errors.New("an error occurred while fetching all tags") //nolint:err113
```
golanglemonade: Can you define an error instead of no linting this?

internal/ent/hooks/tagdefinition.go
```diff
@@ -0,0 +1,164 @@
...
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.TagDefinitionFunc(func(ctx context.Context, m *generated.TagDefinitionMutation) (generated.Value, error) {
+			name, ok := m.Name()
+			if !ok {
+				return next.Mutate(ctx, m)
+			}
+
+			tagDefs, err := m.Client().TagDefinition.Query().
+				All(ctx)
+			if err != nil {
```
golanglemonade: I would have the query do the work instead of looping through the results
```
tagDefs, err := m.Client().TagDefinition.Query().
    Where(
        tagdefinition.Or(
            tagdefinition.NameEqualFold(name),
            func(s *sql.Selector) {
                s.Where(
                    sqljson.ValueContains(
                        tagdefinition.FieldAliases,
                        name,
                    ),
                )
            },
        ),
    )
    ....
```

This way, if nothing is found -> create. Otherwise do not create.

golanglemonade: We probably want to normalize all tag names + aliases to be lower case before we save as well

internal/ent/hooks/tagdefinition.go
```diff
@@ -0,0 +1,164 @@
...
+
+			return next.Mutate(ctx, m)
+		})
+	}, ent.OpDeleteOne|ent.OpDelete|ent.OpUpdateOne|ent.OpUpdate)
+}
+
+// isTagDefinitionInUse checks if a tag definition is in use by any schema.
+//
+// It is just workflow for now but can always be extended. See hooks/customenums.go
+func isTagDefinitionInUse(ctx context.Context, client *generated.Client, tagDefID, tagName string, allErrors *[]string, mu *sync.Mutex) []func() {
```
golanglemonade, line 163: Same question here; can we make a generic function so this doesn't have to be updated each time?

internal/ent/hooks/tags.go
```diff
@@ -60,6 +66,12 @@ func HookTags() ent.Hook {
 					Where(tagdefinition.NameEqualFold(tag)).
 					Exist(ctx)
 				if !exists {
+					if err := rule.CheckCurrentOrgAccess(ctx, m, fgax.CanEdit); err != nil {
```
golanglemonade: Does this need to be here? I think just changing it on the schema policy (as you already did) should work?

internal/ent/schema/workflowdefinition.go
```diff
@@ -95,6 +98,9 @@ func (WorkflowDefinition) Edges() []ent.Edge {
 			name:       "target_tags",
 			edgeSchema: TagDefinition{},
 			comment:    "Tags this workflow targets for scoping",
+			annotations: []schema.Annotation{
+				accessmap.EdgeNoAuthCheck(),
```
golanglemonade: This seems incorrect - you should have view access to the tag to be able to add it? 

```suggestion
				accessmap.EdgeViewCheck(TagDefinition{}.Name()),
```

internal/ent/hooks/customenums.go
adelowo: Ha. I was conflicted on doing this since this pattern didn't exist elsewhere. 

Will update then

PR #1651, opened 2025-11-26
Trustcenter customer entities

adelowo:
- [x] add a new `customer` entity type 
- [x] add a new TrustcenterEntity schema 
- [x] ( for old orgs that won't have this new `customer` entity type, a hook makes sure it exists before creating the first trustcenterentity 
- [x] tests

### DIFF
--- a/fga/model/model.fga
+++ b/fga/model/model.fga
@@ -131,6 +131,9 @@ type organization
     define trust_center_subprocessor_creator: [group#member]
     define can_create_trust_center_subprocessor: can_edit or trust_center_subprocessor_creator
 
+
+    define can_create_trust_center_entity: can_edit
+
     # additional relations
     define user_in_context: [user]
 
@@ -179,7 +182,7 @@ type file
     define tc_doc_deleter: can_delete from tc_doc_parent
     define tc_doc_parent: [trust_center_doc]
 
-    define parent: [user, program, organization, control, procedure, template, document_data, contact, internal_policy, narrative, evidence, note, trust_center_setting, subprocessor, export, trust_center_watermark_config, standard]
+    define parent: [user, program, organization, control, procedure, template, document_data, contact, internal_policy, narrative, evidence, note, trust_center_setting, subprocessor, export, trust_center_watermark_config, standard,trustcenter_entity]
 # programs are associated with an organization but do not inherit access from the organization with
 # the exception of the owner of the organization who will have access to all programs
 # the program creator will be made the admin of the program and all other access with be assigned directly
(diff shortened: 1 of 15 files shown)


internal/ent/hooks/trustcenter_entity.go
```diff
@@ -0,0 +1,65 @@
...
+			trustCenterID, ok := m.TrustCenterID()
+			if !ok || trustCenterID == "" {
+				return nil, ErrTrustCenterIDRequired
+			}
+
+			_, err := m.Client().TrustCenter.Get(ctx, trustCenterID)
+			if err != nil {
+				logx.FromContext(ctx).Error().Err(err).Msg("unable to get trust center")
+				return nil, err
+			}
```
golanglemonade: You can just do a query and don't need an ID as required input. Organizations can only have one trustcenter. We do [this](https://github.com/theopenlane/core/blob/main/internal/graphapi/trustcenterhelpers.go#L15-L36) on some of the resolvers:

```
// check if the organization has a trust center and set the id
		trustCenterID, err := withTransactionalMutation(ctx).TrustCenter.Query().OnlyID(ctx)
		if err != nil {
			if generated.IsNotFound(err) {
				return nil, rout.NewMissingRequiredFieldError("trustCenterID")
			}

			return nil, parseRequestError(ctx, err, action{action: ActionCreate, object: object})
		}
```

fga/model/model.fga
```diff
@@ -566,6 +566,16 @@ type trust_center_doc
     define nda_signed: nda_signed from parent
     define member: member from parent
 
+type trustcenter_entity
```
golanglemonade, line 572: can you add the `can_create_*` for this to the organization level?

internal/graphapi/schema/trustcenterentity.graphql
```diff
@@ -0,0 +1,103 @@
...
+
+extend type Mutation{
+    """
+    Create a new trustcenterEntity
+    """
+    createTrustcenterEntity(
+        """
+        values of the trustcenterEntity
+        """
+        input: CreateTrustcenterEntityInput!
```
golanglemonade, line 21: the create + update needs a file upload, and you'll need a hook for file -> logo file ids

fga/model/model.fga
```diff
@@ -131,6 +131,9 @@ type organization
     define trust_center_subprocessor_creator: [group#member]
     define can_create_trust_center_subprocessor: can_edit or trust_center_subprocessor_creator
 
+
+    define can_create_trust_center_entity: can_edit
+
```
golanglemonade, line 136: feel free to do this in another PR but this should be the same as the others with groups permissions access added:
```suggestion
    define trust_center_entity_creator: [group#member]
    define can_create_trust_center_entity: can_edit or trust_center_entity_creator
```

PR #1655, opened 2025-11-26
Add comments to evidence

adelowo:


golanglemonade:
discussed on slack but so I remember; we still need `addComments` and `deleteCommentID` on the `evidenceUpdate` resolvers

golanglemonade (approved):

PR #1657, opened 2025-11-28
feature/api-token-fields

### LINKED ISSUE #1475: [Feature Request] Update API key generation to use HMAC
After re-reviewing [core#284](https://github.com/theopenlane/core/issues/284) it was determined that argon2id was not an appropriate mechanism for securing API tokens. While the thought was initially valid, the encryption + decryption / verification methods are meant for low entropy human passwords (it's slow and memory-hard to resist brute-forcing weak passwords). API tokens should be high entropy random strings, so using argon2id for them would be computationally expensive, without adding much security.

https://github.com/theopenlane/iam/pull/375 adds new function to store and retreive api tokens. 

These should be in the format:
<prefix>_<public_id>_<secret>

To accomplish this we want to:
- Add two new fields to the token schemas (both api + personal access tokens) `token_public_id` and `token_secret`. Leave the existing `token` field to ensure backwards compatibility and we will drop this field in the future
- update the hooks to generate the token using the new functions and store those in the new fields 
```go
version, secret, err := tokens.GenerateAPITokenKeyMaterial()
```
- update the auth middleware to using the new `VerifyAPIToken` function to determine the `tokenID` for looking up in the database

See the README for more detailed information and examples of the setup: 
https://github.com/theopenlane/iam/blob/b5e86df87fa7553d70a5ab67d01d6111fb9f16a6/tokens/README.md

You'll need to pull in iam `v0.22.0` first into the `core` repo to do this. Be sure to run:

```
task config:generate
```

to get the new config settings and updated as needed. Token manager settings get configured in the [serverOpts](https://github.com/theopenlane/core/blob/main/internal/httpserve/serveropts/option.go#L166)

Achiket123:
resolves #1475


Achiket123:
Hey @golanglemonade can you review this. Also sorry for the delayed submission.

pkg/middleware/auth/auth.go
```diff
@@ -80,11 +80,27 @@ func Authenticate(conf *Options) echo.MiddlewareFunc {
 
 			// Get access token from the request, if not available then attempt to refresh
 			// using the refresh token cookie.
-			bearerToken, err := auth.GetBearerToken(c)
```
golanglemonade, line 83: why can't this function be used anymore? https://github.com/theopenlane/iam/blob/a5cd7151b3459a6217c6a3abbd071eaac5fec5b3/auth/auth.go#L42

Achiket123: Sorry My apologies, I didn't read the documentation. Sorry.
updated it.

golanglemonade:
@Achiket123 Looks like there are some failing tests after changing the function; can you take a look?

Achiket123:
@golanglemonade please see this

golanglemonade:
> @golanglemonade please see this

On my list; no need to keep rebasing. I just need to do some testing + verify if we need to have backwards compatibility. Should get to it later today or tomorrow

Achiket123:
> > @golanglemonade please see this
> 
> On my list; no need to keep rebasing. I just need to do some testing + verify if we need to have backwards compatibility. Should get to it later today or tomorrow

Sure, will be taking other issues.

pkg/middleware/auth/auth.go
```diff
@@ -78,13 +78,35 @@ func Authenticate(conf *Options) echo.MiddlewareFunc {
 			// Create a reauthenticator function to handle refresh tokens if they are provided.
 			reauthenticate := Reauthenticate(conf, validator)
 
-			// Get access token from the request, if not available then attempt to refresh
-			// using the refresh token cookie.
-			bearerToken, err := auth.GetBearerToken(c)
+			var bearerToken string
```
golanglemonade: @Achiket123 It looks like you switch to `GetBearerToken` but then back to this - can you explain why? if it's an issue with the function and the new token format the underlying function in `iam` should be updated instead.

pkg/middleware/auth/auth.go
```diff
@@ -397,11 +430,22 @@ func isValidPersonalAccessToken(ctx context.Context, dbClient *ent.Client,
 }
 
 func isValidAPIToken(ctx context.Context, dbClient *ent.Client, token string) (*auth.AuthenticatedUser, string, error) {
-	t, err := fetchAPITokenFunc(ctx, dbClient, token)
```
golanglemonade, line 400: I was thinking about the best way to roll this out with existing API Tokens and I think we need to have a fallback to the old format - to ensure existing API tokens work and can be migrated. This should mean that we check new format -> if its not the valid format with id + secret -> fallback to old checks -> then error

PR #1664, opened 2025-11-30
fix custom domain deleting trust center tuples

acookin:
The trust center tuples were all getting deleted on custom domain deletion because the delete keys were set in the context when we went to go update the custom domain 🫠 
adding a context key that indicates that the tuples shouldn't be deleted should fix this


golanglemonade:
Moving to draft for now; need to look into this more because that context should only ever be added on edge cleanup, which happens after the main delete operation; something is looking to be deleting the _wrong_ tuples though

PR #1666, opened 2025-12-01
Bulk update resolvers for subprocessors

adelowo:


adelowo:
branch name doesn't have `feat` :))

PR #1683, opened 2025-12-03
Clear trustcenter caches when certain items change/update

adelowo:
Will either use the custom domain name or default to `trust.theopenlane.net/slug` e.g `trust.theopenlane.net/meowmeow/`. 

All items in there will be deleted not just the `settings.json`

### DIFF
--- a/internal/ent/hooks/trustcenter_cache.go
+++ b/internal/ent/hooks/trustcenter_cache.go
@@ -5,12 +5,15 @@ import (
 
 	"entgo.io/ent"
 
+	"github.com/samber/lo"
 	"github.com/theopenlane/riverboat/pkg/riverqueue"
 
 	"github.com/theopenlane/core/internal/ent/generated"
 	"github.com/theopenlane/core/internal/ent/generated/customdomain"
 	"github.com/theopenlane/core/internal/ent/generated/hook"
 	"github.com/theopenlane/core/internal/ent/generated/trustcenter"
+	"github.com/theopenlane/core/internal/ent/generated/trustcenterdoc"
+	"github.com/theopenlane/core/internal/ent/generated/trustcentersubprocessor"
 	"github.com/theopenlane/core/internal/ent/privacy/utils"
 	"github.com/theopenlane/core/pkg/enums"
 	"github.com/theopenlane/core/pkg/jobspec"
@@ -49,11 +52,17 @@ func HookTrustcenterCacheInvalidation() ent.Hook {
 			case generated.TypeTrustcenterEntity:
 				trustCenterIDs, jobClient, shouldClearCache = handleTrustcenterEntityMutation(ctx, m)
 
+			case generated.TypeTrustCenterSubprocessor:
+				trustCenterIDs, jobClient, shouldClearCache = handleTrustCenterSubprocessorMutation(ctx, m)
+
+			case generated.TypeTrustCenterCompliance:
+				trustCenterIDs, jobClient, shouldClearCache = handleTrustCenterComplianceMutation(ctx, m)
+
 			case generated.TypeSubprocessor:
-				jobClient, shouldClearCache, shouldUseTrustCenterFromOrg = handleSubprocessorMutation(m)
+				trustCenterIDs, jobClient, shouldClearCache = handleSubprocessorMutation(ctx, m)
 
 			case generated.TypeStandard:
-				jobClient, shouldClearCache, shouldUseTrustCenterFromOrg = handleStandardMutation(m)
+				trustCenterIDs, jobClient, shouldClearCache = handleStandardMutation(ctx, m)
 			}
 
 			// tests have multiple trust centers because they bypass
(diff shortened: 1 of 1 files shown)


internal/ent/hooks/trustcenter_cache.go
```diff
@@ -0,0 +1,213 @@
...
+			var (
+				trustCenterIDs              []string
+				jobClient                   riverqueue.JobClient
+				shouldClearCache            bool
+				shouldUseTrustCenterFromOrg bool
+			)
+
+			switch mutationType {
+			case generated.TypeTrustCenterDoc:
+				docMut, ok := m.(*generated.TrustCenterDocMutation)
```
golanglemonade: can you brea each of these types into their own functions to reduce the complexity of the function + allow for easier updating?

internal/ent/hooks/trustcenter_cache.go
```diff
@@ -0,0 +1,213 @@
...
+						}
+					}
+
+					if visibility == enums.TrustCenterDocumentVisibilityPubliclyVisible ||
+						visibility == enums.TrustCenterDocumentVisibilityProtected {
+						shouldClearCache = true
+					}
+				}
+
+				if m.Op() == ent.OpDelete || m.Op() == ent.OpDeleteOne {
```
golanglemonade: I would move this above the update and us the `isDeleteOp()` function which covers soft deletes as well

internal/ent/hooks/trustcenter_cache.go
```diff
@@ -0,0 +1,213 @@
...
+				jobClient = stdMut.Job
+				shouldClearCache = true
+				shouldUseTrustCenterFromOrg = true
+			}
+
+			// tests have multiple trust centers because they bypass
+			// with privacy checks. But in reality, if coming in through
+			// graphapi ( which everything does ), it is impossible to create
+			// multiple.
+			//
```
golanglemonade, line 72: 😢 well thats fun, can you make an issue for this?

internal/ent/hooks/trustcenter_cache.go
```diff
@@ -0,0 +1,213 @@
...
+				subMut := m.(*generated.SubprocessorMutation)
+				jobClient = subMut.Job
+				shouldClearCache = true
+				shouldUseTrustCenterFromOrg = true
+
+			case generated.TypeStandard:
+				stdMut := m.(*generated.StandardMutation)
+				jobClient = stdMut.Job
+				shouldClearCache = true
+				shouldUseTrustCenterFromOrg = true
```
golanglemonade: I would check the mutation for the Trust Center ID first, I think we end up setting it because its required for permissions checks in a lot of cases

internal/ent/hooks/trustcenter_cache.go
```diff
@@ -0,0 +1,213 @@
...
+					logx.FromContext(ctx).Error().Err(err).
+						Str("mutation_type", mutationType).
+						Msg("unable to fetch trustcenter")
+					return nil, ErrTrustCenterIDRequired
+				}
+			}
+
+			if shouldClearCache && len(trustCenterIDs) > 0 {
+				for _, tcID := range trustCenterIDs {
+					_ = insertClearCacheJob(ctx, jobClient, genericMut, tcID)
```
golanglemonade: Why return the error at all if this is ignoring it?

pkg/corejobs/clear_trustcenter_cache.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/theopenlane/core/pkg/logx"
+	"github.com/theopenlane/core/pkg/objects/storage"
+)
+
+const (
+	maxClearCacheRetries = 2
+)
+
+// ClearTrustCenterCacheArgs for the worker to clear trust center cache
+type ClearTrustCenterCacheArgs struct {
```
golanglemonade: Instead of putting this in here, can you put the job in the corejobs repo, and just the Args + Kind in the `jobspec` package: https://github.com/theopenlane/core/blob/main/pkg/jobspec/domains.go

pkg/corejobs/clear_trustcenter_cache.go
```diff
@@ -0,0 +1,167 @@
...
+	var prefix string
+
+	if job.Args.TrustCenterSlug != "" {
+		prefix = fmt.Sprintf("trust.theopenlane.net/%s/", strings.TrimSuffix(job.Args.TrustCenterSlug, "/"))
+	}
+
+	if job.Args.CustomDomain != "" {
+		prefix = strings.TrimSuffix(job.Args.CustomDomain, "/")
+	}
+
```
golanglemonade: how come with the slug there is a trialing `/` but with custom domain you remove it?

internal/ent/hooks/trustcenter_cache.go
adelowo: these two do not have trustcenter id mutations on them directly

internal/ent/hooks/trustcenter_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	subMut := m.(*generated.SubprocessorMutation)
+	jobClient := subMut.Job
+	shouldClearCache := true
+	shouldUseTrustCenterFromOrg := true
+
+	return jobClient, shouldClearCache, shouldUseTrustCenterFromOrg
+}
+
+func handleStandardMutation(m ent.Mutation) (riverqueue.JobClient, bool, bool) {
+	stdMut := m.(*generated.StandardMutation)
```
golanglemonade: oh, I see the issue, these are the ones you want:

```
TrustCenterSubprocessorMutation
TrustCenterComplianceMutation
```

We only want the subprocessor specific mutation if the name or the logo of the subprocessor changes and it's linked to a trust center.

Same with standard, we only want that if the name or the logo file or logo url changes _and_ its linked to a trust center .

PR #1690, opened 2025-12-04
Update risk status enums

adelowo:


pkg/enums/riskstatus.go
```diff
@@ -10,24 +10,26 @@ import (
 type RiskStatus string
 
 var (
-	// RiskOpen indicates that the risk is open and has not been mitigated
-	RiskOpen RiskStatus = "OPEN"
```
matoszz, line 14: I don't think you should remove this

golanglemonade (changes requested):
@adelowo none of them should be removed because we need backwards compatibility to release like we did on the evidence ones

golanglemonade (approved):

PR #1693, opened 2025-12-05
feature: export documents as multiple formats (PDF, Markdown, DOCX)

Achiket123:


Achiket123:
Hey @golanglemonade can you review this ?

pkg/enums/exportformat.go
```diff
@@ -12,6 +12,12 @@ type ExportFormat string
 var (
 	// ExportFormatCsv indicates the csv.
 	ExportFormatCsv ExportFormat = "CSV"
+	// ExportFormatMD indicates the markdown.
+	ExportFormatMD ExportFormat = "MD"
```
adelowo, line 16: I think this should be Markdown instead of MD

pkg/corejobs/test/export/main.go
```diff
@@ -1,5 +1,3 @@
-//go:build ignore
-
```
golanglemonade, line 2: this should be left here; you just need to run this main with the `-tag=ignore`. This allows it to be left out of the main compiled binary for the server

pkg/corejobs/watermark_doc.go
```diff
@@ -40,7 +40,7 @@ var (
 
 // WatermarkDocArgs for the worker to process watermarking of a document
 type WatermarkDocArgs struct {
-	// TrustCenterDocumentID is the ID of the trust center document to watermark
+	// ID of the trust center document
 	TrustCenterDocumentID string `json:"trust_center_document_id"`
 }
```
golanglemonade, line 44: Can you revert this change? Looks like you might have have a merge conflict and grabbed your branch instead of latest main? Or you just aren't in up to date with main

golanglemonade:
@Achiket123 I'm going to close this out since most of this code needs to go over https://github.com/theopenlane/riverboat/tree/main/pkg/jobs. Let me know if you are still interested in finishing this up.

Achiket123:
Yeah , i am trying to complete it. But formatting is causing the problem. I am trying to get it with proper formatting. Also some code like for enums and formats is coming from core, while rest of the logic is on riverboat. I had moved the code to riverboat, but stuck on formatting. I tried using external libraries, but i don't know if it will work everywhere.

golanglemonade:
> Yeah , i am trying to complete it. But formatting is causing the problem. I am trying to get it with proper formatting. Also some code like for enums and formats is coming from core, while rest of the logic is on riverboat. I had moved the code to riverboat, but stuck on formatting. I tried using external libraries, but i don't know if it will work everywhere.

@Achiket123 Feel free to put up a PR and I can help figure out formatting stuff with you. 

If it helps, I recently had to add this on the frontend in our editor and had to write custom exports for PDF + DOCx: 
PDF: https://github.com/theopenlane/openlane-ui/blob/main/packages/ui/src/components/editor/plugins/plate-pdf.tsx
DOCX: https://github.com/theopenlane/openlane-ui/blob/main/packages/ui/src/components/editor/plugins/plate-docx.tsx

You basically have to tell it what to do with all the html data. Markdown _should_ be easier, docx is definitely the worst and this attempt isn't even 100% yet.

Achiket123:
So can I create a custom go package the converts the raw html to other format and use it ? If so i will definitely gonna try that. If that didn't work I will put up the code and ask for the help.

PR #1696, opened 2025-12-05
Auto-Set Evidence Status to missing_artifact

adelowo:
- **set status to missing artifact automatically**
- **check if assessment is tied to a template before setting template id**

### DIFF
--- a/internal/ent/hooks/evidence.go
+++ b/internal/ent/hooks/evidence.go
@@ -52,9 +52,15 @@ func HookEvidenceFiles() ent.Hook {
 				hasURL := checkEvidenceHasURL(ctx, m)
 				hasFiles := checkEvidenceHasFiles(ctx, m)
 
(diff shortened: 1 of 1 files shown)


internal/ent/hooks/evidence.go
```diff
@@ -45,6 +48,17 @@ func HookEvidenceFiles() ent.Hook {
...
 
+			if !isDeleteOp(ctx, m) {
+				hasURL := checkEvidenceHasURL(ctx, m)
+				hasFiles := checkEvidenceHasFiles(ctx, m)
+
+				if !hasURL && !hasFiles {
+					m.SetStatus(enums.EvidenceStatusMissingArtifact)
+				} else {
+					m.SetStatus(enums.EvidenceStatusSubmitted)
+				}
```
golanglemonade: I would leave off the `else`. `Submitted` is already the default status
```suggestion
				} 
```

golanglemonade (approved):

PR #1699, opened 2025-12-08
Create contact entity whenever you add a new assessment responder

adelowo:


internal/ent/hooks/assessment_response.go
```diff
@@ -54,6 +56,24 @@ func HookCreateAssessmentResponse() ent.Hook {
...
+					Where(contact.EmailEqualFold(email)).
+					Count(ctx)
+				if err != nil {
+					logx.FromContext(ctx).Err(err).Msg("could not fetch existing contacts")
+					return nil, ErrUnableToCreateContact
+				}
+
+				if count == 0 {
+					_, err = m.Client().Contact.Create().SetEmail(email).
+						Save(ctx)
```
golanglemonade:
```suggestion
					err = m.Client().Contact.Create().SetEmail(email).
						Exec(ctx)
```

internal/ent/hooks/assessment_response.go
```diff
@@ -54,6 +56,24 @@ func HookCreateAssessmentResponse() ent.Hook {
...
+					if err != nil {
+						logx.FromContext(ctx).Err(err).Msg("could not create contact for assessment response")
+						return nil, ErrUnableToCreateContact
```
golanglemonade: let's not return the error here and only log; that way the assessment can still be created even if the contact cannot
```suggestion
```

PR #1719, opened 2025-12-12
Store file in base64 content for database provider

adelowo:
this only stores very small files like `logoFile`, `faviconFile` .


pkg/objects/storage/providers/database/provider.go
```diff
@@ -42,10 +49,17 @@ func (p *Provider) Upload(ctx context.Context, reader io.Reader, opts *storagety
 		return nil, err
 	}
 
+	mimeType := http.DetectContentType(data)
+
+	fileMutation := client.File.UpdateOneID(fileID).
+		SetFileContents(data)
```
matoszz, line 55: I don't think we should do this; persisting files (regardless of size) as b64 encoded incurs a lot of unnecessary penalties; if there are reasons we want to have the image as b64 encoded so that we can respond with the entire contents in the response payload, we should encode the image from the byte field and inject it into the API response payload

matoszz (changes requested):

matoszz:
Closing in favor of https://github.com/theopenlane/core/pull/1753

Issue #1339, opened 2025-10-04
[Feature Request] Add ip address validator


golanglemonade:
There are several fields in the schema that are ip addresses, however, there is no validation on the field. This is to track adding a validator that can be reused on fields that expect an IP address. 

This should allow both ipv4 and ipv6 addresses. 

- Use the go standard library `net` and `ip` packages
- Do not allow loopback or unspecified IPs
```go
ip.IsLoopback()
ip.IsUnspecified()
```
- Add unit tests
- Add to `internal/ent/validator` package

Other validators are located here:
https://github.com/theopenlane/core/tree/main/internal/ent/validator

Bonus: 
Add the validator to existing ip address fields, e.g. :
https://github.com/theopenlane/core/blob/main/internal/ent/schema/jobrunner.go#L63-L65

using :
```go
Validate(validator.ValidateIPAddress()).
```

sathwikshetty33:
Hey, @golanglemonade, I have good experience working in golang, i have also contributed to open source projects in golang, can you [;ease assign me this.

golanglemonade:
Thanks @sathwikshetty33! It looks like this was picked up in this PR https://github.com/theopenlane/core/pull/1342. I'll be adding more issues this week though so if you want to keep an eye out, happy to assign you a new one.

sathwikshetty33:
@golanglemonade  Sure, Thank You!

PR #1721, opened 2025-12-13
feat: add ip address validator

SoulPancake:
Solves #1339 
Added IP address validator

### DIFF
--- a/internal/graphapi/trustcentersetting_test.go
+++ b/internal/graphapi/trustcentersetting_test.go
@@ -1,0 +1,341 @@
+package graphapi_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/samber/lo"
+	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
+
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/graphapi/testclient"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -52,6 +52,7 @@ file_uploads/*
 
 # Editor files
 .vscode
+.idea
```
SoulPancake, line 55: Wouldn't have added it but since .vscode was here, so I thought it just might be worth adding this too

internal/ent/validator/ipaddress.go
```diff
@@ -0,0 +1,31 @@
...
+		}
+
+		// Check if the IP is an unspecified address
+		if ip.IsUnspecified() {
+			return rout.InvalidField("ip_address")
+		}
+
+		return nil
+	}
+}
```
adelowo, line 17: We have ip validation here already https://github.com/theopenlane/core/blob/feat-ISS-1013/pkg/models/ipaddress.go#L9-L20

SoulPancake: @adelowo Ok!
I have refactored to call that from this function
This validator should be maintained here right

> Add to internal/ent/validator package

according to the above desc.

golanglemonade (approved):

golanglemonade:
thanks @SoulPancake!

PR #1723, opened 2025-12-13
Transition workflow schemas to object owned; Approval mixin

matoszz:
- Transitions workflow schemas to object owned from organization owned
- Integrates additional workflow schema policy changes and other misc updates found during development
- adds workflow definitions to fga model
- Adds workflow approval mixin with additional fields
- Updates control, evidence, and internal policies as initial workflow targets; adds entx annotations related to workflow to the control schema
- Adds additional type definitions, scalars, models, and field definitions for the control mixin to include implementation guidance and testing procedures


internal/ent/schema/mixin_wf_approval.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Fields of the ApprovalRequiredMixin
+func (WorkflowApprovalMixin) Fields() []ent.Field {
+	return []ent.Field{
+		field.JSON("proposed_changes", map[string]any{}).
+			Optional().
+			Comment("pending changes awaiting workflow approval"),
+		field.String("proposed_by_user_id").
+			Optional().
+			Comment("user who proposed the changes"),
```
golanglemonade, line 24: Should this be a FK to the users table?

internal/ent/schema/mixin_wf_approval.go
```diff
@@ -0,0 +1,30 @@
...
+		field.Time("proposed_at").
```
golanglemonade, line 25: Should this a have a default value?
```suggestion
		field.Time("proposed_at").
			Default(time.Now()).
```

internal/ent/schema/workflowassignment.go
```diff
@@ -85,26 +90,38 @@ func (w WorkflowAssignment) Edges() []ent.Edge {
...
 		}),
 		uniqueEdgeTo(&edgeDefinition{
 			fromSchema: w,
 			edgeSchema: Group{},
 			name:       "actor_group",
 			field:      "actor_group_id",
 			comment:    "Group that acted on this assignment",
+			annotations: []schema.Annotation{
+				accessmap.EdgeNoAuthCheck(),
+			},
```
golanglemonade: Is something else enforcing a minimum of view access to these edges?

internal/ent/schema/workflowassignmenttarget.go
```diff
@@ -66,20 +70,29 @@ func (w WorkflowAssignmentTarget) Edges() []ent.Edge {
...
 		}),
 		uniqueEdgeTo(&edgeDefinition{
 			fromSchema: w,
 			edgeSchema: Group{},
 			name:       "group_target",
 			field:      "target_group_id",
 			comment:    "Group target when target_type is GROUP",
+			annotations: []schema.Annotation{
+				accessmap.EdgeNoAuthCheck(),
+			},
```
golanglemonade: Same here; why no auth added? (This is an uncommon requirement so want to make sure it's intentional that this will bypass any auth check when adding the edge - this appears on a bunch of updated schemas further down as well but not going to add the same comment on all of them)

internal/ent/schema/workflowinstance.go
```diff
@@ -112,11 +159,20 @@ func (WorkflowInstance) Modules() []models.OrgModule {
...
+func (WorkflowInstance) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		entfga.SelfAccessChecks(),
+	}
+}
+
 // Policy of the WorkflowInstance
 func (WorkflowInstance) Policy() ent.Policy {
 	return policy.NewPolicy(
+		policy.WithQueryRules(),
```
golanglemonade: This is empty, is it needed?

internal/ent/schema/mixin_wf_approval.go
matoszz: no

internal/ent/schema/mixin_wf_approval.go
matoszz: no

internal/ent/schema/workflowassignmenttarget.go
matoszz: its not needed - this is an assignment target table which the system controls

internal/ent/schema/workflowassignment.go
matoszz: nothing is needed to be enforced, the workflow assignment doesn't require permissions to be managed because its created by the system

internal/ent/schema/workflowinstance.go
matoszz: seems like a rhetorical question, if you want me to remove it i can?

PR #1724, opened 2025-12-14
Bug: fix watermarking config logic & font validation error

matoszz:
This part of the ent hook was preventing the trust center watermarking configuration from ever being updated / reset:

```go
			if _, ok := m.WatermarkingEnabled(); ok {
				if oldWatermarkingEnabled, err := m.OldWatermarkingEnabled(ctx); err == nil && oldWatermarkingEnabled {
					m.ResetWatermarkingEnabled()
```

The logic in the hook and corresponding tests now behaves like this:

Create hook behavior:
- Not explicitly set → Inherits from global config
- Explicitly set to true → Stays true (skips the check entirely)
- Explicitly set to false + global config false → Stays false
- Explicitly set to false + global config true → Forced to true (global override)

Update hook behavior:
- Attempting to set to false + global config false → Allowed to be false
- Attempting to set to false + global config true → Forced to true (global override)

Results in roughly:

- If global config IsEnabled == true: Force watermarking to true (overrides any user input)
- If global config IsEnabled == false AND user didn't set a value: Default to false
- If global config IsEnabled == false AND user explicitly set a value: Honor the user's choice

I also changed the hook to explicitly use the trust center ID's over the "owner ID" given several of the queries would have problems in the event there were multiple trust centers in a single organization (which is possible today). 

Additionally, this PR fixes a minor issue with the font field surfacing as required input when performing updates to the trust center configuration.

### DIFF
--- a/internal/ent/hooks/trustcenterdoc.go
+++ b/internal/ent/hooks/trustcenterdoc.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 
 	"entgo.io/ent"
-	"github.com/theopenlane/iam/auth"
 	"github.com/theopenlane/iam/fgax"
 	"github.com/theopenlane/utils/contextx"
 
@@ -63,33 +62,39 @@ func HookCreateTrustCenterDoc() ent.Hook {
 				m.SetOriginalFileID(docFiles[0].ID)
 			}
 
-			watermarkingEnabled, watermarkingEnabledSet := m.WatermarkingEnabled()
-			if !watermarkingEnabled {
-				orgID, _ := auth.GetOrganizationIDFromContext(ctx)
+			_, watermarkingEnabledSet := m.WatermarkingEnabled()
+			trustCenterID, trustCenterIDSet := m.TrustCenterID()
 
-				if orgID != "" {
-					// check the config to see if watermarking is enabled by default
-					config, err := m.Client().TrustCenterWatermarkConfig.Query().
-						Where(trustcenterwatermarkconfig.OwnerID(orgID)).
-						Only(ctx)
-					if err == nil && config != nil {
-						watermarkingEnabled = config.IsEnabled
-
-						m.SetWatermarkingEnabled(watermarkingEnabled)
+			if trustCenterIDSet {
+				// check the config to see if watermarking is enabled by default
+				config, err := m.Client().TrustCenterWatermarkConfig.Query().
+					Where(trustcenterwatermarkconfig.TrustCenterID(trustCenterID)).
+					Only(ctx)
+				if err == nil && config != nil {
+					// If global config is enabled, force document-level watermarking to true
+					if config.IsEnabled {
+						m.SetWatermarkingEnabled(true)
+					} else if !watermarkingEnabledSet {
+						// Global config is disabled, and user didn't explicitly set value, use config default
+						m.SetWatermarkingEnabled(config.IsEnabled)
 					}
 				}
 			}
 
+			// Get the final watermarking enabled value after config overrides
+			finalWatermarkingEnabled, _ := m.WatermarkingEnabled()
+
 			// if we have no uploaded files
 			if !mutationSetsOriginalFileID && len(docFiles) == 0 {
 				// check if watermarking is enabled because if it is a file must be present
-				if watermarkingEnabledSet && watermarkingEnabled {
+				if finalWatermarkingEnabled {
 					return nil, errMissingFileID
 				}
 
 				// otherwise set the visibility to NOT_VISIBLE
 				m.SetVisibility(enums.TrustCenterDocumentVisibilityNotVisible)
-			} else if !watermarkingEnabledSet || !watermarkingEnabled {
+			} else if !finalWatermarkingEnabled {
+				// Watermarking is disabled, so set file_id = original_file_id
 				origFileID, origFileIDSet := m.OriginalFileID()
 				if !origFileIDSet {
 					return nil, errMissingFileID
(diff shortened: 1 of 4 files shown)


internal/ent/hooks/trustcenterdoc.go
```diff
@@ -63,33 +62,39 @@ func HookCreateTrustCenterDoc() ent.Hook {
...
-				if orgID != "" {
-					// check the config to see if watermarking is enabled by default
-					config, err := m.Client().TrustCenterWatermarkConfig.Query().
-						Where(trustcenterwatermarkconfig.OwnerID(orgID)).
-						Only(ctx)
-					if err == nil && config != nil {
-						watermarkingEnabled = config.IsEnabled
-
-						m.SetWatermarkingEnabled(watermarkingEnabled)
+			_, watermarkingEnabledSet := m.WatermarkingEnabled()
```
golanglemonade, line 65: > I also changed the hook to explicitly use the trust center ID's over the "owner ID" given several of the queries would have problems in the event there were multiple trust centers in a single organization (which is possible today).

Just an FYI; @adelowo put this is a comment on another PR; this is _only_ true in tests that an org can have multiple trust centers - not actually true in practice. I haven't looked yet but I believe this is due to where the uniqueness is enforced. In reality; you actually shouldn't need _any_ filter at all if we fix how the tests are setup because the interceptor would just return the trust center for the organization if there was one.

internal/ent/hooks/trustcenterdoc.go
```diff
@@ -151,10 +156,22 @@ func HookUpdateTrustCenterDoc() ent.Hook { // nolint:gocyclo
...
+					if err == nil && doc != nil {
+						config, err := m.Client().TrustCenterWatermarkConfig.Query().Where(trustcenterwatermarkconfig.TrustCenterID(doc.TrustCenterID)).Only(ctx)
+						if err == nil && config != nil {
+							// If global config is enabled, force document-level watermarking to true
+							if config.IsEnabled && !newWatermarkingEnabled {
+								logx.FromContext(ctx).Debug().Msg("global watermark config is enabled, forcing document watermarking to true")
+								m.SetWatermarkingEnabled(true)
+							}
+						}
+					}
```
golanglemonade, line 174: It feels like this is being written for a case when an organization would ever have more than one trust center or watermark config which isn't true. I think we should fix the tests to reflect reality which would simplify this logic drastically to 

```suggestion
						config, err := m.Client().TrustCenterWatermarkConfig.Query().Only(ctx)
						if err == nil && config != nil {
							// If global config is enabled, force document-level watermarking to true
							if config.IsEnabled && !newWatermarkingEnabled {
								logx.FromContext(ctx).Debug().Msg("global watermark config is enabled, forcing document watermarking to true")
								m.SetWatermarkingEnabled(true)
							}
						}
					
```

internal/graphapi/trustcenterdoc_globalconfig_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/privacy"
+	"github.com/theopenlane/core/internal/ent/generated/trustcenterwatermarkconfig"
+	"github.com/theopenlane/core/internal/graphapi/testclient"
+	"github.com/theopenlane/core/pkg/objects/storage"
+)
+
+// TestTrustCenterDocGlobalConfigOverride tests the global watermark config override logic
+func TestTrustCenterDocGlobalConfigOverride(t *testing.T) {
+	trustCenter := (&TrustCenterBuilder{client: suite.client}).MustNew(testUser1.UserCtx, t)
```
golanglemonade, line 20: Anything with trust centers, where we'd only ever allow one per org, should create a new user/organization instead of reusing `testUser1`. 
```
    testUser := suite.userBuilder(context.Background(), t)
	trustCenter := (&TrustCenterBuilder{client: suite.client}).MustNew(testUser.UserCtx, t)
	....
```

internal/graphapi/trustcenterdoc_globalconfig_test.go
```diff
@@ -0,0 +1,163 @@
...
+		assert.Assert(t, updateResp != nil)
+
+		dbDoc, err = suite.client.db.TrustCenterDoc.Get(dbCtx, docID)
+		assert.NilError(t, err)
+		assert.Check(t, is.Equal(false, dbDoc.WatermarkingEnabled), "with global config false, document watermarking can be disabled")
+
+		(&Cleanup[*generated.TrustCenterDocDeleteOne]{client: suite.client.db.TrustCenterDoc, ID: docID}).MustDelete(testUser1.UserCtx, t)
+	})
+
+	t.Run("when global config is true, documents cannot be set to false", func(t *testing.T) {
```
golanglemonade, line 81: This is a change in behavior I want to be sure we want. The requirements I had were:
1. Config created by default, enables watermarking
2. If config is set to true -> all docs by default have water marking enabled
3. If config is set to false -> all docs by default have water marking disabled 
4. In _both_ cases the individual doc can override the global setting

internal/graphapi/trustcenterdoc_test.go
```diff
@@ -1636,27 +1648,31 @@ func TestTrustCenterDocWatermarkingEnabledPreventReset(t *testing.T) {
 		}
 	}
 
-	file := createPDFUpload()
-	expectUpload(t, suite.client.mockProvider, []graphql.Upload{*file})
+	t.Run("global config true prevents disabling watermark on update", func(t *testing.T) {
```
golanglemonade, line 1651: Same here; this is a change in behavior from what I have defined

internal/ent/hooks/trustcenterdoc.go
matoszz: you make it seem like you wouldn't have watermarking as false?

internal/graphapi/trustcenterdoc_globalconfig_test.go
matoszz: I'm not entirely sure what the "global" config is even for then ? Like why have a totally separate setting that you'd change on individual documents? If it's simply a default value, why not remove it entirely and just set the default for individual docs to true and require someone disable it rather than create a bunch of weird situations where you may have conflicting values set

internal/graphapi/trustcenterdoc_test.go
matoszz: when you say "you have defined" you mean somewhere you have written down vs. how it actually works today.... right? because as it stands you can never disable individual documents, and when you update the global configuration to false that doesn't carry forward to individual documents being created (they are always set to true)

matoszz:
seems like the "global" (but not global) config setting is intended to just control what the default value is on creation and nothing else, so closing

PR #1738, opened 2025-12-18
Fix error surfacing in handlers; Prevent double query executions

matoszz:
This PR started as a fun journey down a rabbit hole that started with the question: "Why do some of our log outputs not contain a request ID" but ended up in some interesting places - most notably determining why we were 1) emitting 2-3x log statements for a single operation / error, and 2) executing the same SQL query multiple times for the same operation.

Regarding the "why some of our logs don't have request ID", this ended up being related to the `CustomHTTPErrorHandler` we add in our base echo server. What was happening was:

1. Handler returns error
2. LoggingMiddleware line 122 calls c.Error(err) → triggers CustomHTTPErrorHandler (log 1)
3. LoggingMiddleware line 130 calls logEvent() → logs the request summary (log 2)
4. LoggingMiddleware line 132 returns err back up the chain

Not just resulting in 3x log statements for the same operation, but also not including any of the normal metadata we attach to our logs. An update to the `CustomHTTPErrorHandler` to leverage the `logx.FromContext` will address this so we have metadata attached. I also updated all of the handler error wrappers to not re-surface the same error we were already logging, so this cut 2/3 log statements we were issuing previously (in a random isolated test, identical operations on main resulted in 26 log lines, on this branch resulted in 19, so about a 25% reduction overall in logs we didn't need).

Separately, while on the path of figuring out how to remove log lines that appeared to be identical for the same operation, I investigated our custom template for pagination and FGA filtering, and discovered that for our `CountIDs`, we were:

1. Line 664: `prepareQuery()` - runs Traverser interceptors (`defaultOrgInterceptorFunc`)
1. Line 669: wraps `.IDs()`
1. line 672: `withInterceptors()` - runs ALL interceptors
1. Inside that, `.IDs()` calls `.Scan()` which calls `scanWithInterceptors()` - runs interceptors again

This resulted in us executing the same SQL queries 2-3 times depending on context. Because `.IDs()` → `.Select().Scan()` → `scanWithInterceptors()` already runs all interceptors including FGA filtering -> since `IDs()` already run all the interceptors, we are able to safely drop `.prepareQuery()` and `withInterceptors()` wrapper on the template. The template was updated in this PR and code was regenerated; I tested this via many different batch operations with controls, programs, etc., and confirmed that the FGA filtering is still intact and we are now no longer executing redundant interceptors and subsequently SQL queries.

Overall, a decently large reduction in overhead for the system and less noise. As an added bonus I went through and made a few helpers to log additional information where relevant, and removed some unneeded log lines that weren't really telling us much to begin with. Also, I updated our logging middleware to attach request metadata such as origin IP, user agent, etc., to all our log emissions so that its available all the time rather than only when in debug mode.

### DIFF
--- a/internal/ent/privacy/rule/organization.go
+++ b/internal/ent/privacy/rule/organization.go
@@ -45,7 +45,7 @@ func CheckCurrentOrgAccess(ctx context.Context, m ent.Mutation, relation string)
 			return privacy.Allow
 		}
 
-		return checkOrgAccess(ctx, relation, orgID, m)
+		return checkOrgAccess(ctx, relation, orgID)
 	}
 
 	// else we need to get the object id from the mutation and get the owner id, this should only happen on deletes when using personal access tokens
@@ -53,7 +53,7 @@ func CheckCurrentOrgAccess(ctx context.Context, m ent.Mutation, relation string)
 	if ok {
 		orgID, ok = mut.OwnerID()
 		if ok && orgID != "" {
-			return checkOrgAccess(ctx, relation, orgID, m)
+			return checkOrgAccess(ctx, relation, orgID)
 		}
 	}
 
@@ -86,7 +86,7 @@ func CheckOrgAccessBasedOnRequest(ctx context.Context, relation string, query *g
 	}
 
 	for _, org := range requestedOrgs {
-		if err := checkOrgAccess(ctx, relation, org.ID, nil); err != nil && errors.Is(err, privacy.Deny) {
+		if err := checkOrgAccess(ctx, relation, org.ID); err != nil && errors.Is(err, privacy.Deny) {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


internal/ent/privacy/rule/organization.go
```diff
@@ -95,7 +95,8 @@ func CheckOrgAccessBasedOnRequest(ctx context.Context, relation string, query *g
 }
 
 // checkOrgAccess checks if the authenticated user has access to the organization
-func checkOrgAccess(ctx context.Context, relation, organizationID string) error {
+// and logs additional context about the mutation if provided
+func checkOrgAccess(ctx context.Context, relation, organizationID string, m ent.Mutation) error {
```
golanglemonade: this was updated to include `m ent.Mutation` but then not used

pkg/logx/middleware.go
```diff
@@ -177,27 +184,35 @@ func logEvent(c echo.Context, logger *Logger, config Config, start time.Time, er
...
+			evt.Str("true_client_ip", trueClientIP)
+		}
+
+		if forwardedFor := req.Header.Get("X-Forwarded-For"); forwardedFor != "" {
+			evt.Str("x_forwarded_for", forwardedFor)
+		}
+
+		if realIP := req.Header.Get("X-Real-IP"); realIP != "" {
+			evt.Str("x_real_ip", realIP)
+		}
```
golanglemonade, line 204: any reason you can't use the `attachRequestMetadata` function here?

matoszz: we could extract the metadata population logic into a helper function that both `attachRequestMetadata` and the `logEvent` sections call, but that felt a bit like the juice wasn't worth the squeeze given its only a few lines and the signature variances between zerolog.Context v zerolog.Event add complexity

golanglemonade (approved):

PR #1742, opened 2025-12-22
auto mark invites as accepted when autojoining org

adelowo:


internal/ent/hooks/usersettings.go
```diff
@@ -265,3 +277,29 @@ func autoJoinOrganizationsForUser(ctx context.Context, dbClient *generated.Clien
 
 	return nil
 }
+
+func markPendingInvitesAsAccepted(ctx context.Context, dbClient *generated.Client, email, orgID string) (int, error) {
+	invites, err := dbClient.Invite.Query().
+		Where(
+			invite.RecipientEqualFold(email),
+			invite.OwnerID(orgID),
+			invite.StatusIn(enums.InvitationSent, enums.ApprovalRequired),
```
matoszz: does it actually matter what the status is in this instance? If the user auto-joined the organization I'd think we'd want to set to accepted regardless of what the status was

PR #1745, opened 2025-12-23
fix bug where watermark cannot be updated for orgs [Do not merge for now ]

adelowo:
`OrganizationID` would not be provided in the cases of using a pat, and as such the cache key will never be set. This
is a also wrong check here as other usages of the `ProviderCacheKey` ignores the `err` and only set the cache key for the org if the org exists ( should not be a destructive action ).  

e.g https://github.com/theopenlane/core/blob/0f949492955b3bd9f84b5298718b237215eb936c/internal/objects/service.go#L193-L197

### DIFF
--- a/internal/objects/service.go
+++ b/internal/objects/service.go
@@ -6,12 +6,13 @@ import (
 	"io"
 	"time"
 
+	"github.com/theopenlane/eddy"
+	"github.com/theopenlane/iam/auth"
+
 	"github.com/theopenlane/core/pkg/logx"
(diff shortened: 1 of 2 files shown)


matoszz:
From our slack conversation: I would pull up the relevant logic surrounding setting an organization ID when a PAT is used out of this PR and incorporate it into your branch; this was closed for reasons that had nothing to do with one of the fixes inside, I simply neglected to remember to PR it separately from the jobs work: https://github.com/theopenlane/core/pull/1617

PR #1746, opened 2025-12-23
custom domain validation should only run against pending domains

adelowo:


adelowo:
closing since it was fixed in https://github.com/theopenlane/corejobs/pull/12

PR #1758, opened 2025-12-26
Notifications on mentions

### LINKED ISSUE #1694: [Feature Request] Notifications on mentions
**Describe how the feature might make your life easier or solve a problem**

I want to be alerted when my name is mentioned in a comment, internal policy, procedure, risk, or task so I know about it. 

**Describe the solution you'd like to see with any relevant context**

We want to parse the next of the corresponding text field for data that looks like:
```
<div data-slate-node=\"element\" data-slate-inline=\"true\" data-slate-void=\"true\" data-slate-key=\"01KBR2E182QTFQ27TS4B87G2MB\" data-slate-type=\"mention\" data-slate-value=\"Sarah Funkhouser\" data-slate-id=\"s-DpyYQ0Xk\" class=\"slate-mention inline-block rounded-md bg-muted px-1.5 py-0.5 align-baseline text-sm font-medium\" style=\"position:relative\">
```

`data-slate-key` -> `user.ID` 
` data-slate-value` -> `user.DisplayName`

We _only_ want to do this if that was added; I _think_ you should be able to use the `data-slate-id=` and compare the old vs. new value for the field and see if it was there before and only add a notification when that id is new but we should test this. 

Fields we want to look at:
```
comment.text
internalPolicy.Details
procedure.Details
risk.Details
task.Details
```

Notes:
- Make a package for slate parsing and put it in `pkg`, its possible we might want to use this for other things
- make a function to `checkForMentions` -> this should return a `map[string]Mention{}`  where the key is the `data-slate-id` and `Mention` contains: userID, userDisplayName, `objectType`, `objectID`, `objectName`
- make a function to `checkForNewMentions` that takes old text, new text, and list of slate ids -> returns only new ids
- define the fields we are checking for in slack as const to make it easier to expand later + keep up with any breaking changes
- We can then use this to parse events from these object types -> create notifications

manab-pr:
user will be getting notifications on mentions.

Close #1694

### DIFF
--- a/internal/ent/notifications/mentions.go
+++ b/internal/ent/notifications/mentions.go
@@ -446,32 +446,39 @@ func addMentionNotification(ctx *soiree.EventContext, input mentionNotificationI
 // objectMentionDetails holds the extracted details for mention processing
 type objectMentionDetails struct {
 	objectID       string
+	objectType     string
 	objectName     string
 	ownerID        string
 	newDetailsJSON string
 	oldDetailsJSON string
-	valid          bool
+	// newDetails and oldDetails are plain text fallbacks when JSON fields are empty.
+	// Not all editors use the JSON format - some store content in plain Details field.
+	newDetails string
+	oldDetails string
+	valid      bool
 }
 
-// documentMutation is an interface for mutations that have Name, DetailsJSON, and OwnerID fields.
+// documentMutation is an interface for mutations that have Name, Details, DetailsJSON, OwnerID, and Type fields.
 // This covers Risk, Procedure, and InternalPolicy mutations which share the same field structure.
 type documentMutation interface {
 	Name() (r string, exists bool)
+	Details() (r string, exists bool)
 	DetailsJSON() (r []any, exists bool)
 	OwnerID() (r string, exists bool)
+	Type() string
 }
 
 // oldDocumentDetails holds the old document details fetched from the database
 type oldDocumentDetails struct {
 	name        string
 	ownerID     string
+	details     string
 	detailsJSON []any
 }
 
 // handleObjectMentions is a generic handler for checking mentions in object details fields.
 // It uses type switching on the mutation to handle different object types.
 func handleObjectMentions(ctx *soiree.EventContext, payload *events.MutationPayload) error {
-	var objectType string
 	if payload == nil {
 		return nil
 	}
@@ -490,34 +497,30 @@ func handleObjectMentions(ctx *soiree.EventContext, payload *events.MutationPayl
 	case *generated.TaskMutation:
 		// Task is special - it has Title instead of Name
 		details = extractTaskMentionDetails(allowCtx, client, payload, mut)
-		objectType = "Task"
 	case *generated.RiskMutation:
 		details = extractDocumentMentionDetails(payload, mut, func() (*oldDocumentDetails, error) {
 			risk, err := client.Risk.Get(allowCtx, payload.EntityID)
 			if err != nil {
 				return nil, err
 			}
-			return &oldDocumentDetails{name: risk.Name, ownerID: risk.OwnerID, detailsJSON: risk.DetailsJSON}, nil
+			return &oldDocumentDetails{name: risk.Name, ownerID: risk.OwnerID, details: risk.Details, detailsJSON: risk.DetailsJSON}, nil
 		})
-		objectType = "Risk"
 	case *generated.ProcedureMutation:
 		details = extractDocumentMentionDetails(payload, mut, func() (*oldDocumentDetails, error) {
 			proc, err := client.Procedure.Get(allowCtx, payload.EntityID)
 			if err != nil {
 				return nil, err
 			}
-			return &oldDocumentDetails{name: proc.Name, ownerID: proc.OwnerID, detailsJSON: proc.DetailsJSON}, nil
+			return &oldDocumentDetails{name: proc.Name, ownerID: proc.OwnerID, details: proc.Details, detailsJSON: proc.DetailsJSON}, nil
 		})
-		objectType = "Procedure"
 	case *generated.InternalPolicyMutation:
 		details = extractDocumentMentionDetails(payload, mut, func() (*oldDocumentDetails, error) {
 			policy, err := client.InternalPolicy.Get(allowCtx, payload.EntityID)
 			if err != nil {
 				return nil, err
 			}
-			return &oldDocumentDetails{name: policy.Name, ownerID: policy.OwnerID, detailsJSON: policy.DetailsJSON}, nil
+			return &oldDocumentDetails{name: policy.Name, ownerID: policy.OwnerID, details: policy.Details, detailsJSON: policy.DetailsJSON}, nil
 		})
-		objectType = "InternalPolicy"
 	default:
 		return nil
 	}
@@ -526,18 +529,28 @@ func handleObjectMentions(ctx *soiree.EventContext, payload *events.MutationPayl
 		return nil
 	}
 
-	// If no new details JSON, nothing to check
-	if details.newDetailsJSON == "" {
+	// Determine which text fields to use - prefer JSON but fall back to plain text.
+	// Not all editors use the JSON format - some store content in plain Details field.
+	newText := details.newDetailsJSON
+	oldText := details.oldDetailsJSON
+
+	if newText == "" {
+		newText = details.newDetails
+		oldText = details.oldDetails
+	}
+
+	// If no new details, nothing to check
+	if newText == "" {
 		return nil
 	}
 
 	// Check if the new details contain valid Slate text
-	if !slateparser.IsValidSlateText(details.newDetailsJSON) {
+	if !slateparser.IsValidSlateText(newText) {
 		return nil
 	}
 
 	// Check for new mentions
-	newMentions := slateparser.GetNewMentions(details.oldDetailsJSON, details.newDetailsJSON, objectType, details.objectID, details.objectName)
+	newMentions := slateparser.GetNewMentions(oldText, newText, details.objectType, details.objectID, details.objectName)
 	if len(newMentions) == 0 {
 		return nil
 	}
@@ -551,7 +564,7 @@ func handleObjectMentions(ctx *soiree.EventContext, payload *events.MutationPayl
 	// Create mention notifications
 	input := mentionNotificationInput{
 		mentionedUserIDs: mentionedUserIDs,
-		objectType:       objectType,
+		objectType:       details.objectType,
 		objectID:         details.objectID,
 		objectName:       details.objectName,
 		ownerID:          details.ownerID,
@@ -570,12 +583,18 @@ func handleObjectMentions(ctx *soiree.EventContext, payload *events.MutationPayl
 // Task is handled separately because it has Title instead of Name.
 func extractTaskMentionDetails(allowCtx context.Context, client *generated.Client, payload *events.MutationPayload, taskMut *generated.TaskMutation) objectMentionDetails {
 	details := objectMentionDetails{
-		objectID: payload.EntityID,
-		valid:    true,
+		objectID:   payload.EntityID,
+		objectType: taskMut.Type(),
+		valid:      true,
 	}
 
 	if detailsJSON, exists := taskMut.DetailsJSON(); exists {
 		details.newDetailsJSON = jsonSliceToString(detailsJSON)
+	}
+
+	// Also extract plain Details as fallback - not all editors use JSON format
+	if plainDetails, exists := taskMut.Details(); exists {
+		details.newDetails = plainDetails
 	}
 
 	if title, exists := taskMut.Title(); exists {
@@ -592,6 +611,7 @@ func extractTaskMentionDetails(allowCtx context.Context, client *generated.Clien
 			task, err := client.Task.Get(allowCtx, details.objectID)
 			if err == nil {
 				details.oldDetailsJSON = jsonSliceToString(task.DetailsJSON)
+				details.oldDetails = task.Details
 				if details.objectName == "" {
 					details.objectName = task.Title
 				}
@@ -613,12 +633,18 @@ func extractDocumentMentionDetails[T documentMutation](
 	queryFunc func() (*oldDocumentDetails, error),
 ) objectMentionDetails {
 	details := objectMentionDetails{
-		objectID: payload.EntityID,
-		valid:    true,
+		objectID:   payload.EntityID,
+		objectType: mut.Type(),
+		valid:      true,
 	}
 
 	if detailsJSON, exists := mut.DetailsJSON(); exists {
 		details.newDetailsJSON = jsonSliceToString(detailsJSON)
+	}
+
+	// Also extract plain Details as fallback - not all editors use JSON format
+	if plainDetails, exists := mut.Details(); exists {
+		details.newDetails = plainDetails
 	}
 
 	if name, exists := mut.Name(); exists {
@@ -635,6 +661,7 @@ func extractDocumentMentionDetails[T documentMutation](
 			oldDoc, err := queryFunc()
 			if err == nil {
 				details.oldDetailsJSON = jsonSliceToString(oldDoc.detailsJSON)
+				details.oldDetails = oldDoc.details
 				if details.objectName == "" {
 					details.objectName = oldDoc.name
 				}


manab-pr:
@golanglemonade  added a new pkg as u asked and updated the events as well , review it and lemme know .Thank you :)

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,137 @@
...
+// Returns a map where the key is the data-slate-id and the value is the Mention struct
+func CheckForMentions(text string, objectType string, objectID string, objectName string) map[string]Mention {
+	mentions := make(map[string]Mention)
+
+	// Regular expression to match Slate mention elements
+	// Looks for: <div data-slate-node="element" data-slate-inline="true" data-slate-void="true" data-slate-key="..." data-slate-id="..." data-slate-value="...">
+	mentionRegex := regexp.MustCompile(`<div[^>]*data-slate-node="element"[^>]*data-slate-inline="true"[^>]*data-slate-void="true"[^>]*data-slate-key="([^"]*)"[^>]*data-slate-id="([^"]*)"[^>]*data-slate-value="([^"]*)"[^>]*>`)
+
+	// Also handle alternative attribute order
+	altMentionRegex := regexp.MustCompile(`<div[^>]*data-slate-key="([^"]*)"[^>]*data-slate-id="([^"]*)"[^>]*data-slate-value="([^"]*)"[^>]*data-slate-node="element"[^>]*>`)
```
adelowo: can we move the regex to a package level variable instead of compiling it every time this function is called?

manab-pr: yeah we can do that, @adelowo  if u review the whole commit , then i can change all the things at once then, thanks.

manab-pr: done

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+	"strings"
+)
+
+// Mention represents a user mention extracted from Slate formatted text
+type Mention struct {
+	UserID          string
+	UserDisplayName string
+	ObjectType      string
+	ObjectID        string
+	ObjectName      string
```
golanglemonade: all exported variables should have comments

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// FieldsToCheck defines the database fields that should be checked for mentions
+var FieldsToCheck = []string{
+	"comment.text",
+	"internalPolicy.Details",
+	"procedure.Details",
+	"risk.Details",
+	"task.Details",
```
golanglemonade: does casing matter? I noticed that its `comment.text` and not `comment.Text` whereas the others have capital `D` in `Details`.

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+func CheckForMentions(text string, objectType string, objectID string, objectName string) map[string]Mention {
+	mentions := make(map[string]Mention)
+
+	// Find all matches
+	matches := mentionRegex.FindAllStringSubmatch(text, -1)
+	altMatches := altMentionRegex.FindAllStringSubmatch(text, -1)
+
+	// Process matches from first regex
+	for _, match := range matches {
+		if len(match) >= 4 {
```
golanglemonade: you either need to define `4` as a const  or `//nolint:mnd` bot this line and line 62; the thing we are trying to achieve with that linter is "why is it 4" for future readers of this code.

What is `match[0]` and why isn't it used?

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return result
+}
+
+// ExtractMentionedUserIDs extracts just the user IDs from a mention map
+func ExtractMentionedUserIDs(mentions map[string]Mention) []string {
+	userIDs := make([]string, 0, len(mentions))
+	seen := make(map[string]bool)
+
+	for _, mention := range mentions {
```
golanglemonade, line 106: I would also ensure this is a valid `ulid` since all IDs in the system are ulids, something like before the dedupe code. 

```
	uid, err := ulids.Parse(mention.UserID)
	if err != nil {
		continue
	}
```

This will require the tests to be updated with ulids, you can use `ulids.New().String()` to generate one

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+	case "Procedure":
+		url = fmt.Sprintf("%s/procedures/%s/view", consoleURL, input.objectID)
+	case "Risk":
+		url = fmt.Sprintf("%s/risks/%s/view", consoleURL, input.objectID)
+	case "InternalPolicy":
+		url = fmt.Sprintf("%s/policies/%s/view", consoleURL, input.objectID)
+	case "Evidence":
+		url = fmt.Sprintf("%s/evidence/%s/view", consoleURL, input.objectID)
+	default:
+		url = fmt.Sprintf("%s/comments/%s", consoleURL, input.noteID)
```
golanglemonade: this URL won't be correct; if its a comment we'd need to know the parent object and have a url to that object. Looks like this is just a fallback if we have no parent. If that is the case, just don't return a url

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+
+	consoleURL := client.EntConfig.Notifications.ConsoleURL
+
+	// Build URL based on object type
+	var url string
+	switch input.objectType {
+	case "Task":
+		url = fmt.Sprintf("%s/tasks?id=%s", consoleURL, input.objectID)
+	case "Control":
+		url = fmt.Sprintf("%s/controls/%s/view", consoleURL, input.objectID)
```
golanglemonade: We don't have consistent pathing I guess :( 
```suggestion
		url = fmt.Sprintf("%s/controls/%s", consoleURL, input.objectID)
```

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+	case "Procedure":
+		url = fmt.Sprintf("%s/procedures/%s/view", consoleURL, input.objectID)
+	case "Risk":
+		url = fmt.Sprintf("%s/risks/%s/view", consoleURL, input.objectID)
```
golanglemonade:
```suggestion
		url = fmt.Sprintf("%s/risks/%s", consoleURL, input.objectID)
```

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+	case "InternalPolicy":
+		url = fmt.Sprintf("%s/policies/%s/view", consoleURL, input.objectID)
+	case "Evidence":
+		url = fmt.Sprintf("%s/evidence/%s/view", consoleURL, input.objectID)
```
golanglemonade:
```suggestion
		url = fmt.Sprintf("%s/evidence/?id=%s, consoleURL, input.objectID)
```

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
 	return nil
 }
 
+type noteFields struct {
```
golanglemonade: I think we should split this helper code for mentions into another file (same package, but another file) that is all the helpers for handling object mentions (e.g. `mentions.go`) so its a bit easier to follow and leave all the `handleXMutation` in here as the main entry points

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+	}
+
+	allowCtx := privacy.DecisionContext(ctx.Context(), privacy.Allow)
+
+	var objectID, objectName, ownerID string
+	var newDetailsJSON, oldDetailsJSON string
+
+	// Extract details based on object type
+	switch objectType {
+	case "Task":
```
golanglemonade: To decreate the complexity of this function, I would have all the case statements call an object specific function. So I would make a `handleTaskObjectMentions` that contains this code; This will improve readability as well as see the pattern to add more objects later.

golanglemonade: Any reason you passed a string through here? I think you can just do a switch on type, e.g.

```
switch mut := payload.Mutation.(type) {
	case *generated.TaskMutation:
        // do stuff
    case *generated.InternalPolicyMutation:
  ....
}

Instead of passing a string and then casing to the type after

Comment (golanglemonade) [reply internal/ent/notifications/events.go]: Actually, even better + less code we can use also an interface for all the ones that are the same fields (so excludes comments + tasks, but policies/procedures/risks, we can do something like we do in `hooks/documents.go` as an example and have:

```
type documentMutation interface {
	utils.GenericMutation

	Name() (r string, exists bool)
	Details() (r string, exists bool)
	DetailsJSON() (r []interface{}, exists bool) 
	OwnerID() (r string, exists bool)
}
```

we could leave this as the default case too, so we only do the switch on task, comments (e.g. the different ones) and then fall back and case to this type
```
docMut, ok := payload.Mutation(documentMutation)
if details, exists := docMut.DetailsJSON(); exists && len(details) > 0 {
	newDetailsJSON = fmt.Sprintf("%v", details)
}

if name, exists := docMut.Name(); exists {
	objectName = title
}

if owner, exists := docMut.OwnerID(); exists {
	ownerID = owner
}
....

```

```
type details struct {
    name string
    details string
    ownerID string
}
func queryDocumentFromDB(ctx context.Context, mut documentMutation) (*details, error) {
    ids, err := mut.IDs(ctx)
    if err != nil {
	    return nil, err
    }
    
    table := strcase.SnakeCase(pluralize.NewClient().Plural(mut.Type()))
    query := "SELECT owner_id, name, details, details_json FROM " + table + " WHERE id in ($1)"

	var rows sql.Rows
	if err := m.Client().Driver().Query(ctx, query, lo.ToAnySlice(ids), &rows); err != nil {
		logx.FromContext(ctx).Error().Err(err).Msg("failed to check for object system owned status")

		return nil, err
	}

	defer rows.Close()

	if rows.Next() {
		var (
              name string
              ownerID string
              details string
              detailsJSON string
         )
         
		if err := rows.Scan(&name); err != nil {
			logx.FromContext(ctx).Error().Err(err).Msg("failed to scan name field")

			return nil, err
		}

		if err := rows.Scan(&ownerID); err != nil {
			logx.FromContext(ctx).Error().Err(err).Msg("failed to scan system owned field")

			return nil, err
		}
....

       return &details{name: name, ownerID: ownerID, details: details, detailJSON: detailsJSON}, nil
	}
	
```

Sorry for the bad formatting/probably not entirely accurate code...typing this directly in a comment without code complete :) 

But the idea here is that we won't have to write this for any additional object that follows this same schema (we already have others that we just aren't using yet in the UI) and instead we can just have the generic code that will handle 80% of the cases and we can just add the `handleXMutation` to the events.go and be good to go

**Review by golanglemonade — COMMENTED** [review]
functionality it looks good; most my comments are around scalability + readability of the code for later. Nice work

File: internal/ent/notifications/events.go [hunk 12]
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+		NotificationType: enums.NotificationTypeUser,
+		Title:            "You were mentioned",
+		Body:             fmt.Sprintf("You were mentioned in a comment on %s: %s", input.objectType, input.objectName),
+		Data:             dataMap,
+		OwnerID:          &input.ownerID,
+		Topic:            &topic,
+		ObjectType:       "Note",
+	}
+
+	return newNotificationCreation(ctx, input.mentionedUserIDs, notifInput)
```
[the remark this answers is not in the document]
golanglemonade: hm I think this is okay

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+	"risk.Details",
+	"task.Details",
+}
+
+var (
+	// mentionRegex matches Slate mention elements with standard attribute order
+	mentionRegex = regexp.MustCompile(`<div[^>]*data-slate-node="element"[^>]*data-slate-inline="true"[^>]*data-slate-void="true"[^>]*data-slate-key="([^"]*)"[^>]*data-slate-id="([^"]*)"[^>]*data-slate-value="([^"]*)"[^>]*>`)
+
+	// altMentionRegex handles alternative attribute order for Slate mention elements
+	altMentionRegex = regexp.MustCompile(`<div[^>]*data-slate-key="([^"]*)"[^>]*data-slate-id="([^"]*)"[^>]*data-slate-value="([^"]*)"[^>]*data-slate-node="element"[^>]*>`)
```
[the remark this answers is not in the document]
golanglemonade: I would agree; I dont think order matters so can we just always use the alt regex?

internal/ent/notifications/events.go
```diff
@@ -57,31 +59,33 @@ func handleTaskMutation(ctx *soiree.EventContext, payload *events.MutationPayloa
...
+		}
 	}
 
-	if err := addTaskAssigneeNotification(ctx, input); err != nil {
-		logx.FromContext(ctx.Context()).Error().Err(err).Msg("failed to add task assignee notification")
+	// Check for mentions in task details
+	if err := handleObjectMentions(ctx, payload, "Task"); err != nil {
+		logx.FromContext(ctx.Context()).Error().Err(err).Msg("failed to handle task mentions")
 		return err
 	}
```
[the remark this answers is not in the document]
golanglemonade: this is async; this comment is incorrect - it won't have an affect on the original mutation.

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+
+	topic := "mention_alert"
+	notifInput := &generated.CreateNotificationInput{
+		NotificationType: enums.NotificationTypeUser,
+		Title:            "You were mentioned",
+		Body:             fmt.Sprintf("You were mentioned in a comment on %s: %s", input.objectType, input.objectName),
+		Data:             dataMap,
+		OwnerID:          &input.ownerID,
+		Topic:            &topic,
+		ObjectType:       "Note",
```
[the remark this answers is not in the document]
golanglemonade: we should actually use `mut.Type()` and not hard-code this at all

pkg/slateparser/slate_parser.go
manab-pr: nope it doesn't.

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+	if text, exists := noteMut.Text(); exists {
+		fields.text = text
+	}
+
+	if textJSON, exists := noteMut.TextJSON(); exists {
+		// Convert []any to string for parsing
+		if len(textJSON) > 0 {
+			fields.textJSON = fmt.Sprintf("%v", textJSON)
+		}
+	}
```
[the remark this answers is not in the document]
manab-pr: done

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+	return userIDs
+}
+
+// IsValidSlateText checks if the text contains valid Slate formatted content
+func IsValidSlateText(text string) bool {
+	// Check for basic Slate structure indicators
+	return strings.Contains(text, "data-slate-node") ||
+		strings.Contains(text, "data-slate-key") ||
+		strings.Contains(text, "data-slate-id")
+}
```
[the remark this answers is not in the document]
manab-pr: done

internal/ent/notifications/events.go
```diff
@@ -439,9 +452,612 @@ func newNotificationCreation(ctx *soiree.EventContext, userIDs []string, input *
...
+	oldTextJSON string
+	entityID    string
+	ownerID     string
+	taskID      string
+	controlID   string
+	procedureID string
+	riskID      string
+	policyID    string
+	evidenceID  string
+}
```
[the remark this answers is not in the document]
manab-pr: done

pkg/slateparser/slate_parser.go
```diff
@@ -0,0 +1,138 @@
...
+
+	result := make(map[string]Mention)
+	for _, slateID := range newSlateIDs {
+		if mention, exists := newMentions[slateID]; exists {
+			result[slateID] = mention
+		}
+	}
+
+	return result
+}
```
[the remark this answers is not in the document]
manab-pr: done

pkg/slateparser/slate_parser.go
manab-pr: the entire matched string (the full <div...> element.

pkg/slateparser/slate_parser.go
manab-pr: done

internal/ent/notifications/events.go
manab-pr: Yes, it's just a fallback.

internal/ent/notifications/events.go
manab-pr: :( fixed that .

internal/ent/notifications/events.go
manab-pr: done

manab-pr:
@golanglemonade  review it once and lemme know, if things look good.

internal/ent/notifications/mentions.go
```diff
@@ -0,0 +1,649 @@
...
+	newDetailsJSON string
+	oldDetailsJSON string
+	valid          bool
+}
+
+// documentMutation is an interface for mutations that have Name, DetailsJSON, and OwnerID fields.
+// This covers Risk, Procedure, and InternalPolicy mutations which share the same field structure.
+type documentMutation interface {
+	Name() (r string, exists bool)
+	DetailsJSON() (r []any, exists bool)
```
golanglemonade, line 466: We need to also check `Details` - not all editors (specifically comments) won't use the JSON. This was added to support plate dicussions/comments in a document, so the larger editors will use these, but not always (and actually not in prod today - this should be out soon though). So we should fall back to the Details if the json is empty

internal/ent/notifications/mentions.go
```diff
@@ -0,0 +1,649 @@
...
+		objectType = "Risk"
+	case *generated.ProcedureMutation:
+		details = extractDocumentMentionDetails(payload, mut, func() (*oldDocumentDetails, error) {
+			proc, err := client.Procedure.Get(allowCtx, payload.EntityID)
+			if err != nil {
+				return nil, err
+			}
+			return &oldDocumentDetails{name: proc.Name, ownerID: proc.OwnerID, detailsJSON: proc.DetailsJSON}, nil
+		})
+		objectType = "Procedure"
```
golanglemonade: these shouldn't nee to be hard coded mut.Type() should give you this same value.

PR #1770, opened 2025-12-28
Update Workflow Mixin | Add WorkflowProposal schema | Add Workflow templates

matoszz:
Instead of making a massive PR for all of the workflow capabilities, I've lifted out the main schema updates and added them, along with performing cleanup of the old mixin files and adding new `tmpl` files which are responsible for making the schema "registration" into the workflow capabilities as seamless as possible. The vast majority of the generated helpers are for things like a callable registry for types, CEL context builders, field references, etc.

This PR also includes updates to the Workflow models, as well as a stub implementation of the future package. These stubs were needed for code generation to occur cleanly.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -77,6 +77,7 @@ steps:
               variables:
                 GITHUB_TOKEN: GITHUB_TOKEN
                 SLACK_WEBHOOK_URL: SLACK_WEBHOOK_URL
+                GITHUB_SSH_PRIVATE_KEY: GITHUB_SSH_PRIVATE_KEY
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
               always_pull: true
@@ -87,6 +88,7 @@ steps:
                 - $GOMODCACHE:$GOMODCACHE
               environment:
                 - "GITHUB_TOKEN"
+                - "GITHUB_SSH_PRIVATE_KEY"
                 - "HELM_CHART_REPO"
                 - "HELM_CHART_PATH"
                 - "SLACK_WEBHOOK_URL"
@@ -103,6 +105,7 @@ steps:
               variables:
                 GITHUB_TOKEN: GITHUB_TOKEN
                 SLACK_WEBHOOK_URL: SLACK_WEBHOOK_URL
+                GITHUB_SSH_PRIVATE_KEY: GITHUB_SSH_PRIVATE_KEY
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
               always_pull: true
@@ -113,6 +116,7 @@ steps:
                 - $GOMODCACHE:$GOMODCACHE
               environment:
                 - "GITHUB_TOKEN"
+                - "GITHUB_SSH_PRIVATE_KEY"
                 - "SLACK_WEBHOOK_URL"
                 - "BUILDKITE_BUILD_CHECKOUT_PATH=/workdir"
       - label: ":helm: update helm chart"
@@ -127,6 +131,7 @@ steps:
               variables:
                 GITHUB_TOKEN: GITHUB_TOKEN
                 SLACK_WEBHOOK_URL: SLACK_WEBHOOK_URL
+                GITHUB_SSH_PRIVATE_KEY: GITHUB_SSH_PRIVATE_KEY
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
               always_pull: true
@@ -137,6 +142,7 @@ steps:
                 - $GOMODCACHE:$GOMODCACHE
               environment:
                 - "GITHUB_TOKEN"
+                - "GITHUB_SSH_PRIVATE_KEY"
                 - "HELM_CHART_REPO"
                 - "HELM_CHART_PATH"
                 - "SLACK_WEBHOOK_URL"
@@ -153,6 +159,7 @@ steps:
               variables:
                 GITHUB_TOKEN: GITHUB_TOKEN
                 SLACK_WEBHOOK_URL: SLACK_WEBHOOK_URL
+                GITHUB_SSH_PRIVATE_KEY: GITHUB_SSH_PRIVATE_KEY
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
               always_pull: true
@@ -163,6 +170,7 @@ steps:
                 - $GOMODCACHE:$GOMODCACHE
               environment:
                 - "GITHUB_TOKEN"
+                - "GITHUB_SSH_PRIVATE_KEY"
                 - "HELM_CHART_REPO"
                 - "HELM_CHART_PATH"
                 - "SLACK_WEBHOOK_URL"
@@ -178,6 +186,7 @@ steps:
           - cluster-secrets#v1.0.0:
               variables:
                 GITHUB_TOKEN: GITHUB_TOKEN
+                GITHUB_SSH_PRIVATE_KEY: GITHUB_SSH_PRIVATE_KEY
           - docker#v5.13.0:
               image: "ghcr.io/theopenlane/build-image:latest"
               always_pull: true
@@ -185,6 +194,7 @@ steps:
               propagate-environment: true
               environment:
                 - "GITHUB_TOKEN"
+                - "GITHUB_SSH_PRIVATE_KEY"
                 - "HELM_CHART_REPO"
       - label: ":rocket: update image tag for release"
         key: "image_tag_automation"


internal/ent/generate/entc.go
```diff
@@ -399,3 +402,87 @@ func historySchemaGenerate(extensions ...entc.Extension) {
 		log.Fatal().Err(err).Msg("running ent codegen")
 	}
 }
+
+// moveWorkflowRegistry creates a hook that moves the generated workflow_registry.go file
+// from the generated package to the hooks package to avoid import cycles
+func moveWorkflowRegistry() gen.Hook {
```
golanglemonade: Both of these functions feels really weird, why not just generate them into the package you want instead of generating them into generated and then moving them? 

It feels like the should go into their own package anyways and not directly in hooks? As an another example, we have accessmap that is in `authzgenerated` , it's own package - but then called in hooks; this way we aren't mixing generated code and manually maintained code.

internal/ent/hooks/workflow_edge_extractor.go
```diff
@@ -0,0 +1,529 @@
...
+// extractChangedEdges inspects the mutation to determine which edge relationships were modified.
+// It returns: edge names, added IDs per edge, removed IDs per edge.
+func extractChangedEdges(m ent.Mutation) ([]string, map[string][]string, map[string][]string) {
+	var edgeNames []string
+	addedIDs := make(map[string][]string)
+	removedIDs := make(map[string][]string)
+
+	switch m.(type) {
+	case *generated.ActionPlanMutation:
+		mut := m.(*generated.ActionPlanMutation)
```
golanglemonade: Commening here instead of the template because its eaiser to read but you can do:

```suggestion
	switch mut := m.(type) {
	case *generated.ActionPlanMutation:
```

This removes the additional case line from each case statement

internal/ent/hooks/workflow_edge_extractor.go
```diff
@@ -0,0 +1,529 @@
...
+		if ids := mut.ApproverIDs(); len(ids) > 0 || mut.ApproverCleared() {
```
golanglemonade: this is a unique edge, should this be the singular instead? 

```
ids , ok := mut.ApproverID();
```

> ApproverIDs returns the "approver" edge IDs in the mutation. Note that IDs always returns len(IDs) <= 1 for unique edges, and you should use ApproverID instead. It exists only for internal usage by the builders.

internal/ent/hooks/workflow_edge_extractor.go
```diff
@@ -0,0 +1,529 @@
+package hooks
+
+import (
+	"entgo.io/ent"
+	"github.com/theopenlane/core/internal/ent/generated"
+)
+
+// Code generated by ent. DO NOT EDIT.
+// This file is generated to detect edge mutations for workflow triggers.
```
golanglemonade: This should be above the package name declaration for go tooling to treat this as generated

internal/ent/hooks/workflow_registry.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/theopenlane/core/internal/ent/generated/workflowobjectref"
+	wf "github.com/theopenlane/core/internal/workflows"
+)
+
+// Code generated by ent. DO NOT EDIT.
+// This file is generated to keep workflow registries in sync with ent schemas.
+
+func init() {
+	// Register WorkflowObjectRef resolvers for workflow-addressable schemas.
+	// Only entities with WorkflowApprovalMixin are registered: ActionPlan, Control, Evidence, InternalPolicy, Procedure, Subcontrol
```
golanglemonade: nit: I probably wouldn't put the types (`ActionPlan, Control, Evidence, InternalPolicy, Procedure, Subcontrol`) in the comments here for something generated, its more prone to merge conflicts because a single line would be changed with two PRs whereas, git can usually resolve different lines added

internal/ent/hooks/workflow_edge_extractor.go
```diff
@@ -0,0 +1,529 @@
...
+		if ids := mut.ApproverIDs(); len(ids) > 0 || mut.ApproverCleared() {
+			edgeNames = append(edgeNames, "approver")
+			if len(ids) > 0 {
+				addedIDs["approver"] = ids
+			}
+			if mut.ApproverCleared() {
+				removedIDs["approver"] = []string{}
+			}
+		}
+		if ids := mut.DelegateIDs(); len(ids) > 0 || mut.DelegateCleared() {
```
golanglemonade: Any reason you didn't use the more generic so you don't have to template out all this long functions? 

```
.AddedEdges()
.RemovedEdges()
.ClearedEdges()
```

Check if they are edges the workflow cares about, and if so, grab the ids with: 
```
AddedIDs("<edge_name>")
RemovedIDs("<edge_name>")
```

internal/ent/schema/mixin_grouppermissions.go
```diff
@@ -130,6 +131,13 @@ func withGroupPermissionsInterceptor() groupPermissionsOption {
 	}
 }
 
+// withWorkflowGroupEdges marks group permission edges as workflow-eligible.
+func withWorkflowGroupEdges() groupPermissionsOption {
+	return func(g *GroupPermissionsMixin) {
+		g.WorkflowEdgeEligible = true
+	}
+}
+
```
golanglemonade, line 140: Are most of the group permissions edges workflow eligible? If so, I'd reverse this logic to exclude it instead of having to add it to every declaration

internal/ent/schema/mixin_wf_approval.go
```diff
@@ -13,18 +14,23 @@ type WorkflowApprovalMixin struct {
...
+			Comment("internal marker field for workflow eligibility, not exposed in API"),
 	}
 }
+
+// Interceptors of the WorkflowApprovalMixin
+//func (WorkflowApprovalMixin) Interceptors() []ent.Interceptor {
+//	return []ent.Interceptor{
+//		interceptors.WorkflowApprovalInterceptor(),
+//	}
+//}
```
golanglemonade: I'm assuming this is commented out because you are going to come back to it but just wanted to check it was on accident.

internal/ent/schema/workflowdefinition.go
```diff
@@ -83,6 +83,19 @@ func (WorkflowDefinition) Fields() []ent.Field {
...
+			Default([]string{}),
+		field.Strings("approval_edges").
+			Comment("Derived: edges that are approval-gated for this definition; not user editable").
+			Optional().
+			Default([]string{}),
+		field.Enum("approval_submission_mode").
+			Comment("Derived: MANUAL_SUBMIT (default) or AUTO_SUBMIT for approval domains; not user editable").
+			GoType(enums.WorkflowApprovalSubmissionMode("")).
+			Optional().
+			Default(string(enums.WorkflowApprovalSubmissionModeManualSubmit)),
```
golanglemonade, line 104: If these are derived, should they have an `entgql.Skip` Create/Update annotation?

internal/ent/schema/workflowproposal.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+	}.getMixins(w)
+}
+
+// Modules this schema has access to.
+func (WorkflowProposal) Modules() []models.OrgModule {
+	return []models.OrgModule{models.CatalogBaseModule}
+}
+
+// Hooks returns the hooks for the WorkflowProposal schema
```
golanglemonade, line 142: Intentionally commented out?

internal/ent/generate/entc.go
matoszz: afaik you can't generate them into their own package using gen.Hooks and I need to have them called during the ent generation to know about the mixin + field level annotations to compose the output, and it felt 100000000x easier to do this via the existing available hooks within ent than write a bunch of bespoke ast / code walkers to do the same thing that was already happening during the codegen

internal/ent/hooks/workflow_registry.go
matoszz: but it would only be changed if schemas were added or removed, so the code would change with the comment... always? not sure what you're saying

internal/ent/hooks/workflow_edge_extractor.go
matoszz: OK I think I see what you're suggesting, let me try an update and see

internal/ent/schema/mixin_grouppermissions.go
matoszz: its only added to like 2 declarations? So answer is likely no most of the group permissions edges aren't workflow eligible only a select few

internal/ent/schema/mixin_wf_approval.go
matoszz: I have to build out more of the workflows package to be able to use the interceptor so leaving it as  a placeholder

internal/ent/schema/workflowproposal.go
matoszz: Yes I need to finish more of the package before I can use the coded hooks

internal/ent/schema/workflowdefinition.go
matoszz: I'm assuming you realize, here, that I will have to go modify all of the already generated code now to remove these and that these weren't added or changed in this PR, right?

internal/ent/generate/entc.go
matoszz: I see how you did it with accessmap vs. how i'm doing it, let me see if I can rework it to function similiarly

internal/graphapi/actionplan.resolvers.go
```diff
@@ -18,6 +19,16 @@ import (
...
+// HasPendingWorkflow is the resolver for the hasPendingWorkflow field.
+func (r *actionPlanResolver) HasPendingWorkflow(ctx context.Context, obj *generated.ActionPlan) (bool, error) {
+	panic(fmt.Errorf("not implemented: HasPendingWorkflow - hasPendingWorkflow"))
+}
+
+// ActiveWorkflowInstance is the resolver for the activeWorkflowInstance field.
+func (r *actionPlanResolver) ActiveWorkflowInstance(ctx context.Context, obj *generated.ActionPlan) (*generated.WorkflowInstance, error) {
+	panic(fmt.Errorf("not implemented: ActiveWorkflowInstance - activeWorkflowInstance"))
+}
+
```
golanglemonade, line 31: doesnt need to be as part of this PR but this need to be implemented at somepoint

golanglemonade (approved):

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/568

### Changes Preview:

golanglemonade (approved):

PR #1771, opened 2025-12-28
Default domain access for AnonJWT | DNS cleanup | misc Trust Center fixes

matoszz:
- Normalized CNAME input and trust‑center domain matching to handle schemes, case, and trailing dots to prevent failed lookups and auth for valid domains
- Reworked anonymous JWT handling to allow default‑domain access even when custom domains exist, and to recognize preview domains so users can reach both default + preview endpoints reliably
- Consolidated preview setting lifecycle into a single upsert path, deleted duplicates, and always linked the trust center to avoid orphaned/duplicated preview settings
- Trust center hooks now correctly attach provided setting IDs, avoid recreating preview domains, and ensure Pirsch + cache jobs fire on add/change/clear
- Added FGA tuple sync on TrustCenterCompliance updates (most notably when clearing / changing standard IDs to prevent stale access relationships when standards change
- Refactored file upload handling into a shared helper and reused it across settings/entities/watermark config to reduce duplicate logic and enforce single‑file constraints
- Custom domain deletion now clears both custom/preview domain pointers and includes CustomDomainID in delete jobs
- DNS verification deletes now enqueue preview domain cleanup so preview DNS records are removed when verification records are cleared
- Trust center cache handling now favors refresh jobs over invalidations and listens to more mutation types  so snapshots stay current without manual clears
- NDA email now prefers custom domains when present and avoids panics on template errors to produce correct links


internal/ent/hooks/dnsverification.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+
+			if trustCenterConfig.PreviewZoneID == "" {
+				return next.Mutate(ctx, m)
+			}
+
+			customDomains, err := m.Client().CustomDomain.Query().
+				Where(customdomain.DNSVerificationID(id)).
+				Select(customdomain.FieldID).
+				All(ctx)
```
golanglemonade: looks like you just need the IDs here so using `IDs(ctx) would be more efficient

internal/ent/hooks/dnsverification.go
```diff
@@ -0,0 +1,62 @@
...
+			if err != nil {
+				return nil, err
+			}
+
```
golanglemonade, line 39: nit: just in case you didn't know this was an option, you can use `IDs(ctx)` which returns an array of IDs if thats the only thing you need.

```suggestion
			customDomains, err := m.Client().CustomDomain.Query().
				Where(customdomain.DNSVerificationID(id)).
				IDs(ctx)
			if err != nil {
				return nil, err
			}

```

internal/ent/hooks/dnsverification.go
```diff
@@ -0,0 +1,62 @@
...
+					CustomDomainID:           cd.ID,
+					TrustCenterPreviewZoneID: trustCenterConfig.PreviewZoneID,
+				}, nil); err != nil {
+					return nil, err
+				}
+			}
+
+			return next.Mutate(ctx, m)
+		})
+	}, ent.OpDelete|ent.OpDeleteOne)
```
golanglemonade: This needs to include UpdateOperations because of soft delete
```suggestion
	}, ent.OpUpdate|ent.OpUpdateOne|ent.OpDelete|ent.OpDeleteOne)
```

internal/ent/hooks/listeners_trustcenter_cache.go
```diff
@@ -367,54 +430,24 @@ func shouldInvalidateCacheForStandard(mut *entgen.StandardMutation, operation st
...
-
 	if jobClient == nil {
-		logx.FromContext(ctx).Warn().Msg("no job client available, skipping cache invalidation job")
+		logx.FromContext(ctx).Debug().Msg("no job client available, skipping cache refresh job")
+
 		return nil
 	}
 
-	_, err = jobClient.Insert(ctx, args, nil)
+	_, err := jobClient.Insert(ctx, jobspec.SyncTrustCenterCacheArgs{
```
golanglemonade: also switch this to the new `enqueueJob` func?

```suggestion
	err := enqueueJob(ctx, jobspec.SyncTrustCenterCacheArgs{
```

internal/ent/hooks/trustcenter.go
```diff
@@ -258,14 +277,42 @@ func HookTrustCenterDelete() ent.Hook {
...
+
+				if tc.PreviewDomainID != "" {
+					if cd, err := m.Client().CustomDomain.Get(ctx, tc.PreviewDomainID); err == nil && cd.CnameRecord != "" {
+						if err := enqueueJob(ctx, m.Job, jobspec.ClearTrustCenterCacheArgs{
+							CustomDomain: cd.CnameRecord,
+						}, nil); err != nil {
+							return nil, err
+						}
+					}
+				}
```
golanglemonade, line 313: I don't think we should ever be using a cache for a preview domain, it should just always pull from the api because that will allow for rapid changes to immediately show without waiting for jobs to clear them. The cache is also there so we won't fail if our api is down, but if the api is down you wouldn't be able to make settings changes and preview them anyways

internal/ent/hooks/trustcenter.go
```diff
@@ -294,21 +347,53 @@ func HookTrustCenterUpdate() ent.Hook {
...
+			if customDomainCleared || (mutationCustomDomainIDExists && mutationCustomDomainID == "") {
+				if previousCustomDomainID != nil && previousPirschDomainID != "" {
+					if err := enqueueJob(ctx, m.Job, jobspec.DeletePirschDomainArgs{
+						PirschDomainID: previousPirschDomainID,
+					}, nil); err != nil {
+						return nil, err
+					}
+				}
+
+				if previousCustomDomainID != nil && m.Job != nil {
```
golanglemonade: nit: you only check m.Job !=nil on this one instance (e.g. the one above does't check it) and its not really needed since its checked within `enqueueJob`
```suggestion
				if previousCustomDomainID != nil {
```

internal/ent/hooks/trustcenter.go
```diff
@@ -294,21 +347,53 @@ func HookTrustCenterUpdate() ent.Hook {
...
+
+				if m.Job != nil {
+					if cd, err := m.Client().CustomDomain.Get(ctx, *previousCustomDomainID); err == nil && cd.CnameRecord != "" {
+						if err := enqueueJob(ctx, m.Job, jobspec.ClearTrustCenterCacheArgs{
+							CustomDomain: cd.CnameRecord,
+						}, nil); err != nil {
+							return nil, err
+						}
+					}
+				}
```
golanglemonade: Same here. If can see you want to avoid the queries if there is no job client to have this check for efficiency but if that were the reason, I would do it once before all of these and return early and not before some of them.  Since `enqueueJob` checks this, its not needed here to avoid nil pointers. 
```suggestion
			if cd, err := m.Client().CustomDomain.Get(ctx, *previousCustomDomainID); err == nil && cd.CnameRecord != "" {
				if err := enqueueJob(ctx, m.Job, jobspec.ClearTrustCenterCacheArgs{
					CustomDomain: cd.CnameRecord,
				}, nil); err != nil {
					return nil, err
				}
			}
```

internal/ent/hooks/trustcenterwatermarkconfig.go
```diff
@@ -62,25 +61,19 @@ func HookTrustCenterWatermarkConfig() ent.Hook {
...
-		m.SetFileID(logoFiles[0].ID)
-		m.ClearText()
-
-		adapter := objects.NewGenericMutationAdapter(m,
-			func(mut *generated.TrustCenterWatermarkConfigMutation) (string, bool) { return mut.ID() },
-			func(mut *generated.TrustCenterWatermarkConfigMutation) string { return mut.Type() },
-		)
-
-		return objects.ProcessFilesForMutation(ctx, adapter, key, "trust_center_watermark_config")
+	key := "watermarkConfig"
```
golanglemonade: I believe this needs to be added to `internal/objects/validators/mime.go` `validMimeTypes` to ensure the correct types only allowed here.

nit: I personally feel like `watermarkFile` would be a better name for this instead of `watermarkConfig` because its not the config being uploaded, but the watermark image?

internal/graphapi/trustcenterdomain.resolvers.go
```diff
@@ -25,7 +26,12 @@ func (r *mutationResolver) CreateTrustCenterDomain(ctx context.Context, input mo
 
 	transactionCtx := withTransactionalMutation(ctx)
 
-	mappableDomainID, err := transactionCtx.MappableDomain.Query().Where(mappabledomain.Name(cnameTarget)).FirstID(ctx)
+	normalizedCnameTarget := cnameTarget
+	if normalized, err := domain.NormalizeHostname(cnameTarget); err == nil {
+		normalizedCnameTarget = normalized
+	}
```
golanglemonade, line 32: If we can't normalize the hostname we are using the original? Is that what we want here?

internal/graphapi/trustcenterpreviewsetting_helpers.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+		} else if createInput != nil {
+			update := buildPreviewSettingUpdateInput(*createInput)
+			previewSetting, err = previewSetting.Update().SetInput(update).Save(ctx)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
```
golanglemonade, line 112: I think the `else` can be removed and just have:  if preview setting is nil, create it - then check for update/create input and do that since the update is the same in either case. 

```suggestion
	if previewSetting == nil {
		create := buildPreviewSettingCreateInput(trustCenter.ID, createInput)
		previewSetting, err = client.TrustCenterSetting.Create().SetInput(create).Save(ctx)
		if err != nil {
			return nil, err
		}
	} 
	
	if updateInput != nil {
		previewSetting, err = previewSetting.Update().SetInput(*updateInput).Save(ctx)
		if err != nil {
			return nil, err
		}
	} else if createInput != nil {
		update := buildPreviewSettingUpdateInput(*createInput)
		previewSetting, err = previewSetting.Update().SetInput(update).Save(ctx)
		if err != nil {
			return nil, err
		}
	}

```

internal/graphapi/trustcenterpreviewsetting_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	var gqlResp updateTrustCenterPreviewSettingResponse
+	err = graphClient.Client.Post(
+		testUser1.UserCtx,
+		"UpdateTrustCenterPreviewSetting",
+		updateTrustCenterPreviewSettingDocument,
+		&gqlResp,
+		map[string]any{
+			"input": testclient.UpdateTrustCenterSettingInput{},
+		},
```
golanglemonade: Why isn't this using the client queries and doing a `Post` instead?

internal/graphapi/trustcenterpreviewsetting_test.go
```diff
@@ -0,0 +1,99 @@
...
+type updateTrustCenterPreviewSettingResponse struct {
+	UpdateTrustCenterPreviewSetting struct {
+		TrustCenterSetting struct {
+			ID string `json:"id"`
+		} `json:"trustCenterSetting"`
+	} `json:"updateTrustCenterPreviewSetting"`
+}
+
+func TestCreateTrustCenterPreviewSettingLinksTrustCenter(t *testing.T) {
+	trustCenter := (&TrustCenterBuilder{client: suite.client}).MustNew(testUser1.UserCtx, t)
```
golanglemonade: I'm surprised this isn't causing flakey tests since orgs can only have 1 trust center. 🤔

internal/ent/hooks/trustcenter.go
golanglemonade: chatted on slack; keeping this with cache

internal/ent/hooks/trustcenterwatermarkconfig.go
matoszz: 😆 this is what I get for trying to acommodate your other nit when it was named logoFile to begin with 😮‍💨

internal/graphapi/trustcenterdomain.resolvers.go
matoszz: what's the alternative? just fail?

internal/graphapi/trustcenterpreviewsetting_test.go
matoszz: what am i missing... it is doing a post?

internal/graphapi/trustcenterdomain.resolvers.go
golanglemonade: yeah, return an error that its not valid/bad request

internal/graphapi/trustcenterpreviewsetting_test.go
golanglemonade: What is all this for at the top? `updateTrustCenterPreviewSettingDocument`? Why are you asserting its a test client? 

I would expect this just to be something like this, using the test client and existing queries:

```
resp, err := suite.client.api.UpdateTrustCenterSetting(testUser1.UserCtx, id, testclient.UpdateTrustCenterSettingInput{}, nil, nil)
```

To be honest, I'm not even sure what this part of the test is testing....you update a preview setting with empty input and you get an id back?

internal/objects/validators/mime.go
```diff
@@ -41,6 +41,7 @@ var validMimeTypes = map[string][]string{
 	"actionPlanFile":     importSchemaMimeTypes,
 	"trustCenterDocFile": {"application/pdf", "application/vnd.openxmlformats-officedocument.wordprocessingml.document"},
 	"templateFiles":      sharedMimeTypes,
+	"watermarkFile":      sharedMimeTypes,
```
golanglemonade: Watermark file needs to be an image, probably just want:
```suggestion
	"watermarkFile":      {"image/jpeg", "image/png"},
```

internal/graphapi/trustcenterpreviewsetting_test.go
matoszz: you keep saying "existing queries" but there are no types for this, there is no `UpdateTrustCenterSetting` so not sure what you're referring to.... i'll just delete the test....

adelowo (approved):

PR #1781, opened 2025-12-31
Changes for Harmonize functionality | ControlClone field updates, FGA model additions

matoszz:
Testing capabilities I developed in harmonize repo and working through issues that have come up. I created additional objects to get seeded, but ran into problems related to features being enabled, permissions surrounding role of the user creating the object. Additionally testing the "compare" logic (to determine if the framework data we have in the repo lines up with the data loaded into the API to present a "diff" logic, which would also be needed to do a "framework upgrade") I was finding a diff for freshly seeded standards and it was due to missing data in the client requests / responses. Summary of changes:

- Added a "devmode" to entitlements so that we create all the available feature modules when in dev mode
- Updated organization hooks, entitlements reconciler, and added configuration parameters to support devmode for entitlements
- Added testingProcedures and evidenceRequests to graphql queries and the clone controls functions
- Flips a few of the historically un-used schemas over to object owned instead of org owned (worked through asset, finding, vulnerability, and entity
- Adds FGA schema annotations and checks
- updates FGA model and tests for these schemas^

Additionally, the `externalReadOnly` directive was logging "no id found in context" errors hundreds of times during bulk control/subcontrol creation. Here's the original code:

```
  fieldSet := checkFieldSet(ctx, skipCreateOperations)  // Line 140
  allowed := checkSourceAllowed(ctx, source)            // Line 141

  if fieldSet && !allowed {
      return nil, ErrReadOnlyField
  }
```

The issue is that both functions are called unconditionally, even though:
- checkFieldSet with skipCreateOperations returns false for any operation named "Create*"
- When fieldSet is false, the result of checkSourceAllowed doesn't matter (due to && logic)
- But checkSourceAllowed still runs and tries to look up an object by id
- During create operations there's no id, so it logs an error

For a CreateBulkSubcontrol with 50 subcontrols × 9 fields with the directive = 450x error logs
the Fix:

```
  fieldSet := checkFieldSet(ctx, skipCreateOperations)
  if fieldSet && !checkSourceAllowed(ctx, source) { <--- short circuit
      return nil, ErrReadOnlyField
  }
```

when `fieldSet` is false, Go doesn't evaluate the right side (checkSourceAllowed) at all - it already knows the entire condition is false.

This means during create operations:
- fieldSet = false (from skipCreateOperations)
- checkSourceAllowed is never called
- No error is logged
- The directive allows the operation to proceed

### DIFF
--- a/internal/entitlements/catalog_prices.go
+++ b/internal/entitlements/catalog_prices.go
@@ -24,7 +24,7 @@ func TrialMonthlyPrices(useSandbox bool) []entitlements.Price {
 
 // AllMonthlyPrices returns prices for all monthly modules regardless of trial status
 func AllMonthlyPrices(useSandbox bool) []entitlements.Price {
-	return monthlyPrices(func(f catalog.Feature) bool {
+	return monthlyPrices(func(_ catalog.Feature) bool {
 		return true
 	}, useSandbox)
 }
--- a/internal/graphapi/directives/directives.go
+++ b/internal/graphapi/directives/directives.go
@@ -138,9 +138,7 @@ var ExternalReadOnlyDirective = func(ctx context.Context, _ any, next graphql.Re
 	}
 
 	fieldSet := checkFieldSet(ctx, skipCreateOperations)
-	allowed := checkSourceAllowed(ctx, source)
-
-	if fieldSet && !allowed {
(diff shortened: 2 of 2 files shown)


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/578

### Changes Preview:

PR #1787, opened 2026-01-01
Additional schema updates related to harmonize work

matoszz:
Additional updates associated with seeding demo orgs / seeding new object types. Ran into issues related to creation / usage of action plans as well as some missing objects on the create access mixin. Adds the releavnt objects to the mixin as well as converts the action plan schema over to object owned (expecting in the future action plans are object owned under controls, control implementation, etc.). Updates the FGA model and tests associated, regnerates output.


internal/ent/privacy/utils/utils.go
```diff
@@ -150,5 +150,10 @@ func PaymentMethodCheckRequired(client *generated.Client) bool {
 		return false
 	}
 
+	// In dev mode, bypass payment method checks to keep local workflows unblocked.
+	if client.EntConfig.Modules.DevMode {
```
golanglemonade, line 154: just an fyi - you could also set `BypassEmailDomains` for local dev as well to skip on the same path as we use in production

golanglemonade (approved):

PR #1792, opened 2026-01-05
More forgiving CSV uploads; granular error messages surfaced

matoszz:
Updated CSV bulk upload error handling so parse failures return actionable validation errors with line/column/header info and a JSON quoting hint, making something like “missing quotes in tags” self‑diagnosable. Example of new error messages (the UI needs to be updated to use these, though):

```
12:01PM ERR internal/graphapi/task.resolvers.go:60 | failed to unmarshal bulk data error=csv parse error on line 2, column 1 (header "Tags"): invalid character 's' looking for beginning of value list or object values must be valid JSON and the cell should be quoted, e.g. "[\"security\",\"compliance\"]" org_id=[01KE7GEFXHG65NRQ3YGDVK6P9W] remote_ip=::1 request_id=WhupViCnxuTNvRITrjylLSlYfWxxmcXP request_protocol=HTTP/1.1 severity=ERROR user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36 user_id=01KE7GEETYBFKK7K6RYA5MHMED
```

- Normalized enums during CSV imports using a single trim → snake case → upper flow, defaulted empty enum values appropriately, and cleared zero `DateTime` pointers created by blank CSV cells so defaults apply
- Added tests to cover enum normalization, `DateTime` pointer clearing, and CSV parse error wrapping
- Enhanced validation errors to carry field names and expose them in GraphQL extensions so clients can surface the exact fields
- Adjusted custom domain status enum helpers to accept uppercase inputs

Verified the behavior by testing several permutations of CSV uploads; confirmed things like `Open` normalize to `OPEN`, `In Review` normalizes to `IN_REVIEW`, etc.


internal/graphapi/common/helpers.go
```diff
@@ -179,12 +183,348 @@ func UnmarshalBulkData[T any](input graphql.Upload) ([]*T, error) {
...
+
+	info := enumInfoForType(field.Type().Elem(), cache)
+	if !info.ok {
+		// Not an enum type (no Values method), skip normalization
+		return
+	}
+
+	// Enum values are string-backed; normalize the raw string representation
+	raw := field.Elem().String()
+	normalized, clear := normalizeEnumValue(raw, info)
```
golanglemonade: pretty sure go won't like the var name `clear` because its a stdlib function name

PR #1820, opened 2026-01-10
Prevent deletion of standard if linked to trustcenter or control

adelowo:
always block deletions if used by trustcenter compliance schema but for controls, we should 
only care if it is not a system owned standard before we check if it is being used by other controls.

### DIFF
--- a/internal/ent/hooks/standard.go
+++ b/internal/ent/hooks/standard.go
@@ -30,6 +30,9 @@ var (
 	// ErrStandardInUseByTrustCenter defines an error that denotes a standard cannot be deleted
 	// because it is in use by an active trust center
 	ErrStandardInUseByTrustCenter = errors.New("standard cannot be deleted because it is in use by a trust center")
+	// ErrSystemOwnedStandardCannotBeDeleted defines an error that denotes a system-owned standard
+	// can only be deleted by a system admin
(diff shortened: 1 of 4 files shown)


internal/ent/hooks/standard.go
```diff
@@ -16,41 +16,54 @@ import (
...
 		return hook.StandardFunc(func(ctx context.Context, m *generated.StandardMutation) (generated.Value, error) {
-			// only run on delete operations
 			if !isDeleteOp(ctx, m) {
 				return next.Mutate(ctx, m)
 			}
 
-			// only system admins edit system-owned standards
-			// check early to avoid unnecessary database queries to both
-			// the database and the authz service
 			if !auth.IsSystemAdminFromContext(ctx) {
```
golanglemonade: Doesn't this mean we won't check non-system owned standards since non-system admin users will skip everything below?

adelowo: phew, I wonder how long this bug has been here. fixed in https://github.com/theopenlane/core/pull/1820/changes/ee4232d1d5aecce73e95abb1b282fa11f2749874 

the checks will now apply to regular users too not only admins

adelowo:
was typing too fast with commit message . should be `[fix existing bug where regular users bypassed every single check` not `non regular`

golanglemonade (approved):

PR #1833, opened 2026-01-13
mark as read

### LINKED ISSUE #1749: Mark notifications as read
The new `notifications` schema has:

```
subscription{
  notificationCreated {
    body
    id
    topic
    title
    data
    readAt
  }
}
```

We want to create a `update` resolver that will be able to mark a notification as `read`. This will be on the UI side so a user can dismiss a notification they've already seen.

manab-pr:
added update mutation for read_at field.

Close #1749


internal/ent/schema/notification.go
```diff
@@ -45,33 +44,43 @@ func (Notification) Fields() []ent.Field {
 	return []ent.Field{
 		field.String("user_id").
 			Comment("the user this notification is for").
-			Optional(),
+			Optional().
+			Annotations(entgql.Skip(entgql.SkipMutationUpdateInput)),
```
golanglemonade, line 50: This should be skipAll on everything except for readAt

internal/graphapi/query/notification.graphql
```diff
@@ -0,0 +1,118 @@
...
+    title
+    topic
+    updatedAt
+    updatedBy
+    userID
+  }
+}
+
+mutation UpdateNotification($updateNotificationId: ID!, $input: UpdateNotificationInput!) {
+  updateNotification(id: $updateNotificationId, input: $input) {
```
golanglemonade, line 2: Everything except the update should be removed, this will remove the generated resolvers as well

internal/ent/schema/notification.go
```diff
@@ -45,33 +44,43 @@ func (Notification) Fields() []ent.Field {
...
+
```
adelowo, line 51: we should probably make this field immutable with `.Immutable()`  with the other ones

internal/ent/schema/notification.go
manab-pr: okay.

internal/graphapi/query/notification.graphql
manab-pr: okay.

manab-pr:
> never submitted my review 🤦‍♀️
@golanglemonade 
who did haha :)

internal/ent/schema/notification.go
manab-pr: okay

manab-pr: `user_id` is linked to an edge though , making the edge immutable too then.

internal/ent/schema/notification.go
manab-pr: @golanglemonade  if i add `skipAll` to all the fields except read_at , then the internal CreateNotificationInput will not get these fields then .

golanglemonade: thats correct; we don't want graphql resolvers for creation - only input

manab-pr: then we can not use skipAll in there.

golanglemonade: disregard; you should leave this as is and just remove the resolver generation

manab-pr: yeah, will do .

manab-pr:
@golanglemonade  the problem is that , as you have suggested  
```
updateNotification(
        """
        ID of the notification
        """
        id: ID!
        """
        New values for the notification
        """
        input: UpdateNotificationInput!
    ): NotificationUpdatePayload!

"""
Return response for updateNotification mutation
"""
type NotificationUpdatePayload {
    """
    Updated notification
    """
    notification: Notification!
}
```
it should only contain this update func  , but the problem is i can not skip  create , if i do then createNotificationInput is not generating, so what i did , if an external client try to create/delete notifications then it will throw error, directly without any checking , and although there is one more way to do it , manually removing the create resolvers but i don't think that's viable solution.

**so, this solution works fine.**
let me know what do you think of it.

internal/graphapi/notification.resolvers.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/theopenlane/core/internal/graphapi/model"
+	"github.com/theopenlane/core/pkg/logx"
+	"github.com/theopenlane/utils/rout"
+)
+
+// CreateNotification is the resolver for the createNotification field.
+// Notifications can only be created internally by the system, not via GraphQL API
+func (r *mutationResolver) CreateNotification(ctx context.Context, input generated.CreateNotificationInput) (*model.NotificationCreatePayload, error) {
+	return nil, rout.ErrPermissionDenied
+}
```
golanglemonade: @manab-pr you shouldn't need to do this. Instead you should update the `internal/graphapi/schema/notification.graphql` to _not_ include these, and then gqlgen won't generate them.

golanglemonade:
> @golanglemonade the problem is that , as you have suggested
> 
> ```
> updateNotification(
>         """
>         ID of the notification
>         """
>         id: ID!
>         """
>         New values for the notification
>         """
>         input: UpdateNotificationInput!
>     ): NotificationUpdatePayload!
> 
> """
> Return response for updateNotification mutation
> """
> type NotificationUpdatePayload {
>     """
>     Updated notification
>     """
>     notification: Notification!
> }
> ```
> 
> it should only contain this update func , but the problem is i can not skip create , if i do then createNotificationInput is not generating, so what i did , if an external client try to create/delete notifications then it will throw error, directly without any checking , and although there is one more way to do it , manually removing the create resolvers but i don't think that's viable solution.
> 
> **so, this solution works fine.** let me know what do you think of it.

You don't need to skip Create, you just need to not include it in `internal/graphapi/schema/notification.graphql` - this is what is used to create the resolvers. These schemas are *not* automatically updated so removing it after the first gen, will keep them removed unless for some reason the schema file is completed deleted.

internal/ent/schema/notification.go
```diff
@@ -138,11 +156,11 @@ func (Notification) Annotations() []schema.Annotation {
 			Exclude: true,
 		},
 		// skip schema gen, this is used for subscriptions only
-		entx.SchemaGenSkip(true),
+		// entx.SchemaGenSkip(true),
 		// skip query gen, this is used for subscriptions only
-		entx.QueryGenSkip(true),
-		entgql.Skip(entgql.SkipMutationUpdateInput | entgql.SkipWhereInput | entgql.SkipOrderField),
-		entgql.Mutations(entgql.MutationCreate()), // only allow creation via mutations, updates are not supported
+		// entx.QueryGenSkip(true),
```
golanglemonade:
```suggestion
=```

File: internal/ent/schema/notification.go [hunk 6]
```diff
@@ -138,11 +156,11 @@ func (Notification) Annotations() []schema.Annotation {
 			Exclude: true,
 		},
 		// skip schema gen, this is used for subscriptions only
-		entx.SchemaGenSkip(true),
+		// entx.SchemaGenSkip(true),
```
golanglemonade:
```suggestion
```

internal/graphapi/schema/notification.graphql
```diff
@@ -0,0 +1,103 @@
...
+    Look up notification by ID
+    """
+     notification(
+        """
+        ID of the notification
+        """
+        id: ID!
+    ):  Notification!
+}
+
```
golanglemonade:
```suggestion
```

internal/graphapi/schema/notification.graphql
```diff
@@ -0,0 +1,103 @@
...
+"""
+Return response for createBulkNotification mutation
+"""
+type NotificationBulkCreatePayload {
+    """
+    Created notifications
+    """
+    notifications: [Notification!]
+}
```
golanglemonade, line 25:
```suggestion
extend type Mutation{
    """
    Update an existing notification
    """
    updateNotification(
        """
        ID of the notification
        """
        id: ID!
        """
        New values for the notification
        """
        input: UpdateNotificationInput!
    ): NotificationUpdatePayload!
}

"""
Return response for updateNotification mutation
"""
type NotificationUpdatePayload {
    """
    Updated notification
    """
    notification: Notification!
}
```

internal/graphapi/notification.resolvers.go
manab-pr: done

internal/graphapi/schema/notification.graphql
manab-pr: all done.

golanglemonade (dismissed):

PR #1843, opened 2026-01-14
Trust center doc with custom enums for category

adelowo:


adelowo:
https://github.com/theopenlane/openlane-ui/pull/1327

PR #1855, opened 2026-01-16
Workflow types | Schema scopes

matoszz:
- Add structured workflow assignment approval/rejection/invalidation metadata
- Introduce workflow metadata + assignment actions in GraphQL (approve/reject, myWorkflowAssignments) and remove direct create/update/bulk APIs for workflow assignments/targets/instances; update schemas, gqlgen configs/scalars, models, queries, and test client
- Build out internal workflow helpers/registries for object loading, CEL context building, and target resolution, plus new workflow errors and tests
- Harden entgql pagination template to guard cursor field bounds and regenerate ent/gqlgen outputs
- Bump github.com/theopenlane/core/common to v1.0.2 to leverage the new workflow types

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v6
         with:
-          go-version-file: 'go.mod'
+          go-version-file: 'cli/go.mod'
           cache: true
           cache-dependency-path: |
             **/go.sum
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/rs/zerolog v1.34.0
 	github.com/samber/lo v1.52.0
 	github.com/spf13/cobra v1.10.2
-	github.com/theopenlane/core v0.50.0
+	github.com/theopenlane/core v1.3.1
 	github.com/theopenlane/core/common v1.0.1
 	github.com/theopenlane/go-client v0.5.0
 	github.com/theopenlane/httpsling v0.3.0
@@ -41,6 +41,8 @@ require (
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/fatih/color v1.18.0 // indirect
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
+	github.com/fumiama/go-docx v0.0.0-20250506085032-0c30fd09304b // indirect
+	github.com/fumiama/imgsz v0.0.4 // indirect
 	github.com/fxamacker/cbor/v2 v2.9.0 // indirect
 	github.com/gabriel-vasile/mimetype v1.4.12 // indirect
 	github.com/go-jose/go-jose/v4 v4.1.3 // indirect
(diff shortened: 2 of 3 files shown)


db/migrations-goose-postgres/20260116205149_workflowassignmenttypes_history.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- create index "trustcenterentityhistory_history_time" to table: "trust_center_entity_history"
+CREATE INDEX "trustcenterentityhistory_history_time" ON "trust_center_entity_history" ("history_time");
```
golanglemonade: I need to figure out why this keeps getting added, I think we might need to just remove it from the init sql or something - but I'll take a look - this will fail to start the server as is.

internal/graphapi/workflowassignmentextended.resolvers.go
```diff
@@ -0,0 +1,61 @@
...
+	"entgo.io/contrib/entgql"
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/graphapi/common"
+	"github.com/theopenlane/core/internal/graphapi/model"
+	"github.com/theopenlane/gqlgen-plugins/graphutils"
+)
+
+// ApproveWorkflowAssignment is the resolver for the approveWorkflowAssignment field.
+func (r *mutationResolver) ApproveWorkflowAssignment(ctx context.Context, id string) (*model.WorkflowAssignmentApprovePayload, error) {
+	panic(fmt.Errorf("not implemented: ApproveWorkflowAssignment - approveWorkflowAssignment"))
```
golanglemonade, line 21: These two resolvers are still the default panic, are you planning on doing the implementation on a later PR?

internal/workflows/reflection.go
```diff
@@ -0,0 +1,29 @@
...
+
+import "reflect"
+
+// StringField extracts a string field by name from a struct or pointer to struct.
+func StringField(node any, field string) string {
+	if node == nil {
+		return ""
+	}
+
+	val := reflect.ValueOf(node)
```
golanglemonade: do we have to use reflection here? There is performance implications of doing so that generally is good to avoid if possible so just want to see if we can. 

I'm _assuming_ (so correct me if I'm wrong) but the node here is the node from the response, which would be a `map[string]any`? 

If so I think we can skip reflection and do something like 

```
    raw, ok := node[field]
	if !ok || raw == nil {
		return ""
	}

	switch v := raw.(type) {
	case string:
		return v
	case *string:
		if v == nil {
			return ""
		}
		return *v
	case []byte:
		return string(v)

	default:
		return ""
```

internal/workflows/types.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// Object captures the workflow target along with its concrete ent entity when available.
+type Object struct {
+	// ID is the workflow object identifier.
+	ID string
+	// Type is the workflow object type.
+	Type enums.WorkflowObjectType
+	// Node is the concrete ent entity when available.
+	Node any
```
golanglemonade, line 17: I think we can make this slightly more strict with this marshal the types so we have a more consistent type we can work with:
```suggestion
	Node map[string]any
```

golanglemonade (commented):
this is great! only concern is around the reflection part - happy to chat in slack if thats easier to understand

internal/workflows/types.go
matoszz: it might not be readily visible in the current cut of the code, but switching to `map[string]any` would be a broader refactor; most of the call sites depend on it being a concrete struct or pointer; the engine reads fields and interfaces off the concrete entity which a map would break. `Node` on `Object` is effectively a cached ent entity

internal/graphapi/workflowassignmentextended.resolvers.go
matoszz: Yes, correct. They'll be added in my next PR

golanglemonade (approved):

PR #1887, opened 2026-01-21
fix : notification mention , modified regex

### LINKED ISSUE #1874: [Bug] Notifications for mentions do not work

No notifications for mentions (in objects or comments) working

1. Create a task
2. Mention another user inside the task or with a comment on the task
3. No notification created

1. Create a task
2. Mention another user inside the task or with a comment on the task
3. Notification created for the mentioned user

manab-pr:
modified regex ,so that it allows the fields in any order.

closes #1874

### DIFF
--- a/pkg/slateparser/slate_parser.go
+++ b/pkg/slateparser/slate_parser.go
@@ -40,7 +40,7 @@ func CheckForMentions(text string, objectType string, objectID string, objectNam
 	matches := divTagRegex.FindAllStringSubmatch(text, -1)
 
 	for _, match := range matches {
-		if len(match) < 2 {
+		if len(match) < 2 { //nolint:mnd
 			continue
 		}
 


golanglemonade (approved):

pkg/slateparser/slate_parser.go
```diff
@@ -41,22 +36,36 @@ var (
...
-			displayName := match[3] // data-slate-value (user.DisplayName)
-
-			mentions[slateID] = Mention{
-				UserID:          userID,
-				UserDisplayName: displayName,
-				ObjectType:      objectType,
-				ObjectID:        objectID,
-				ObjectName:      objectName,
-			}
+		if len(match) < 2 {
```
golanglemonade: We can nolint this since its pretty well explained why there need to be more than 2 parts 
```suggestion
		if len(match) < 2 { //nolint:mnd
```

PR #1889, opened 2026-01-21
Organization delete should happen async

adelowo:
- org deletion now async in gala so the frontend does not wait indefinitely 
- register event listener in tests
- handle custom enum deletions being blocked ( see note below )
- updated tests

> When deleting an organization, the cascade deletion of the organization will try to delete `CustomTypeEnums` before other objects (assets, controls and others.) were deleted. The `HookCustomTypeEnumDelete` checks if global enums are in use by other records - since those objects still existed, it blocked the deletion with error "enum is in use". So introduced a context check to skip the global usage check during organisation edge deletion


internal/graphapi/tools_test.go
```diff
@@ -250,6 +250,12 @@ func (suite *GraphTestSuite) SetupSuite(t *testing.T) {
 	db, err := entdb.NewTestClient(ctx, suite.tf, jobOpts, opts)
 	requireNoError(t, err)
 
+	eventer := hooks.NewEventerPool(db)
```
golanglemonade: you should be able to add 

```
entdb.WithEventer(eventer),
```

to the `opts` for ent. 

And update this comment in the `NewTestClient` function - this is okay (and needed here) I had just turned it off because of all the subscriptions logs for duplicate ids showing up in test logs. 
```
	// Do not enable eventer or metrics in test clients
	// to avoid polluting test output
	clientOpts := []Option{
		WithModules(),
	}
```

internal/graphapi/tools_test.go
```diff
@@ -409,6 +415,26 @@ func requireNoError(t *testing.T, err error) {
 	}
 }
 
+func waitForCondition(t *testing.T, condition func() bool, msg string) {
+	t.Helper()
+
+	timeout := 5 * time.Second
```
golanglemonade: You think we can do this shorter? Maybe a minute or two? 5 minutes in test feels like a lifetime

golanglemonade (commented):
nice! Did you happen to seed an org and test deleting and make sure everything was cleaned up? You should be able to use harmonize to seed a bunch of data if not and see whats still there.

internal/graphapi/tools_test.go
adelowo: it is 5 seconds, do I still need to reduce it?

adelowo:
> nice! Did you happen to seed an org and test deleting and make sure everything was cleaned up? You should be able to use harmonize to seed a bunch of data if not and see whats still there.

yep tested correctly. edges are cleaned up correctly

internal/graphapi/tools_test.go
golanglemonade: 🤦‍♀️ no, lol.

internal/ent/hooks/customenums.go
```diff
@@ -249,6 +250,12 @@ func HookCustomTypeEnumDelete() ent.Hook {
 				return next.Mutate(ctx, m)
 			}
 
+			// skip the "in use" error/check when deleting via organization cascade
+			// the organization edge cleanup would deletion order properly via cascades
```
golanglemonade:
```suggestion
			// the organization edge cleanup needs to cascade deletes
```

golanglemonade (commented):
Looks good! Only error I saw while testing related to trust center request nda tuple deletion 🚀

adelowo:
closing this as it is being rolled into https://github.com/theopenlane/core/pull/2300

PR #1896, opened 2026-01-22
Campaigns, Platforms, and Identity Holders (SCI related feature request)

matoszz:
This PR primarily adds new ent schemas for campaign, campaign_target, identity_holder, and platform, plus their history schemas. Additionally adds their graphql resolvers, schemas, queries, and the associated generated output. 

The outcome(s) / primitives we intend to achieve with these schemas is, at a high level:
- Systems / Assets (scope boundaries) == `platform` schema
- People (employees, contractors, reviewers) == `identity_holder` schema
- Entities (vendors, subprocessors, providers) == `entities` schema
- Assessments & attestations (human acknowledgement) == `campaigns` and `campaign_targets` schema
- Time & cadence (when something was reviewed, scanned, trained, or accepted) == `campaigns` and `campaign_targets` schema

These will be used to provide user-facing solutions oriented towards Personnel management, vendor management, awareness training, vulnerability management, and asset management / asset inventory.  In addition to the schemas:

- Schema relationship updates to wire these into existing objects (assets/entities/assessments/workflows/users/scan/etc.), including new edge names, ownership fields, and workflow references
- New FGA model and tests for the added entities
- Added the ability to define a single shared enum like: `CustomTypeEnum { object_type: "", field: "environment", name: "Production" }` so that any schema with `newCustomEnumMixin(..., withEnumFieldName("environment"), withGlobalEnum())` can use it without duplicating enum records per object type (basically a `CustomTypeEnum` value with an empty `object_type` so that it's shared across schemas)

### DIFF
--- a/internal/ent/hooks/customenums.go
+++ b/internal/ent/hooks/customenums.go
@@ -105,7 +105,6 @@ func IsValidEnumField(objectType, field string) bool {
 	return findTableWithColumn(columnName) != nil
 }
 
-
 // HookCustomTypeEnumCreate validates that the object_type and field combination is valid
 func HookCustomTypeEnumCreate() ent.Hook {
 	return hook.On(func(next ent.Mutator) ent.Mutator {
--- a/internal/ent/hooks/customenums_test.go
+++ b/internal/ent/hooks/customenums_test.go
@@ -88,7 +88,6 @@ func TestIsValidEnumField(t *testing.T) {
 	})
 }
 
-
 func TestFindTablesWithColumn(t *testing.T) {
 	t.Run("finds tables with environment_id column", func(t *testing.T) {
 		tables := findTablesWithColumn("environment_id")
--- a/internal/ent/schema/mixin_customenums.go
+++ b/internal/ent/schema/mixin_customenums.go
@@ -152,4 +152,3 @@ func (c CustomEnumMixin) getEnumReverseRefName() string {
 
 	return fmt.Sprintf("%s_%s", sch.Name(), c.fieldName)
 }
-


fga/model/model.fga
```diff
@@ -474,6 +486,50 @@ type asset
...
+type platform
+  relations
+    define can_view: [user, service] or can_edit or (viewer but not blocked)
+    define can_edit: [user, service] or (editor but not blocked)
+    define can_delete: [user, service] or (editor but not blocked)
+    define audit_log_viewer: ([user, service] or audit_log_viewer from parent) and can_view
+    define editor: [group#member, organization#owner] or can_edit from parent
+    define viewer: [group#member] or editor or can_view from parent or member from parent
+    define blocked: [user, group#member]
+    define parent: [user, service, organization, entity]
```
golanglemonade: comment mismatch between defined parent (also includes entity) vs. comment text

fga/tests/tests.yaml
```diff
@@ -35,6 +35,10 @@ tests:
           can_create_standard: true
           can_create_template: true
           can_create_evidence: true
+          can_create_system: true
+          can_create_identity-holder: true
```
golanglemonade:
```suggestion
          can_create_identity_holder: true
```

fga/tests/tests.yaml
```diff
@@ -66,6 +70,10 @@ tests:
           can_create_standard: false
           can_create_template: false
           can_create_evidence: true # group tuple set to allow
+          can_create_system: false
+          can_create_identity-holder: false
```
golanglemonade:
```suggestion
          can_create_identity_holder: false
```

fga/tests/tests.yaml
```diff
@@ -95,6 +103,10 @@ tests:
           can_create_scheduled_job: true
           can_create_template: true
           can_create_evidence: true
+          can_create_system: true
+          can_create_identity-holder: true
```
golanglemonade:
```suggestion
          can_create_identity_holder: true
```

golanglemonade: ok theres a lot more than I though...these are all wrong :-D

go.mod
```diff
@@ -10,6 +10,8 @@ tool (
 	gotest.tools/gotestsum
 )
 
+replace github.com/theopenlane/core/common => /Users/manderson/newcore/core/common
```
golanglemonade: no one else has manderson's laptop :-D 

If you want this though we can do this I believe:
```suggestion
replace github.com/theopenlane/core/common => ./common
```

internal/ent/hooks/customenums.go
```diff
@@ -123,7 +143,11 @@ func HookCustomTypeEnumDelete() ent.Hook {
 
 			funcs := make([]func(), 0)
 			for _, enum := range enums {
-				funcs = append(funcs, isEnumInUse(ctx, client, enum.ID, strings.ToLower(enum.ObjectType), enum.ObjectType, enum.Name, &errs, &mu))
+				if enum.ObjectType == "" {
+					continue
+				}
```
golanglemonade: We don't want to make sure the enum is not in use if its a global enum before deleting?

internal/ent/hooks/customenums.go
```diff
@@ -157,9 +181,13 @@ func isEnumInUse(ctx context.Context, client *generated.Client, enumID, edgeName
 		label string
 	}
 
-	table := pluralize.NewClient().Plural(edgeName)
+	edgeName := strings.ToLower(objectType)
+	if enumField == "" {
```
golanglemonade, line 273: good catch, I bet this was a bug with risks

internal/ent/schema/asset.go
```diff
@@ -50,40 +65,189 @@ func (Asset) Fields() []ent.Field {
...
+			Optional().
+			Annotations(
+				entgql.OrderField("region"),
+			),
+		field.String("security_tier").
+			Comment("security tier assigned to the asset").
+			Optional().
+			Annotations(
+				entgql.OrderField("security_tier"),
+			),
```
golanglemonade: does this need to be a an enum or custom enum instead of just a string to ensure consistency?

internal/ent/schema/asset.go
```diff
@@ -50,40 +65,189 @@ func (Asset) Fields() []ent.Field {
...
+			Optional().
+			Annotations(
+				entgql.OrderField("access_model"),
+			),
+		field.String("encryption_status").
+			Comment("encryption status for data associated with the asset").
+			Optional().
+			Annotations(
+				entgql.OrderField("encryption_status"),
+			),
```
golanglemonade: Same question here about enums? Ok if the answer is just "we will do that later, but before we ship the product"

internal/ent/schema/asset.go
```diff
@@ -50,40 +65,189 @@ func (Asset) Fields() []ent.Field {
...
+				entgql.QueryField(),
+				entgql.MultiOrder(),
+			),
+		edge.From("connected_by_assets", Asset.Type).
+			Ref("connected_assets").
+			Annotations(
+				entgql.RelayConnection(),
+				entgql.QueryField(),
+				entgql.MultiOrder(),
+			),
```
golanglemonade: Any reason these aren't using the helper functions?

internal/ent/schema/campaign.go
```diff
@@ -0,0 +1,343 @@
...
+				return c.Validate()
+			}).
+			Optional().
+			Nillable(),
+		field.String("recurrence_timezone").
+			Comment("timezone used for the recurrence schedule").
+			Optional().
+			Annotations(
+				entgql.OrderField("recurrence_timezone"),
+			),
```
golanglemonade, line 171: we are going to want a validator on here - you should be able to use https://pkg.go.dev/time#LoadLocation

internal/ent/schema/campaign.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// Mixin of the Campaign
+func (c Campaign) Mixin() []ent.Mixin {
+	return mixinConfig{
+		prefix: "CMP",
+		additionalMixins: []ent.Mixin{
+			newObjectOwnedMixin[generated.Campaign](c,
+				withParents(Organization{}),
+				withOrganizationOwner(true),
```
golanglemonade, line 242: this doent really matter, but I dont _think_ you need both here. they basically end up being the same thing. Although it might be an edge case when you want to start with object owned but only have the org for now. I know that you don't have to specifically set withParents( Organization{}) if you have the `withOrganizationOwner(true)`

internal/ent/schema/campaign_target.go
```diff
@@ -0,0 +1,206 @@
...
+			Comment("the email address targeted by the campaign").
+			NotEmpty().
+			Annotations(
+				entx.FieldSearchable(),
+				entgql.OrderField("email"),
+			).
+			Validate(func(email string) error {
+				_, err := mail.ParseAddress(email)
+				return err
+			}),
```
golanglemonade, line 74: if I have a user or contact associated are we going to add a hook to auto set some of these fields like email, name, etc?

internal/ent/schema/entity.go
```diff
@@ -91,6 +93,202 @@ func (Entity) Fields() []ent.Field {
...
+			Annotations(
+				entgql.OrderField("auto_renews"),
+			),
+		field.Int("termination_notice_days").
+			Comment("number of days required for termination notice").
+			Optional().
+			Annotations(
+				entgql.OrderField("termination_notice_days"),
+			),
+		field.Float("annual_spend").
```
golanglemonade, line 151: if this is going to be displayed in the UI we should include an `spend_currency` field + default to USD

internal/ent/schema/entity.go
```diff
@@ -91,6 +93,202 @@ func (Entity) Fields() []ent.Field {
...
+			Comment("annual spend associated with the entity").
+			Optional().
+			Annotations(
+				entgql.OrderField("annual_spend"),
+			),
+		field.String("billing_model").
+			Comment("billing model for the entity relationship").
```
golanglemonade, line 165: is this annual/monthly? invoice/credit card, or something different?

internal/ent/schema/entity.go
```diff
@@ -91,6 +93,202 @@ func (Entity) Fields() []ent.Field {
...
+
+				return validator.ValidateURL()(u)
+			}).
+			Optional().
+			Annotations(
+				entgql.OrderField("status_page_url"),
+			),
+		field.String("source_platform_id").
+			Comment("the platform that sourced the entity record").
+			Optional(),
```
golanglemonade: This would mean a entity can only be associated with one platform. Feels like you could have multiple?

internal/ent/schema/entity.go
```diff
@@ -91,6 +93,202 @@ func (Entity) Fields() []ent.Field {
...
+			Annotations(
+				entgql.OrderField("internal_owner"),
+				entx.FieldSearchable(),
+			),
+		field.String("internal_owner_user_id").
+			Comment("the internal owner user id for the entity").
+			Optional(),
+		field.String("internal_owner_group_id").
+			Comment("the internal owner group id for the entity").
+			Optional(),
```
golanglemonade: I've seen these fields a ton on different schemas - would it make sense to make a mixin for these and their edges? especially if we are going to do anything with hooks, etc. with them it would be easier to manage

internal/ent/schema/identity_holder.go
```diff
@@ -0,0 +1,323 @@
...
+			edgeSchema: WorkflowObjectRef{},
+			name:       "workflow_object_refs",
+			ref:        "identity_holder",
+		}),
+		edge.To("access_platforms", Platform.Type).
+			Annotations(
+				entgql.RelayConnection(),
+				entgql.QueryField(),
+				entgql.MultiOrder(),
+			),
```
golanglemonade: this should be able to just use `edgeToWithPagination...` helper?

golanglemonade (commented):
Looks good overall - few comments here + in slack; only this else was assume we are following up with tests + implementing functions with panics right now? :)

internal/ent/schema/campaign_target.go
matoszz: that was the intent, yea

internal/ent/schema/entity.go
matoszz: was intended to be generic, e.g. "per user" "per seat" "unlimited" "the 9 million other variants"

matoszz:
@golanglemonade I pushed the alternate branch I re-created after your feedback and it overwrite the commit history that existed in here. All the original descriptions are still the same, with some color:

- Addressed all PR feedback in terms of edges needing helpers, the use of EdgeAuthCheck needing to be EdgeViewCheck, fixed FGA tests
- Incorporated feedback / requests around various fields being better suited as enums and moved them accordingly and removed from source schemas
- Additionally took feedback around repeated field usage surrounding "internal_owner" and associated fields, turned into a "Responsibility" mixin with various toggles, options, etc., updated all schemas that used those fields to instead use the mixin
- Replaced hardcoded validObjectTypes map with runtime introspection of migrate.Tables. Empty object_type now valid
- When AllowGlobal: true, tries global lookup first, falls back to object-scoped
- Now handles global enums by querying all tables that reference the enum column (via UNION), not just the object's table

PR #1900, opened 2026-01-23
Implement NDA requests flow based off trustcenter settings

adelowo:
- added hook to nda request
- hook checks for nda requirement settings then adds to approval queue or actually send the email
- if email previously approved, just send the email directly to them
- dropped existing paths and queries to send the request email


internal/graphapi/trustcenterndarequest_test.go
```diff
@@ -0,0 +1,445 @@
...
+
+	testCases := []struct {
+		name           string
+		input          testclient.CreateTrustCenterNDARequestInput
+		client         *testclient.TestClient
+		ctx            context.Context
+		expectedErr    string
+		expectedStatus enums.TrustCenterNDARequestStatus
+	}{
+		{
```
golanglemonade, line 37: Can you add a test for anon Trust Center user since that will be the most common create test?

PR #1903, opened 2026-01-24
Workflows: Action Parameters, Pre-filtering, Domain/Approval grouping

matoszz:
This PR encompasses a large expansion of the root workflows package, in addition to some tweaks to the existing code; the `WorkflowEvent` and `WorkflowObjectRef` schemas were updated to be constrained further by adding skip annotations for create and update mutations, removing the generated resolvers, removing the graphql queries, removing the graphql schemas associated. This narrows the exposed surface for workflows and ensures we have clear entry points.

Workflow package features:

- Domain/approval grouping: derive stable domain keys, split proposed changes by approval domains, compute proposal hashes, and enumerate approval domains from workflow definitions
- Event emission: wrap/emit workflow events, capture enqueue results, and serialize/parse structured failure details for deterministic re‑emission
- Field eligibility: compute eligible workflow fields per object type and filter changed fields to those eligible
- Object reference helpers: query builders and ID lookup for workflow object references
- Shared helpers: build proposed change maps, match workflow triggers to events, resolve owner IDs from context, build action contexts, find proposals by object refs/state priority, and validate CEL expressions
- Action parameter structs for approvals, notifications, webhooks, and integrations
- Trigger prefiltering to support efficient evaluation based on operations and fields/edges
- String normalization utility for target/field lists
- Comprehensive new tests covering the above, plus coverage utilities for fakes/no‑op drivers
- Context helpers refactored to share allow/bypass‑with‑org logic
- Error definitions expanded to cover emit failures, nil clients, missing org IDs, and string field parsing errors
- String field extraction now returns errors and includes a string‑slice variant; typed‑nil detection uses JSON “null” sentinel (no reflection).
- Generated helpers now return explicit errors on nil client / missing org ID
- Added observability fields for workflow objects
- Resolver wiring updated to handle the new error‑returning string helpers

Misc updates associated with the work:
- The `GenericMutation` interface was expanded with a few additional fields (e.g. `ClearedFields, Field` etc.)
- pre-commit updates
- Add workflowengine any type to ent/config.tmpl
- Tweaked workflow event enums + regenerated output
- Workflow package README.md

### DIFF
--- a/internal/workflows/options_test.go
+++ b/internal/workflows/options_test.go
@@ -43,7 +43,7 @@ func TestConfigOptions(t *testing.T) {
 	override := Config{
 		Enabled: false,
 		CEL: CELConfig{
-			Timeout:  time.Second,
+			Timeout:   time.Second,
 			CostLimit: 12,
 		},
 	}
--- a/internal/workflows/shared_helpers.go
+++ b/internal/workflows/shared_helpers.go
@@ -28,7 +28,7 @@ type WorkflowCreationError struct {
 	// Stage is the creation stage that failed
 	Stage WorkflowCreationStage
 	// Err is the underlying error
-	Err   error
+	Err error
 }
 
 // Error formats the workflow creation error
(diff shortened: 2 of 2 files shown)


internal/workflows/domain_test.go
```diff
@@ -0,0 +1,435 @@
...
+				Params: secondaryParams,
+			},
+		},
+	}
+
+	domains, err := ApprovalDomains(doc)
+	require.NoError(t, err)
+
+	assert.Equal(t, [][]string{
+		{" status ", "name"},
```
golanglemonade, line 275: should this have a leading space? 
```suggestion
		{" status ", "name"},
```

oh I guess the input does so thats why it passes haha

PR #1928, opened 2026-01-28
Campaigns: WithTargets, Launch Campaign, Send to Incomplete, Send Test Email

matoszz:
Adds additional campaign capabilities for creating campaign targets along with campaigns in bulk upload, extends the graphQL API for "launching" a campaign (so that you can draft it without sending), capabilities for sending follow ups for overdue / incomplete, functionality to send a test email so you could view a sample of the campaign without polluting the list or analytics. Additionally it pulls in latest `entx` + `gqlgen-plugins` and bulk modifies in-place graphql schemas along with their resolvers. Configuration was updated to false after the modify run. 

- Extended campaign system with bulk dispatch capabilities for sending assessment emails to targets, ability to create campaigns with targets in a single operation, and test email functionality before launching                                                                                                                                
- Refactored assessment response hook for idempotency with race condition handling via constraint violation detection, automatic status reset when resending overdue responses, due date propagation on resend, and campaign target status synchronization                                                                                    
- Enhanced CSV bulk upload with generated CSV input types, foreign key reference resolution during imports, and improved JSON validation error messages for map fields
- GraphQL schema extensions including new bulk create mutations across many entities, extended workflow instance queries, and campaign dispatch/test operations          
- Database migrations adding indexes for campaign and assessment response queries                                   
- Minor fixes for CSV map field error messaging


internal/graphapi/campaign_dispatch.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	if subjectID, err := auth.GetSubjectIDFromContext(ctx); err == nil {
+		args.RequestedBy = subjectID
+	}
+
+	opts := args.InsertOpts()
+	opts.ScheduledAt = *scheduledAt
+
+	_, err := r.db.Job.Insert(ctx, args, &opts)
```
golanglemonade, line 369: Do you have a riverboat jobs PR for the job definition?

internal/graphapi/campaign_dispatch.go
```diff
@@ -0,0 +1,445 @@
...
+	if auth.IsSystemAdminFromContext(ctx) {
+		return nil
+	}
+
+	subjectID, err := auth.GetSubjectIDFromContext(ctx)
+	if err != nil {
+		return parseRequestError(ctx, err, common.Action{Action: common.ActionGet, Object: "user"})
+	}
+
+	allow, err := r.db.Authz.CheckAccess(ctx, fgax.AccessCheck{
```
golanglemonade, line 431: Is there a reason this is done in a hook instead of a policy on the schema?

internal/graphapi/campaign_with_targets.go
```diff
@@ -0,0 +1,86 @@
...
+		campaignInput.RecipientCount = &targetCount
+	}
+
+	res, err := withTransactionalMutation(ctx).Campaign.Create().SetInput(*campaignInput).Save(ctx)
+	if err != nil {
+		return nil, parseRequestError(ctx, err, common.Action{Action: common.ActionCreate, Object: "campaign"})
+	}
+
+	var createdTargets []*generated.CampaignTarget
+	if len(targets) > 0 {
```
golanglemonade: In `CreateCampaignWithTargetsCSV` you return an error if there are no targets, but on this path it seems like thats okay? Should be consistent between the resolver paths.

internal/graphapi/common/helpers.go
```diff
@@ -281,6 +292,429 @@ func normalizeCSVEnumInputs(data any) {
...
+		kind := fieldType.Kind()
+		if kind == reflect.Pointer {
+			kind = fieldType.Elem().Kind()
+		}
+		if kind == reflect.Struct {
+			collectCSVHeaderNames(fieldType, headers, combined)
+			continue
+		}
+
+		for _, name := range combined {
```
golanglemonade, line 574: This function looks basically the same as `collectCSVListFieldKinds` except this last range; can we reuse it?

internal/graphapi/common/helpers.go
```diff
@@ -281,6 +292,429 @@ func normalizeCSVEnumInputs(data any) {
...
+		names := make([]string, 0, len(parts))
+		for _, part := range parts {
+			part = strings.TrimSpace(part)
+			if part == "" {
+				continue
+			}
+			if part == "omitempty" || part == "partial" || strings.HasPrefix(part, "default=") {
+				continue
+			}
+			names = append(names, part)
```
golanglemonade: Hm, I thought `field.Tag.Get("csv")` returns the key value, e.g. something like

```
 `csv:"name,omitempty"`
 ```
 
 `default` would be it's own tag. The only thing that would be a part is `omitempty`, I think? 
 
 Why skip `omitempty` empty? Should still be included - just would mean its not required.

File: internal/graphapi/campaign_with_targets.go (hunk 3 above)
Comment (matoszz) [reply internal/graphapi/campaign_with_targets.go]: good call; shouldn't silently allow using `withTargets` when no targets are provided. Will update.

File: internal/graphapi/common/helpers.go [hunk 6]
```diff
@@ -281,6 +292,429 @@ func normalizeCSVEnumInputs(data any) {
...
+			headers[normalizeCSVHeader(name)] = struct{}{}
+		}
+	}
+}
+
+// csvFieldNames extracts CSV header names from struct tags or field names.
+func csvFieldNames(field reflect.StructField) []string {
+	tag := strings.TrimSpace(field.Tag.Get("csv"))
+	if tag != "" {
+		parts := strings.Split(tag, ",")
```
golanglemonade: nit: return early for readability instead of the majority the function in an `if` :

```suggestion
func csvFieldNames(field reflect.StructField) []string {
	tag := strings.TrimSpace(field.Tag.Get("csv"))
	if tag == "" {
	   return []string{field.Name}
	}
	
	parts := strings.Split(tag, ",")
	....
```

internal/graphapi/common/helpers.go
matoszz: not directly but I can refactor for both functions to use a shared helper - will do that

internal/graphapi/csv_lookup_registry.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return nil, common.NewValidationError("organization id not found in context")
+		}
+
+		return fn(ctx, client, orgID, values)
+	}
+}
+
+// wrapGeneratedCreate wraps a generated create function to get client and orgID from context.
+func wrapGeneratedCreate(fn csvgenerated.CSVCreateFn) csvCreateFn {
```
golanglemonade: the function returned here:

```
	return func(ctx context.Context, values []string) (map[string]string, error) {
		client := withTransactionalMutation(ctx)

		if err != nil {
			return nil, common.NewValidationError("organization id not found in context")
		}

		return fn(ctx, client, orgID, values)
	}
```

are exactly the same. Define it once and return the same thing?

internal/graphapi/common/helpers.go
matoszz: will remove the unnecessary partial and default= checks

internal/graphapi/csv_lookup_registry.go
matoszz: Sure, can do 👍

internal/graphapi/csv_reference.go
```diff
@@ -0,0 +1,710 @@
...
+
+// set stores resolved values in the cache.
+func (c *csvRuleLookupCache) set(key csvLookupCacheKey, resolved map[string]string) {
+	if len(resolved) > 0 {
+		c.cache[key] = resolved
+	}
+}
+
+// resolveCSVReferenceRules resolves lookup rules and writes IDs into target fields on the inputs.
+func resolveCSVReferenceRules(ctx context.Context, inputs any, rules ...CSVReferenceRule) error {
```
golanglemonade: looking at the usage of this function, I think you can be stricter about the `inputs` type to not be `any` and then require reflection. You should be able to do 
```suggestion
func resolveCSVReferenceRules[T any](ctx context.Context, inputs []*T, rules ...CSVReferenceRule) error {
```

All the `inputs` is `data` from the result of `UnmarshalBulkData`

internal/graphapi/evidence_test.go
```diff
@@ -583,9 +583,6 @@ func TestMutationCreateBulkCSVEvidence(t *testing.T) {
 	bulkFile, err := storage.NewUploadFile("testdata/uploads/evidence.csv")
 	assert.NilError(t, err)
 
-	invalidBulkFile, err := storage.NewUploadFile("testdata/uploads/evidence_invalid.csv")
```
golanglemonade, line 586: Is this file valid now? Might be good to update the test to confirm so?

internal/graphapi/campaign_dispatch.go
golanglemonade: chatted in slack, this is fine here as it short circuits early.  The first `create` action in the flow is an Assestment response and its possible someone would have access there but not to a campaign.

PR #1931, opened 2026-01-29
Migrate trustcenter subprocesors categories

adelowo:


internal/ent/schema/trustcentersubprocessor.go
```diff
@@ -61,10 +57,6 @@ func (TrustCenterSubprocessor) Fields() []ent.Field {
 		field.JSON("countries", []string{}).
 			Comment("country codes or country where the subprocessor is located").
 			Optional(),
-		field.String("category").
-			Comment("Category of the subprocessor, e.g. 'Data Warehouse' or 'Infrastructure Hosting'").
-			NotEmpty().
-			MaxLen(trustCenterSubprocessorCategoryMaxLen),
```
golanglemonade, line 67: can you leave this field and we can follow-up on another PR to remove. Otherwise we need to coordinate the release

PR #1932, opened 2026-01-29
Workflow test fixes

matoszz:
Expanded `ClearWorkflowDefinitions()` function to delete all workflow-related entities with some basic ordering to ensure we don't hit foreign key constraints. Changes should provide additional test isolation by ensuring we've cleared out any records prior to seeding the test data.


internal/workflows/engine/resolver_test.go
```diff
@@ -322,6 +330,8 @@ func (s *WorkflowEngineTestSuite) TestResolveControlTarget() {
 //   The resolver must validate target configurations and fail fast with clear errors
 //   rather than silently returning empty results or causing downstream failures.
 func (s *WorkflowEngineTestSuite) TestResolveInvalidTarget() {
+	s.ClearWorkflowDefinitions()
```
golanglemonade, line 333: Instead of putting this on every test you can use the  `AfterTest` or `BeforeTest` and just do it once in the tools_test file: https://pkg.go.dev/github.com/stretchr/testify/suite#AfterTest

```
// AfterTest runs after each test
func (s * WorkflowEngineTestSuite) AfterTest() {
	s.ClearWorkflowDefinitions()
}
```

matoszz: That would certainly be a more intelligent way to do it 😆

matoszz: @golanglemonade apparently this only runs on the master test setup and teardown, not on the individual subtests; so i think the way its structured now is correct as I still had to add the majority of these calls back into subtests even with the `BeforeTest` and `AfterTest`

PR #1933, opened 2026-01-29
feat: updates for nda requets

golanglemonade:
- Adds `approvedAt` and `signedAt` columns to trust center nda requests + updated hooks to set based on status
- adds documentdata as edge to nda requets
- no longer fails update if the cache refresh fails 
- adds `DECLINED` status for nda requests
- updates check for signed NDAs based on nda requests, instead of document data json parsing
- adds tests for emails sent for nda requests
- removes requirement for trust center doc to be linked to a standard, this should be optional
- removes max length check on overview of trust center
- adds bulk update resolvers for approval + deny of nda requests
- hopefully fixes flakey trust center tests by adding cleanup function


internal/ent/hooks/documentdata.go
```diff
@@ -92,6 +82,31 @@ func HookDocumentDataTrustCenterNDA() ent.Hook {
...
+
+				return nil, fmt.Errorf("unexpected type %T: %w", v, ErrInternalServerError)
+			}
+
+			// update nda requests that it has been signed
+			_, err = m.Client().TrustCenterNDARequest.Update().Where(
+				trustcenterndarequest.EmailEqualFold(anon.SubjectEmail),
+				trustcenterndarequest.TrustCenterID(anon.TrustCenterID),
+				trustcenterndarequest.StatusNEQ(enums.TrustCenterNDARequestStatusSigned),
+			).SetStatus(enums.TrustCenterNDARequestStatusSigned).SetDocumentDataID(createdDocData.ID).Save(ctx)
```
matoszz, line 98: if you don't use the return object why not use `Exec`?

PR #1940, opened 2026-01-30
Update document data resolver to allow for uploads

adelowo:
worker queue needs to be able to upload the pdf it has signed/attested after the nda request. so the resolver 
is now updated to fit that. But this is limited to system admins for now. 

The pre-existing way to add files to document data is to use `AddFileIDs` in mutations for regular users/tokens. that still exists. but this is specifically for admins. non admins can use the mutation still as usual, files will just not be processed .

this links the document data and nda signed data to the attested/signed file

### DIFF
--- a/internal/ent/hooks/documentdata.go
+++ b/internal/ent/hooks/documentdata.go
@@ -170,13 +170,17 @@ func HookDocumentDataFile() ent.Hook {
 				return nil, errMissingTemplate
 			}
 
-			_, err = m.Client().Template.Query().
+			exists, err := m.Client().Template.Query().
 				Select(template.FieldKind).
 				Where(template.KindEQ(enums.TemplateKindTrustCenterNda)).
 				Where(template.ID(id)).
-				Only(ctx)
+				Exist(ctx)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


internal/ent/hooks/documentdata.go
```diff
@@ -144,6 +146,56 @@ func HookDocumentDataTrustCenterNDA() ent.Hook {
...
+				return nil, errMissingTemplate
+			}
+
+			_, err = m.Client().Template.Query().
+				Select(template.FieldKind).
+				Where(template.KindEQ(enums.TemplateKindTrustCenterNda)).
+				Where(template.ID(id)).
+				Only(ctx)
+			if err != nil {
+				return nil, err
```
golanglemonade, line 179: This would error if the document data wasn't an NDA - any reason we'd want to prevent that? You can also use `Exists(ctx)` if you only care that it exists and not get the template back

adelowo: No reason really. I added the check because it's only used by the nda right now

golanglemonade (approved):

PR #1942, opened 2026-01-30
feat: add role to auth context, cleanup noisy logs on org owner

golanglemonade:
depends on: https://github.com/theopenlane/iam/pull/426

- Adds `OrganizationRole` to AuthContext in order to skip checks on org level access that are soley based on read/write to the organization.
- Also removes error level logging for org level checks, sets to debug - and sets to info for any other relation - that we still have the data but its not really an error in many cases because of the use of the function.

### DIFF
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/spf13/cobra v1.10.2
 	github.com/theopenlane/core v1.5.15
 	github.com/theopenlane/core/common v1.0.6
-	github.com/theopenlane/go-client v0.6.2
+	github.com/theopenlane/go-client v0.7.1
 	github.com/theopenlane/httpsling v0.3.0
 	github.com/theopenlane/iam v0.23.6
 	github.com/theopenlane/utils v0.6.2
@@ -35,7 +35,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/clipperhouse/displaywidth v0.6.2 // indirect
 	github.com/clipperhouse/stringish v0.1.1 // indirect
-	github.com/clipperhouse/uax29/v2 v2.3.1 // indirect
+	github.com/clipperhouse/uax29/v2 v2.4.0 // indirect
 	github.com/danieljoos/wincred v1.2.3 // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.4.0 // indirect
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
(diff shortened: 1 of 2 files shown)


pkg/middleware/auth/auth.go
```diff
@@ -626,6 +644,40 @@ func isPATSSOAuthorized(ctx context.Context, db *ent.Client, tokenID, orgID stri
...
+		return nil
+	}
+
+	role, err := getRole(ctx, db, userID, orgID)
+	if err != nil {
+		logx.FromContext(ctx).Debug().Err(err).Msg("failed to get organization role, proceeding with other auth checks")
+		return nil
+	}
+
+	orgRole, ok := auth.ToOrganizationRoleType(strcase.SnakeCase(role.String()))
```
matoszz: just confirming you want snakecase here?

pkg/middleware/auth/auth.go
```diff
@@ -626,6 +644,40 @@ func isPATSSOAuthorized(ctx context.Context, db *ent.Client, tokenID, orgID stri
...
+	if !ok {
+		logx.FromContext(ctx).Debug().Str("role", role.String()).Msg("invalid organization role, proceeding with other auth checks")
+	}
+
+	return &orgRole
+}
+
+func getRole(ctx context.Context, db *ent.Client, userID, orgID string) (enums.Role, error) {
```
matoszz, line 669: can't you limit the fields returned and have the query itself as the return? not important just a suggestion

pkg/middleware/auth/token_sso_test.go
```diff
@@ -79,6 +85,11 @@ func TestPATTokenSSOAuthorization(t *testing.T) {
 		return false, nil
 	}
 
+	getOrgRoleFunc = func(ctx context.Context, db *ent.Client, userID, orgID string) *auth.OrganizationRoleType {
+		// return nil, sso doesn't depend on role
```
matoszz, line 89: doesn't the SSO bypass depend on being an owner role?

pkg/middleware/auth/auth.go
golanglemonade: Actually I don't need that anymore because I updated the function to handle that instead. I'll update

pkg/middleware/auth/auth.go
golanglemonade: You sound like me 😂 yep can do

pkg/middleware/auth/token_sso_test.go
golanglemonade: If this is nil, the code path stays the same and falls back to fga checks. This was in the PAT code path to set the value but it's not checked this way. But I should go see if I can update/remove some db queries checking for roles.

PR #1944, opened 2026-01-30
Allow system admin/impersonation to update documentdata

adelowo:
- **check for impersonation context and allow system admin**
- **Fix merge conflicts**


internal/ent/hooks/documentdata.go
```diff
@@ -162,7 +162,9 @@ func HookDocumentDataFile() ent.Hook {
 			}
 
 			if !auth.IsSystemAdminFromContext(ctx) {
-				return nil, generated.ErrPermissionDenied
+				if user, ok := auth.SystemAdminFromContext(ctx); ok && !user.IsSystemAdmin {
+					return nil, generated.ErrPermissionDenied
+				}
```
golanglemonade, line 169: isn't this just the same check again?

internal/ent/schema/documentdata.go
```diff
@@ -55,6 +56,7 @@ func (d DocumentData) Mixin() []ent.Mixin {
 	return mixinConfig{
 		additionalMixins: []ent.Mixin{
 			newObjectOwnedMixin[generated.DocumentData](d,
+				withSkipForSystemAdmin(true),
```
golanglemonade: This allows the `owner_id` field to be empty - is that what you want here? Or just allowing a system admin to make updates?

internal/ent/hooks/documentdata.go
golanglemonade: wow, those are named way too close, I see the difference now

PR #1955, opened 2026-02-02
Notification Preferences/Templates, Email Templates/Branding, Vulnerability Updates

matoszz:
Adding several new + complimentary schemas to enhance campaigns capability, workflows,  integrations. Email branding and templating can be re-used outside of campaigns capability (intend to also plumb it into workflows). Notification preferences / templates are being added in anticipation of workflows and built-in alerting capabilities not becoming too noisy. Extending / enhancing the vulnerabilities capability as a part of SCI work. 

- `EmailBranding`: Organization-owned email branding config (colors, logo, fonts) with edges to Campaign and EmailTemplate 
- `EmailTemplate`: Versioned email templates with subject/body/preheader fields, JSON schema support, and links to EmailBranding, Integration, and Workflow entities
- `NotificationPreference`: Per-user channel preferences with topic routing, delivery scheduling (quiet hours, mute), and self-access policy
- `NotificationTemplate`: Channel-specific templates (email, Slack, Teams) with topic pattern matching, structured blocks, and workflow integration
- `IntegrationRun`: Internal execution history for integrations with timing, payload file references, and metrics (not exposed via GraphQL)
- `IntegrationWebhook`: Internal webhook endpoint config with auto-generated secrets and delivery tracking (not exposed via GraphQL)
- `Integration`: Added unique constraint on (owner, kind), provider metadata JSON, searchable name/kind, and edges to new template and webhook schemas
- `WorkflowDefinition`: Added edges to `NotificationTemplate` and `EmailTemplate`                                      
- `WorkflowInstance`: Added edge to `EmailTemplate`                         
- `WorkflowAssignment`: Added `due_at` timestamp for delegation/escalation scheduling
- `resolveVulnerability` resolver: Sets status to resolved and optionally creates a linked Review record                                                          
- `vulnerabilitySummary`: Aggregation query returning counts by severity, source, and status for dashboard widgets                                                         
- `WorkflowAdmin`: Force-complete, cancel (single and bulk), and reassign operations for workflow instances/assignments                                                                               
- `WorkflowProposal`: Full lifecycle management (query, update changes, submit, withdraw) with preview/diff support

### DIFF
--- a/internal/graphapi/workflowadmin_helpers.go
+++ b/internal/graphapi/workflowadmin_helpers.go
@@ -117,7 +117,7 @@ func closeWorkflowAssignments(ctx context.Context, client *generated.Client, ins
 
 	allowCtx := workflows.AllowContext(ctx)
 	if ownerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(allowCtx, ownerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(allowCtx, &ownerID); err != nil {
 			return err
 		}
 	}
@@ -195,7 +195,7 @@ func (r *mutationResolver) forceCompleteWorkflowInstance(ctx context.Context, id
 	workflows.MarkSkipEventEmission(skipCtx)
 
 	if instance.OwnerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(skipCtx, instance.OwnerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(skipCtx, &instance.OwnerID); err != nil {
 			return nil, err
 		}
 	}
@@ -280,7 +280,7 @@ func (r *mutationResolver) cancelWorkflowInstance(ctx context.Context, id string
 	workflows.MarkSkipEventEmission(skipCtx)
 
 	if instance.OwnerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(skipCtx, instance.OwnerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(skipCtx, &instance.OwnerID); err != nil {
 			return nil, err
 		}
 	}
--- a/internal/graphapi/workflowproposalextended.resolvers.go
+++ b/internal/graphapi/workflowproposalextended.resolvers.go
@@ -61,7 +61,7 @@ func (r *mutationResolver) UpdateWorkflowProposalChanges(ctx context.Context, in
 	}
 
 	if proposal.OwnerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(allowCtx, proposal.OwnerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(allowCtx, &proposal.OwnerID); err != nil {
 			return nil, err
 		}
 	}
@@ -114,7 +114,7 @@ func (r *mutationResolver) SubmitWorkflowProposal(ctx context.Context, id string
 	}
 
 	if proposal.OwnerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(allowCtx, proposal.OwnerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(allowCtx, &proposal.OwnerID); err != nil {
 			return nil, err
 		}
 	}
@@ -175,7 +175,7 @@ func (r *mutationResolver) WithdrawWorkflowProposal(ctx context.Context, id stri
 	}
 
 	if proposal.OwnerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(allowCtx, proposal.OwnerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(allowCtx, &proposal.OwnerID); err != nil {
 			return nil, err
 		}
 	}
@@ -284,7 +284,7 @@ func (r *queryResolver) WorkflowProposalsForObject(ctx context.Context, objectTy
 		return nil, parseRequestError(ctx, err, common.Action{Action: common.ActionGet, Object: "workflowobject"})
 	}
 	if ownerID != "" {
-		if err := auth.SetOrganizationIDInAuthContext(allowCtx, ownerID); err != nil {
+		if err := common.SetOrganizationInAuthContext(allowCtx, &ownerID); err != nil {
 			return nil, err
 		}
 	}


internal/ent/schema/emailbranding.go
```diff
@@ -0,0 +1,136 @@
...
+			Comment("friendly name for this email branding configuration").
+			NotEmpty().
+			MaxLen(nameMaxLen).
+			Annotations(
+				entx.FieldSearchable(),
+				entgql.OrderField("name"),
+			),
+		field.String("brand_name").
+			Comment("brand name displayed in templates").
+			MaxLen(trustCenterNameMaxLen).
```
golanglemonade: `trustCenterNameMaxLen` ?

internal/ent/schema/emailbranding.go
```diff
@@ -0,0 +1,136 @@
...
+			Comment("button text color for emails").
+			Validate(validator.HexColorValidator).
+			Optional(),
+		field.String("link_color").
+			Comment("link color for emails").
+			Validate(validator.HexColorValidator).
+			Optional(),
+		field.String("font_family").
+			Comment("font family for emails").
+			Optional(),
```
golanglemonade: Should probably be an enum?

```suggestion
		field.Enum("font_family").
			Comment("font family for emails").
			GoType(enums.Font("")).
			Default(enums.FontHelvetica.String()).
			Optional(),
```

internal/ent/schema/integrationrun.go
```diff
@@ -0,0 +1,200 @@
...
+			Comment("integration connection this run belongs to").
+			Optional(),
+		field.String("operation_name").
+			Comment("operation identifier executed for this run").
+			Optional().
+			Annotations(
+				entgql.OrderField("OPERATION_NAME"),
+			),
+		field.String("operation_kind").
+			Comment("operation category executed for this run").
```
golanglemonade, line 63: should this be an enum? I see this isn't a schema with resolvers so less of an issue but seems like a `kind` should still be typed?

internal/ent/schema/integrationrun.go
```diff
@@ -0,0 +1,200 @@
...
+			Annotations(
+				entgql.OrderField("OPERATION_KIND"),
+			),
+		field.String("run_type").
+			Comment("run type such as RUN or SYNC").
+			Optional().
+			Annotations(
+				entgql.OrderField("RUN_TYPE"),
+			),
+		field.String("status").
```
golanglemonade: same here? status should be an enum?

internal/ent/schema/integrationwebhook.go
```diff
@@ -0,0 +1,179 @@
...
+		field.String("integration_id").
+			Comment("integration connection this webhook belongs to").
+			Optional(),
+		field.String("name").
+			Comment("display name for the webhook endpoint").
+			Optional().
+			Annotations(
+				entgql.OrderField("NAME"),
+			),
+		field.String("status").
```
golanglemonade: status should be an enum?

internal/ent/schema/notification.go
```diff
@@ -74,6 +74,11 @@ func (Notification) Fields() []ent.Field {
 			Optional().
 			Immutable().
 			Annotations(entgql.Skip(entgql.SkipMutationUpdateInput)),
+		field.String("template_id").
```
golanglemonade, line 77: should this be an edge to the notification template table?

internal/ent/schema/notificationpreference.go
```diff
@@ -0,0 +1,195 @@
...
+		field.Enum("status").
+			Comment("status of the channel configuration").
+			GoType(enums.NotificationChannelStatus("")).
+			Default(enums.NotificationChannelStatusEnabled.String()).
+			Annotations(
+				entgql.OrderField("STATUS"),
+			),
+		field.String("provider").
+			Comment("provider for the channel, e.g. slack, email, teams").
+			Optional(),
```
golanglemonade, line 68: What is the difference between this and channel which is `slack`, `email`, `teams`?

internal/ent/schema/notificationpreference.go
```diff
@@ -0,0 +1,195 @@
...
+			Nillable(),
+		field.String("quiet_hours_start").
+			Comment("start of quiet hours in HH:MM").
+			Optional(),
+		field.String("quiet_hours_end").
+			Comment("end of quiet hours in HH:MM").
+			Optional(),
+		field.String("timezone").
+			Comment("timezone for quiet hours and digests").
+			Optional(),
```
golanglemonade: this should have a validator, e.g. from another place we have timezone:

```
Validate(func(s string) error {
				if s == "" {
					return nil
				}

				_, err := time.LoadLocation(s)

				return err
			}).
```

internal/graphapi/workflowadmin.resolvers.go
```diff
@@ -0,0 +1,306 @@
...
+func (r *mutationResolver) CancelWorkflowInstance(ctx context.Context, id string, reason *string) (*model.WorkflowInstanceAdminPayload, error) {
+	if !workflowsEnabled(r.db) {
+		return nil, ErrWorkflowsDisabled
+	}
+	if id == "" {
+		return nil, rout.NewMissingRequiredFieldError("id")
+	}
+
+	client := withTransactionalMutation(ctx)
+	allowCtx := workflows.AllowContext(ctx)
```
golanglemonade: Why is this allowing - theres no permissions checks before - would't that allow anyone to cancel any workflow?

internal/graphapi/workflowadmin.resolvers.go
```diff
@@ -0,0 +1,306 @@
...
+		return nil, rout.NewMissingRequiredFieldError("targets")
+	}
+
+	client := withTransactionalMutation(ctx)
+	assignment, err := client.WorkflowAssignment.Get(ctx, input.ID)
+	if err != nil {
+		return nil, parseRequestError(ctx, err, common.Action{Action: common.ActionUpdate, Object: "workflowassignment"})
+	}
+
+	allowCtx := workflows.AllowContext(ctx)
```
golanglemonade: Same here - we shouldn't be passing an allowContext here?

internal/graphapi/vulnerability_extended.resolvers.go
```diff
@@ -0,0 +1,96 @@
...
+	client := withTransactionalMutation(ctx)
+	vuln, err := client.Vulnerability.Get(ctx, id)
+	if err != nil {
+		return nil, parseRequestError(ctx, err, common.Action{Action: common.ActionUpdate, Object: "vulnerability"})
+	}
+
+	status := strings.TrimSpace(input.Status)
+	update := vuln.Update()
+	if status != "" {
+		update.SetStatus(status).SetOpen(false)
```
golanglemonade, line 36: setting a non-empty status implies it is no longer open?

internal/graphapi/vulnerability_extended.resolvers.go
```diff
@@ -0,0 +1,96 @@
...
+
+		if details := strings.TrimSpace(lo.FromPtrOr(input.Details, "")); details != "" {
+			reviewCreate.SetDetails(details)
+		} else if reason := strings.TrimSpace(lo.FromPtrOr(input.Reason, "")); reason != "" {
+			reviewCreate.SetSummary(reason)
+		}
+
+		if userID, err := auth.GetSubjectIDFromContext(ctx); err == nil && userID != "" {
+			reviewCreate.SetReviewerID(userID)
+		}
```
golanglemonade: You need to check the Subject type; this won't be a user id when its an API token but the `reviewer_id` field looks to be a edge to the `User` table

internal/graphapi/workflowadmin_helpers.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		if generated.IsNotFound(err) {
+			return rout.ErrPermissionDenied
+		}
+		return parseRequestError(ctx, err, common.Action{Action: common.ActionGet, Object: "org_membership"})
+	}
+
+	if membership.Role != enums.RoleOwner && membership.Role != enums.RoleAdmin {
+		return rout.ErrPermissionDenied
+	}
```
golanglemonade: You shouldn't need to query the db at all here. You can use the auth context:

```
if err != nil {
    return nil, err
}

// make sure the `ownerID` matches the organizationID
if au.OrganizationID != ownerID {
     return rout.ErrPermissionDenied
}

if au.OrganizationRole != auth.OwnerRole && au.OrganizationRole != auth.AdminRole {
     return rout.ErrPermissionDenied
}
```

internal/graphapi/workflowadmin_helpers.go
```diff
@@ -0,0 +1,327 @@
...
+			}
+
+			bypassCtx := workflows.AllowBypassContext(ctx)
+			bypassCtx = workflows.WithSkipEventEmission(bypassCtx)
+			workflows.MarkSkipEventEmission(bypassCtx)
+			if instance.OwnerID != "" {
+				if err := auth.SetOrganizationIDInAuthContext(bypassCtx, instance.OwnerID); err != nil {
+					return nil, err
+				}
+			}
```
golanglemonade: this is already done on line 201 and I don't see anything that would possibly set the `instance.OwnerID` after the first time

internal/graphapi/workflowadmin.resolvers.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, parseRequestError(ctx, err, common.Action{Action: common.ActionGet, Object: "workflowassignment"})
+	}
+
+	if err := r.requireWorkflowAdmin(ctx, assignment.OwnerID); err != nil {
+		return nil, err
+	}
+
+	if assignment.OwnerID != "" {
+		if err := auth.SetOrganizationIDInAuthContext(allowCtx, assignment.OwnerID); err != nil {
```
golanglemonade: Can you use `auth.SetOrganizationInAuthContext` in all places you have it instead of `auth.SetOrganizationIDInAuthContext`. The first checks that the user is actually authenticated in the org, the second:

```
// SetOrganizationIDInAuthContext sets the organization ID in the auth context
// this should only be used when creating a new organization and subsequent updates
// need to happen in the context of the new organization
```

internal/graphapi/workflowassignment_helpers.go
```diff
@@ -55,6 +62,39 @@ func (r *mutationResolver) validateAssignmentDecision(ctx context.Context, id st
...
+	membership, err := withTransactionalMutation(ctx).OrgMembership.Query().
+		Where(
+			orgmembership.UserIDEQ(userID),
+			orgmembership.OrganizationIDEQ(orgID),
+			orgmembership.RoleEQ(enums.RoleOwner),
+		).
+		Exist(ctx)
+	if err != nil || !membership {
+		return rout.ErrPermissionDenied
+	}
```
golanglemonade: No need for the query, you can just check the org role from the context. I would just use `auth. HasFullOrgWriteAccessFromContext(ctx)` so it will include the future super admins.

internal/graphapi/workflowassignmentextended.resolvers.go
```diff
@@ -122,8 +122,114 @@ func (r *mutationResolver) RejectWorkflowAssignment(ctx context.Context, id stri
...
+	metadata["change_requested_at"] = decidedAt.Format(time.RFC3339)
+	metadata["change_requested_by"] = decisionCtx.UserID
+	if reason != nil && *reason != "" {
+		metadata["change_reason"] = *reason
+	}
+	if len(inputs) > 0 {
+		metadata["change_inputs"] = inputs
+	}
+
+	allowCtx := workflows.AllowContext(ctx)
```
golanglemonade, line 152: Just confirming - this one doesn't need any further auth checks because you only need read access (which is done by `validateAssignmentDecision`) to make this request?

internal/graphapi/workflowproposal_helpers.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if domainObjectType != "" && !strings.EqualFold(domainObjectType, objectType.String()) {
+		return fmt.Errorf("%w: proposal domain does not match object type", rout.ErrBadRequest)
+	}
+
+	if len(changes) == 0 {
+		return nil
+	}
+
+	allowed := make(map[string]struct{}, len(fields))
```
golanglemonade: with your love of `lo` I'm surprised you didn't use `lo.SliceToMap` :-D

internal/graphapi/workflowassignmentextended.resolvers.go
matoszz: correct, `validateAssignmentDecision` -> `assertAssignmentActor` so the allow context is required after that validation occurs to perform the update against `WorkflowAssignment`

internal/graphapi/workflowproposal_helpers.go
matoszz: just a missed opportunity is all /cattyping

PR #1981, opened 2026-02-06
Ability to revoke NDA trustcenter requests

adelowo:


internal/graphapi/trustcenterndarequest_test.go
```diff
@@ -1056,3 +1059,100 @@ func TestMutationRequestNewTrustCenterToken(t *testing.T) {
...
+func TestMutationRevokeNDARequestsRemovesDocAccess(t *testing.T) {
+	cleanupTrustCenterData(t)
+
+	trustCenter := (&TrustCenterBuilder{client: suite.client}).MustNew(testUser1.UserCtx, t)
+
+	_ = (&TemplateBuilder{
+		client:        suite.client,
+		Kind:          enums.TemplateKindTrustCenterNda,
+		TrustCenterID: trustCenter.ID,
+	}).MustNew(testUser1.UserCtx, t)
```
golanglemonade: does this need to get cleaned up to avoid conflicts?

internal/graphapi/schema/trustcenterndarequest.graphql
```diff
@@ -144,4 +153,14 @@ type BulkUpdateStatusPayload {
     Updated nda request IDs
     """
     totalUpdated: Int!
+}
+
+"""
+Return response for revokeNDARequests mutation
```
golanglemonade: looks like your comments have `revokeNDARequests` but the mutation is `deleteBulkTrustCenterNDARequest` - I dont really care which as long as the comments match the name of the mutation

adelowo: ha seen, I initially had it as revoke but switched to the bulk delete format. will update

PR #1984, opened 2026-02-07
chore: break up linter + tests

golanglemonade:
- splits out handler tests into own job
- splits linter into three jobs, excludes all generated paths
- removes `if_changes` from main run for chart automation to avoid missing updates
- bump trivy


.buildkite/pipeline.yaml
```diff
@@ -629,8 +729,8 @@ steps:
               severity: CRITICAL,HIGH
               ignore-unfixed: true
               scanners: misconfig,secret,vuln
-              skip-files: "cosign.key,Dockerfile.dev"
-              trivy-version: "0.67.0"
+              skip-files: "cosign.key,Dockerfile.dev,internal/graphapi/generated/*,internal/ent/generated/*,internal/ent/historygenerated/*"
+              trivy-version: "0.68.2"
```
golanglemonade, line 733: I tried bumping this to 0.69.+ but it had panics so just going to 0.68.x

PR #1992, opened 2026-02-10
Add doc automation step to pipeline.yaml

MuriloLopes477:
- Add core's trigger step for tag based docs automation

depends on: https://github.com/theopenlane/openlane-docs/pull/83


.buildkite/pipeline.yaml
```diff
@@ -197,6 +197,11 @@ steps:
                 - "HELM_CHART_PATH"
                 - "SLACK_WEBHOOK_URL"
                 - "BUILDKITE_BUILD_CHECKOUT_PATH=/workdir"
+      - label: ":docs: update docs for release"
```
golanglemonade: This is currently in the `precheck`group; can you move it to the `publish` group? Should be able to just put this at the very end of this file

MuriloLopes477: Yes, will do.

.buildkite/pipeline.yaml
```diff
@@ -806,3 +806,8 @@ steps:
         depends_on: "apollo-publish-schema"
         if: build.tag != null
         soft_fail: true
+      - label: ":docs: update docs for release"
+          trigger: openlane-docs
+          depends_on: "image_tag_automation"
+          if: build.tag != null
+          soft_fail: true
```
golanglemonade: I think the indentation is off here:


```suggestion
      - label: ":docs: update docs for release"
        trigger: openlane-docs
        depends_on: "image_tag_automation"
        if: build.tag != null
        soft_fail: true
```

golanglemonade (approved):

PR #1997, opened 2026-02-10
Workflows <-> Integrations Interface | Provider Implementations | Service Activation Flow

matoszz:
- Added base provider abstraction with shared capabilities, operations, and client descriptor handling  
- Added activation service coordinating OAuth and non-OAuth credential flows with optional health checks              
- Added CEL-based vulnerability ingest mapper with field transformation, provider-specific overrides, and schema validation                      
- Refactored Github Webhook handler / added async ingest event processing via event bus for webhook payloads                                                         
- Added default CEL mappings for GitHub Dependabot, code scanning, and secret scanning alerts           
- Collapsed AWS into single provider with product sub-providers; added AWS config builder with STS assume-role and session credential support for each subprovider and generic client builder
- Added authenticated HTTP client with credential injection for non-SDK providers
- Added operation helpers for config decoding, result building, and failure formatting
- Added normalized string types (`TrimmedString`, `LowerString`, `UpperString`) for config normalization             
- Added integration executor for running provider operations as workflow actions
- Added notification template support with Slack/Teams dispatch via integration operations  
- Added CEL-based template renderer with recursive value rendering and depth limits
- Consolidated duplicated approval/review execution logic into single `executeGatedAction` function              
- Changed AddServerOptions to accept variadic arguments for cleaner server initialization

### DIFF
--- a/common/integrations/auth/oauth.go
+++ b/common/integrations/auth/oauth.go
@@ -63,7 +63,7 @@ func HTTPGetJSON(ctx context.Context, client *http.Client, url string, bearer st
 
 	options := []httpsling.Option{
 		httpsling.Get(url),
-		httpsling.Header(httpsling.HeaderAccept, "application/json"),
+		httpsling.Header(httpsling.HeaderAccept, httpsling.ContentTypeJSON),
 	}
 
 	if bearer != "" {
@@ -101,7 +101,7 @@ func HTTPPostJSON(ctx context.Context, client *http.Client, url string, bearer s
 
 	options := []httpsling.Option{
 		httpsling.Post(url),
-		httpsling.Header(httpsling.HeaderAccept, "application/json"),
+		httpsling.Header(httpsling.HeaderAccept, httpsling.ContentTypeJSON),
 	}
 
 	if body != nil {
--- a/common/integrations/operations/operation_helpers.go
+++ b/common/integrations/operations/operation_helpers.go
@@ -65,6 +65,8 @@ func HealthCheckRunner[T any](tokenType TokenType, endpoint string, failureMsg s
 			client, token, err = auth.ClientAndOAuthToken(input)
 		case TokenTypeAPI:
 			client, token, err = auth.ClientAndAPIToken(input)
+		default:
+			return types.OperationResult{}, ErrUnsupportedTokenType
 		}
 
 		if err != nil {
--- a/internal/graphapi/workflow_validators.go
+++ b/internal/graphapi/workflow_validators.go
@@ -8,6 +8,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/samber/lo"
+
 	"github.com/theopenlane/core/common/enums"
 	"github.com/theopenlane/core/common/models"
 	"github.com/theopenlane/core/internal/ent/generated"
@@ -472,41 +474,48 @@ func validateWorkflowDefinitionTemplateRefs(ctx context.Context, client *generat
 		}
 	}
 
-	for id := range ids {
-		exists, err := client.NotificationTemplate.Query().
+	if len(ids) > 0 {
+		idList := lo.Keys(ids)
+
+		foundIDs, err := client.NotificationTemplate.Query().
 			Where(
-				notificationtemplate.IDEQ(id),
+				notificationtemplate.IDIn(idList...),
 				notificationtemplate.ActiveEQ(true),
 				notificationtemplate.Or(
 					notificationtemplate.OwnerIDEQ(ownerID),
 					notificationtemplate.SystemOwnedEQ(true),
 				),
 			).
-			Exist(ctx)
+			IDs(ctx)
 		if err != nil {
 			return err
 		}
-		if !exists {
-			return fmt.Errorf("%w: %s", ErrNotificationTemplateNotFound, id)
+
+		if missing := lo.Without(idList, foundIDs...); len(missing) > 0 {
+			return fmt.Errorf("%w: %s", ErrNotificationTemplateNotFound, missing[0])
 		}
 	}
 
-	for key := range keys {
-		exists, err := client.NotificationTemplate.Query().
+	if len(keys) > 0 {
+		keyList := lo.Keys(keys)
+
+		foundKeys, err := client.NotificationTemplate.Query().
 			Where(
-				notificationtemplate.KeyEQ(key),
+				notificationtemplate.KeyIn(keyList...),
 				notificationtemplate.ActiveEQ(true),
 				notificationtemplate.Or(
 					notificationtemplate.OwnerIDEQ(ownerID),
 					notificationtemplate.SystemOwnedEQ(true),
 				),
 			).
-			Exist(ctx)
+			Select(notificationtemplate.FieldKey).
+			Strings(ctx)
 		if err != nil {
 			return err
 		}
-		if !exists {
-			return fmt.Errorf("%w: %s", ErrNotificationTemplateKeyNotFound, key)
+
+		if missing := lo.Without(keyList, foundKeys...); len(missing) > 0 {
+			return fmt.Errorf("%w: %s", ErrNotificationTemplateKeyNotFound, missing[0])
 		}
 	}
 
--- a/internal/integrations/ingest/vulnerabilities.go
+++ b/internal/integrations/ingest/vulnerabilities.go
@@ -306,7 +306,7 @@ func upsertVulnerability(ctx context.Context, db *generated.Client, orgID string
 			}
 		}
 
-		if _, err := update.Save(ctx); err != nil {
+		if err := update.Exec(ctx); err != nil {
 			return false, err
 		}
 
@@ -318,7 +318,7 @@ func upsertVulnerability(ctx context.Context, db *generated.Client, orgID string
 		create.AddIntegrationIDs(integrationID)
 	}
 
-	if _, err := create.Save(ctx); err != nil {
+	if err := create.Exec(ctx); err != nil {
 		return false, err
 	}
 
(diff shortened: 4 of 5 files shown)


common/integrations/auth/oauth.go
```diff
@@ -0,0 +1,161 @@
...
+		var err error
+		requester, err = httpsling.New(httpsling.WithHTTPClient(client))
+		if err != nil {
+			return err
+		}
+	}
+
+	options := []httpsling.Option{
+		httpsling.Get(url),
+		httpsling.Header(httpsling.HeaderAccept, "application/json"),
```
golanglemonade: nit: and reason you didn't use the constant on the value too? 
```suggestion
		httpsling.Header(httpsling.HeaderAccept, httpsling.ContentTypeJSONUTF8),
```

common/integrations/auth/oauth.go
```diff
@@ -0,0 +1,161 @@
...
+		var err error
+		requester, err = httpsling.New(httpsling.WithHTTPClient(client))
+		if err != nil {
+			return err
+		}
+	}
+
+	options := []httpsling.Option{
+		httpsling.Post(url),
+		httpsling.Header(httpsling.HeaderAccept, "application/json"),
```
golanglemonade: Here too
```suggestion
		httpsling.Header(httpsling.HeaderAccept, httpsling.ContentTypeJSONUTF8),
```

common/integrations/operations/operation_helpers.go
```diff
@@ -0,0 +1,83 @@
...
+	return func(ctx context.Context, input types.OperationInput) (types.OperationResult, error) {
+		var client *auth.AuthenticatedClient
+		var token string
+		var err error
+
+		switch tokenType {
+		case TokenTypeOAuth:
+			client, token, err = auth.ClientAndOAuthToken(input)
+		case TokenTypeAPI:
+			client, token, err = auth.ClientAndAPIToken(input)
```
golanglemonade, line 67: Does this need a fallback default error if its not either of those?

```
default:
	return types.OperationResult{}, fmt.Errorf("%w: %v", ErrIUnsupportedTokenType, tokenType)
```

common/integrations/operations/operation_helpers.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		if err != nil {
+			return types.OperationResult{}, err
```
golanglemonade, line 73: It seems strange to me that is is returning an empty OperationResult instead of either `nil` (return a pointer (`*types.OperationResult, error) or an OperationResult with the status as error - but hard to tell from just looking at this part of the code

internal/ent/schema/integration.go
```diff
@@ -149,7 +149,6 @@ func (i Integration) Edges() []ent.Edge {
 func (Integration) Indexes() []ent.Index {
 	return []ent.Index{
 		index.Fields(ownerFieldName, "kind").
-			Unique().
 			Annotations(entsql.IndexWhere("deleted_at is NULL")),
```
golanglemonade, line 152: just to confirm; this allows an organization to have multiple of the same kind of organization?

internal/ent/schema/integrationwebhook.go
```diff
@@ -105,7 +105,7 @@ func (IntegrationWebhook) Fields() []ent.Field {
 			Optional().
 			Nillable().
 			Annotations(
-				entgql.OrderField("LAST_DELIVERY_AT"),
+				entgql.OrderField("last_delivery_at"),
```
golanglemonade, line 108: lol you know I saw all these in caps and almost commented originally but the docs actually have them as CAPS; and I was unsure why we had only enums as ALL CAPS and everything else as lower

internal/graphapi/workflow_validators.go
```diff
@@ -400,6 +442,115 @@ func validateWorkflowDefinitionConflicts(ctx context.Context, client *generated.
...
+					notificationtemplate.OwnerIDEQ(ownerID),
+					notificationtemplate.SystemOwnedEQ(true),
+				),
+			).
+			Exist(ctx)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			return fmt.Errorf("%w: %s", ErrNotificationTemplateNotFound, id)
```
golanglemonade: are keys and ids defined separately? What if I have a valid key but an invalid id? We want it to error here and not go to the other loop - which seems right if - either all all valid or fail on a single one missing.

Instead of loops, This could be a single query with `IDIn` and make sure the count match?

internal/integrations/ingest/vulnerabilities.go
```diff
@@ -0,0 +1,362 @@
...
+		ProviderState:     c.providerStateMap,
+	}.Map()
+
+	allowed, err := c.mapper.EvaluateFilter(ctx, spec.FilterExpr, vars)
+	if err != nil {
+		return nil, false, err
+	}
+	if !allowed {
+		return nil, false, nil
+	}
```
golanglemonade, line 167: nit: you can simplify this: 
```suggestion
	if err != nil  || !allowed {
		return nil, false, err
	}
```

internal/integrations/ingest/vulnerabilities.go
```diff
@@ -0,0 +1,362 @@
...
+		} else {
+			result.Summary.Updated++
+		}
+	}
+
+	return result, nil
+}
+
+// decodeAlertPayload converts a raw alert payload into a map
+func decodeAlertPayload(payload json.RawMessage) (map[string]any, error) {
```
golanglemonade, line 248: Didn't your last PR have this exact same function but exported? Maybe another package though?

internal/integrations/ingest/vulnerabilities.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != nil {
+		return input, err
+	}
+
+	if err := json.Unmarshal(bytes, &input); err != nil {
+		return input, err
+	}
+
+	return input, nil
+}
```
golanglemonade, line 270: we have at least one, if not more than that instances of a generic func for this; I feel like we just need a library for this (not needed in this PR but I keep seeing the exactly same code in many places around type -> Json and back

```
// convertToMap converts a generic input struct to a map[string]any
func convertToMap[T any](ctx context.Context, input T) (map[string]any, error) {
	byteInput, err := json.Marshal(input)
	if err != nil {
		return nil, err
	}

	var mapInput map[string]any
	if err := json.Unmarshal(byteInput, &mapInput); err != nil {
		return nil, err
	}

	return mapInput, nil
}

```

internal/integrations/ingest/vulnerabilities.go
```diff
@@ -0,0 +1,362 @@
...
+
+			if !exists {
+				update.AddIntegrationIDs(integrationID)
+			}
+		}
+
+		if _, err := update.Save(ctx); err != nil {
+			return false, err
+		}
+
```
golanglemonade:
```suggestion
		if err := update.Exec(ctx); err != nil {
			return false, err
		}

```

internal/integrations/ingest/vulnerabilities.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	create := db.Vulnerability.Create().SetInput(input)
+	if integrationID != "" {
+		create.AddIntegrationIDs(integrationID)
+	}
+
+	if _, err := create.Save(ctx); err != nil {
+		return false, err
+	}
```
golanglemonade:
```suggestion
	if err := create.Exec(ctx); err != nil {
		return false, err
	}
```

internal/integrations/providers/github/vulnerabilities.go
```diff
@@ -0,0 +1,436 @@
...
+type githubVulnerabilityConfig struct {
+	// RepositorySelector controls which repositories to scan
+	operations.RepositorySelector
+	// Pagination controls page sizing for list calls
+	operations.Pagination
+	// PayloadOptions controls payload inclusion
+	operations.PayloadOptions
+
+	// AlertTypes selects which alert types to collect
+	AlertTypes []types.LowerString `json:"alert_types"`
```
golanglemonade, line 39: saw this a bunch, so just wanted to ask why the configs themselves are all unexported (e.g. `githubVulnerabilityConfig) but the variables themselves are exported (e.g. `AlertTypes`)?

internal/ent/schema/integrationwebhook.go
matoszz: well afaik only enums are supposed to be capitalized which is why i changed it

internal/integrations/ingest/vulnerabilities.go
matoszz: this isn't the same logic, though; in 1 return i'm returning as an err and the other the err is nil, they aren't the same

internal/ent/schema/integration.go
matoszz: Correct

internal/integrations/ingest/vulnerabilities.go
matoszz: There's a `toMap` in mapper.go that's similar but takes any (requiring marshal+unmarshal), whereas `decodeAlertPayload` takes json.RawMessage (already serialized JSON, only needs unmarshal); they serve different purposes

internal/integrations/providers/github/vulnerabilities.go
matoszz: why do the fields need to be un-exported if the function isn't? The only reason to un-export a field is if the function itself is exported, and this isn't?

internal/integrations/ingest/vulnerabilities.go
golanglemonade: yeah it is, because the `err` is `nil` in the second one, so they equivalate to the same thing

PR #2002, opened 2026-02-11
Allow questionnaire responses to be saved as draft

adelowo:
- response can be saved as draft on frontend side
- next time the same user comes to fill the questionnaire, If they have existing data, it is sent to the frontend to fill the UI

### DIFF
--- a/internal/httpserve/handlers/questionnaire_test.go
+++ b/internal/httpserve/handlers/questionnaire_test.go
@@ -417,9 +417,10 @@ func (suite *HandlerTestSuite) TestGetQuestionnaireReturnsDraftData() {
 	require.NoError(t, err)
 
 	testEmail := "draftfetch@example.com"
+	anonUserID := fmt.Sprintf("anon_questionnaire_%s", assessment.ID)
 
 	anonUser := &auth.AnonymousQuestionnaireUser{
-		SubjectID:      fmt.Sprintf("anon_questionnaire_%s", ulids.New().String()),
+		SubjectID:      anonUserID,
 		SubjectEmail:   testEmail,
 		OrganizationID: testUser1.OrganizationID,
 		AssessmentID:   assessment.ID,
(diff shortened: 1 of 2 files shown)


internal/httpserve/handlers/questionnaire.go
```diff
@@ -195,35 +200,62 @@ func (h *Handler) SubmitQuestionnaire(ctx echo.Context, openapi *OpenAPIContext)
...
+	var documentDataID string
 
-	if assessment.TemplateID != "" {
-		documentDataQuery = documentDataQuery.SetTemplateID(assessment.TemplateID)
-	}
+	if assessmentResponse.DocumentDataID != "" {
+		_, err = h.DBClient.DocumentData.UpdateOneID(assessmentResponse.DocumentDataID).
+			SetData(req.Data).
+			Save(allowCtx)
+		if err != nil {
```
golanglemonade: nit:

```suggestion
		err = h.DBClient.DocumentData.UpdateOneID(assessmentResponse.DocumentDataID).
			SetData(req.Data).
			Exec(allowCtx)
		if err != nil {
```

internal/httpserve/handlers/questionnaire_test.go
```diff
@@ -380,6 +380,162 @@ func (suite *HandlerTestSuite) TestGetQuestionnaireAlreadyCompleted() {
...
+		OrganizationID: testUser1.OrganizationID,
+		AssessmentID:   assessment.ID,
+	}
+
+	questionnaireCtx := contextx.With(ctx, auth.QuestionnaireContextKey{})
+	questionnaireCtx = auth.WithAnonymousQuestionnaireUser(questionnaireCtx, anonUser)
+	questionnaireCtx = auth.WithAuthenticatedUser(questionnaireCtx, &auth.AuthenticatedUser{
+		SubjectID:      testUser1.ID,
+		OrganizationID: testUser1.OrganizationID,
+	})
```
golanglemonade, line 434: Why is the subject id testtUser1, when it also has `anonUser` in the ctx?

internal/httpserve/handlers/questionnaire_test.go
```diff
@@ -882,3 +1038,218 @@ func (suite *HandlerTestSuite) TestSubmitQuestionnaireAuthenticatedUser() {
...
+
+	assessmentResponse, err := suite.db.AssessmentResponse.Create().
+		SetAssessmentID(assessment.ID).
+		SetEmail(testEmail).
+		SetOwnerID(testUser1.OrganizationID).
+		SetStatus(enums.AssessmentResponseStatusSent).
+		Save(questionnaireCtx)
+	require.NoError(t, err)
+
+	anonUserID := fmt.Sprintf("anon_questionnaire_%s", ulids.New().String())
```
golanglemonade: We need to make sure this isn't a random ulid in the prod code path - the user id needs to be set off the assessmentID

PR #2013, opened 2026-02-16
Fix bug with hasTrustCenterXXXX scoping

adelowo:
tested with 
CleanShot 2026-02-16 at 18 29 09@2x


internal/ent/schemautil/trustcenter_scope.go
```diff
@@ -0,0 +1,51 @@
...
+			if anon.TrustCenterID != "" {
+				s.Where(sql.EQ(s.C("trust_center_id"), anon.TrustCenterID))
+				return
+			}
+		}
+
+		orgIDs, err := auth.GetOrganizationIDsFromContext(ctx)
+		if err != nil || len(orgIDs) == 0 {
+			return
+		}
```
golanglemonade, line 33: We would silently not filter if there is not org id set - I can't think of any time this would happen, but seems like we should return an error instead?

PR #2022, opened 2026-02-17
fix: Race condition: concurrent access to shared config without synchronization

### LINKED ISSUE #2021: [Bug] : Race condition: concurrent access to shared config without synchronization
The reload() function modifies shared state (s.Config.Handler.JWTKeys and s.Config.Handler.TokenManager) without any mutex protection. This function is called both synchronously at line 43 and asynchronously from a goroutine at line 78. Multiple file system events could trigger concurrent reload() calls, leading to race conditions when reading/writing s.Config.Handler fields. This is a classic TOCTOU race where config could be read by other goroutines while being updated.

file  : internal/httpserve/serveropts/keywatch.go 
line : 33 to 35  

Fixed race condition

manab-pr:
Closes #2021 

Added a sync.Mutex to protect the reload() function .


manab-pr:
@golanglemonade  plz review it , this is another one .

PR #2038, opened 2026-02-18
Ability to update framework controls on revision changes

adelowo:


internal/graphapi/control_clone.go
```diff
@@ -545,6 +567,169 @@ func (r *mutationResolver) bulkCreateSubcontrolNoTransaction(ctx context.Context
...
+		}
+
+		var standardRevision *string
+		if cu.sourceControl.Edges.Standard != nil {
+			standardRevision = &cu.sourceControl.Edges.Standard.Revision
+		}
+
+		var newSubcontrols []*generated.CreateSubcontrolInput
+
+		for _, sc := range cu.sourceControl.Edges.Subcontrols {
```
golanglemonade, line 676: can this be handled by `cloneSubcontrols` instead of duplicating the create logic here?

PR #2043, opened 2026-02-20
Trust Center Controls Display

matoszz:
- Added `trust_center_visibility` enum field to Control (`PUBLICLY_VISIBLE`, `PROTECTED`, `NOT_VISIBLE`, defaults to `NOT_VISIBLE`) controlling the publishing state for trust center display                                                                                                   
- Added `is_trust_center_control` boolean field to Control, set by the system during control clone, excluded from create/update mutation inputs 
- Updated FGA authorization model to allow wildcard users (user:*, service:*) on the `can_view` relation for controls, enabling anonymous public read access
- Added FGA test tuples and assertions for anonymous trust center visitors and trust center org owners                                      
- Added `HookControlTrustCenterVisibility` hook that manages FGA wildcard viewer tuples when visibility changes on create or update           
- Extracted `ControlVisibilityTupleAction` as a pure function for testability with unit tests covering all visibility transition combinations
- Added `InterceptorTrustCenterControl` interceptor that filters control queries for anonymous trust center users to only return publicly visible trust center controls
- Enabled anonymous trust center access on the Control schema
- Skip group-based permission filtering for anonymous trust center users since they don't belong to groups
- Added `CatalogTrustCenterModule` to the Control feature entitlements alongside `CatalogComplianceModule`
- Added `TestQueryControlTrustCenterVisibility` integration tests covering anonymous and authenticated user access to public, hidden, and regular controls
- Fixed privacy rule test that was using `CatalogTrustCenterModule` as a non-qualifying module for Control, updated to use `CatalogEntityManagementModule`
- DB migrations, GraphQL schema, resolvers, test client, CSV, and integration mapping code regenerated


internal/ent/hooks/control.go
```diff
@@ -81,3 +84,118 @@ func HookControlReferenceFramework() ent.Hook {
...
+		})
+	}, ent.OpCreate|ent.OpUpdateOne)
+}
+
+// ControlVisibilityTupleAction determines whether wildcard viewer tuples should be
+// written or deleted based on the trust center visibility state of a control
+func ControlVisibilityTupleAction(isTrustCenterControl bool, newVisibility, oldVisibility enums.TrustCenterDocumentVisibility, visibilityChanged bool) (shouldWrite, shouldDelete bool) {
+	if !isTrustCenterControl {
+		return false, false
+	}
```
golanglemonade: This is already checked in the hook itself, shouldn't need another check here

internal/ent/hooks/control.go
```diff
@@ -81,3 +84,118 @@ func HookControlReferenceFramework() ent.Hook {
...
+				return v, nil
+			}
+
+			if m.Op().Is(ent.OpCreate) {
+				return handleControlVisibilityCreate(ctx, m, ctrl)
+			}
+
+			return handleControlVisibilityUpdate(ctx, m, ctrl)
+		})
+	}, ent.OpCreate|ent.OpUpdateOne)
```
golanglemonade: This needs to work for `ent.OpUpdate` in order to work for bulk edited controls

internal/ent/schema/control.go
```diff
@@ -64,6 +65,18 @@ func (Control) Fields() []ent.Field {
 		field.String("standard_id").
 			Comment("the id of the standard that the control belongs to, if applicable").
 			Optional(),
+		field.Enum("trust_center_visibility").
+			GoType(enums.TrustCenterDocumentVisibility("")).
```
golanglemonade: Should this be limited to public + private - protected doesn't really make sense here?

internal/ent/interceptors/controls.go
```diff
@@ -4,11 +4,54 @@ import (
...
+		// skip mutations
+		opCtx := graphql.GetOperationContext(ctx)
+		if opCtx.Operation.Operation == ast.Mutation {
+			return nil
+		}
+
+		if auth.IsSystemAdminFromContext(ctx) {
+			return nil
+		}
+
```
golanglemonade: You shouldn't need any of this, none of this to applies to a trust center anon user
```suggestion
```

internal/ent/schema/control.go
```diff
@@ -218,6 +240,7 @@ func (c Control) Policy() ent.Policy {
 func (Control) Modules() []models.OrgModule {
 	return []models.OrgModule{
 		models.CatalogComplianceModule,
+		models.CatalogTrustCenterModule,
```
golanglemonade, line 244: This will give access to all controls to users just paying for the trust center module; we may want to add an interceptor that will only return trust center controls if the user only has the compliance module

internal/ent/schema/control.go
```diff
@@ -64,6 +65,18 @@ func (Control) Fields() []ent.Field {
...
+			GoType(enums.TrustCenterDocumentVisibility("")).
+			Default(enums.TrustCenterDocumentVisibilityNotVisible.String()).
+			Optional().
+			Comment("visibility of the control on the trust center, controls the publishing state for trust center display"),
+		field.Bool("is_trust_center_control").
+			Default(false).
+			Optional().
+			Annotations(
+				entgql.Skip(entgql.SkipMutationCreateInput, entgql.SkipMutationUpdateInput),
+			).
```
golanglemonade, line 79: How is this field being set if we don't allow this to be set by system admin and there is no hook to do it based on the standard?

internal/graphapi/control_test.go
```diff
@@ -2877,3 +2879,118 @@ func TestMutationUpdateBulkControl(t *testing.T) {
...
+		SetTrustCenterVisibility(enums.TrustCenterDocumentVisibilityNotVisible).
+		SetOwnerID(testUser1.OrganizationID).
+		Save(dbCtx)
+	assert.NilError(t, err)
+
+	// create a regular control (not a trust center control)
+	regularControl := (&ControlBuilder{client: suite.client}).MustNew(testUser1.UserCtx, t)
+
+	// add wildcard viewer tuple for the publicly visible control (simulates what the hook does)
+	wildcardTuples := fgax.CreateWildcardViewerTuple(publicControl.ID, "control")
```
golanglemonade: Can you use the api client instead so its tested instead of doing it manually?

internal/graphapi/control_test.go
```diff
@@ -2877,3 +2879,118 @@ func TestMutationUpdateBulkControl(t *testing.T) {
...
+	anonCtx := createAnonymousTrustCenterContext(trustCenter.ID, testUser1.OrganizationID)
+	anonCtx = contextx.With(anonCtx, auth.TrustCenterContextKey{})
+
+	testCases := []struct {
+		name            string
+		client          *testclient.TestClient
+		ctx             context.Context
+		queryID         string
+		errorMsg        string
+		expectedResults int // for list queries, -1 means use queryID for get
```
golanglemonade: all the test cases have `-1`, this isn't needed
```suggestion
```

internal/graphapi/control_test.go
```diff
@@ -2877,3 +2879,118 @@ func TestMutationUpdateBulkControl(t *testing.T) {
 	(&Cleanup[*generated.ControlImplementationDeleteOne]{client: suite.client.db.ControlImplementation, ID: controlImplementation.ID}).MustDelete(testUser1.UserCtx, t)
 	(&Cleanup[*generated.GroupDeleteOne]{client: suite.client.db.Group, IDs: []string{ownerGroup.ID, delegateGroup.ID, groupMember.GroupID}}).MustDelete(testUser1.UserCtx, t)
 }
```
golanglemonade, line 2880: Can you add a test to `TestQueryTrustCenterAsAnonymousUser` that uses the `GetTrustCenterFrontendQuery` and ensure you can get all the fields the UI will use to pull this data. I ran into a lot of issues and use this test to make sure child objects can be pulled via the top level.

internal/ent/schema/control.go
matoszz: in the clone control function

internal/ent/hooks/control.go
```diff
@@ -81,3 +85,222 @@ func HookControlReferenceFramework() ent.Hook {
...
+	ctrl, ok := v.(*generated.Control)
+	if !ok {
+		return v, nil
+	}
+
+	// only manage wildcard tuples for trust center controls
+	if !ctrl.IsTrustCenterControl {
+		return v, nil
+	}
+
```
golanglemonade: On `updateOne` you can improve this by moving everything after the mutation, checking if its a trust center control first and then using `m.OldTrustCenterVisibilty()` - this way you aren't running a query on every control update to get the old value when its not needed most of the time

internal/ent/hooks/control.go
```diff
@@ -81,3 +85,222 @@ func HookControlReferenceFramework() ent.Hook {
...
+	updatedIDs := getMutationIDs(ctx, m)
+	if len(updatedIDs) == 0 {
+		return next.Mutate(ctx, m)
+	}
+
+	oldControls, err := m.Client().Control.Query().
+		Where(control.IDIn(updatedIDs...)).
+		Select(control.FieldID, control.FieldTrustCenterVisibility).
```
golanglemonade: Is `IsTrustCenterControl` immutable? Should it be? Then you don't have to do two queries and can grab that in the initial query and not run two queries in here?

internal/graphapi/trustcenter_test.go
```diff
@@ -831,7 +833,55 @@ func TestQueryTrustCenterAsAnonymousUser(t *testing.T) {
...
+
+		tc := resp.TrustCenters.Edges[0].Node
+
+		assert.Check(t, tc.ID != "")
+		assert.Check(t, tc.GetSetting() != nil)
+		assert.Check(t, tc.TrustCenterCompliances.Edges != nil)
+		assert.Check(t, tc.TrustCenterDocs.Edges != nil)
+		assert.Check(t, tc.TrustCenterEntities.Edges != nil)
+		assert.Check(t, tc.TrustCenterSubprocessors.Edges != nil)
+		assert.Check(t, tc.Posts.Edges != nil)
```
golanglemonade, line 869: This doesnt actually check for the controls edge

PR #2045, opened 2026-02-20
Deprecate Soiree | Add in-memory option to Gala | Port All Listeners

matoszz:
- Replaced soiree with gala as the event dispatch system across the entire codebase -- built gala's in-memory dispatch mode, a worker pool backed by alitto/pond with Prometheus metrics, and per-envelope queue/retry controls, then migrated every listener category (entitlements, trust center cache, workflows, notifications, integration ingest) to gala and removed all legacy Eventer code
- Unified event hook emission into a single multi-runtime path that accepts variadic providers, deduplicates runtimes, and generates concern-based topic names (direct, workflow, notification), replacing both the legacy and intermediate single-runtime hooks
- Introduced a structured mutation dispatch model via the internal/mutations package with MutationConcern constants and a ChangeSet contract shared across MutationGalaPayload, TriggerInput, and WorkflowInstanceContext, replacing all raw ent.Mutation references
- Converted the workflow engine and reconciler from soiree to gala, simplifying event emission by removing generic type parameters, replacing async error channel draining with synchronous handling, and collapsing observability emission to a single code path
- Migrated integration operations to gala with typed operation policies -- defined IntegrationOperationEnvelope with per-type timeout, retry, and idempotency configuration and replaced the old Emitter dependency with gala-based listener registration
- Consolidated notification handlers by deduplicating approval/review completion logic into gatedActionCompletedDetails, removing the documentMutation interface in favor of explicit field parameters, and adding an isUpdateOperation helper
- Eliminated soiree from all server wiring and test infrastructure -- replaced dual emitter fields with a single Handler.Gala, removed soiree.Pool references across serve/resolvers/tests, dropped soiree.ShutdownAll() from shutdown, and rewired all test suites to gala.NewInMemory()
- Fixed correctness issues including mistagged JSON struct fields on WorkflowAssignmentRejection, normalized GitHub alert type constants into a shared location, replaced lo.Assign with stdlib maps.Clone, and removed unused error constants and soiree references from schema comments
- Improved workflow schema generation and helpers by refactoring the generator to use a decorator map with reflect-based value handling, updating action type descriptions, and extracting shared helpers to reduce duplication
- Moved subscription duration tracking from per-subscription to per-connection scope and replaced context flag key structs with gala-native context flags

### DIFF
--- a/internal/integrations/providers/github/vulnerabilities.go
+++ b/internal/integrations/providers/github/vulnerabilities.go
@@ -14,6 +14,12 @@ import (
 	"github.com/theopenlane/core/common/integrations/types"
 )
 
+const (
+	githubAlertTypeDependabot     = "dependabot"
+	githubAlertTypeCodeScanning   = "code_scanning"
+	githubAlertTypeSecretScanning = "secret_scanning"
+)
+
 type githubInstallationRepositoriesResponse struct {
 	// TotalCount is the number of repositories returned by GitHub
 	TotalCount int `json:"total_count"`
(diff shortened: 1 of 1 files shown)


internal/ent/notifications/events_test.go
```diff
@@ -16,586 +21,161 @@ func TestNeedsTaskDBQuery(t *testing.T) {
...
-			},
-			expected: true,
-		},
-		{
-			name: "missing owner ID",
-			fields: &taskFields{
-				title:    "Test Task",
-				entityID: "task-123",
-			},
-			expected: true,
```
golanglemonade, line 46: Any reason these test cases were removed?

internal/ent/notifications/mentions.go
```diff
@@ -99,41 +99,35 @@ func handleNoteMutation(ctx *soiree.EventContext, payload *events.MutationPayloa
 		oldText = fields.oldTextJSON
 	}
 
-	// If no valid text, nothing to process
+	// If no valid text, nothing to process.
```
golanglemonade: nit: why were all these comments updated to add a `.`?

internal/entdb/client.go
```diff
@@ -55,26 +55,19 @@ type client struct {
...
-// If workflowConfig is provided and enabled, it also creates and sets the WorkflowEngine on the client.
-func WithEventer(eventer *hooks.Eventer, workflowConfig *workflows.Config) Option {
+// WithWorkflows wires workflow-related hooks and optionally configures the workflow engine.
+func WithWorkflows(workflowConfig *workflows.Config) Option {
 	return func(c *ent.Client) {
-		eventer.Initialize(c)
-
 		if workflowConfig != nil && workflowConfig.Enabled {
-			wfEngine, err := engine.NewWorkflowEngineWithConfig(c, eventer.Emitter, workflowConfig)
+			wfEngine, err := engine.NewWorkflowEngineWithConfig(c, nil, workflowConfig)
```
golanglemonade: hard for me to tell without puling the brach, but I noticed `eventer.Emitter` was removed in a lot of places, is it still need as input to `NewWorkflowEngineWithConfig`?

matoszz: Not strictly; I refactored it a bit. I had updated to `nil` because gala didn't exist when `entdb.New()` ran, but I split it out so that we create gala before initializing the DB client, and then pass it in which should simplify the rationalization around ordering

PR #2050, opened 2026-02-22
Additional Integrations Specification Data | Shared Helpers and Deduplication

matoszz:
Primarily updated provider JSON configs with expanded schema definitions and UI metadata. The additional data / fields allow for the UI to derive "connect" or "configure" buttons for integration cards, as well as better functional grouping for "coming soon" functionality.

- Introduced generics for config decoding and metadata persistence, eliminating per-type boilerplate
- Consolidated duplicated string normalization, default constants, and health check runners into shared helpers
- Replaced verbose map/filter/fallback patterns with lo utilities and simplified JSON marshaling paths             
- Inlined trivial single-use wrappers and removed dead code
- Enriched operation failure reporting with automatic error detail injection

### DIFF
--- a/common/integrations/operations/operation_helpers.go
+++ b/common/integrations/operations/operation_helpers.go
@@ -48,6 +48,9 @@ func HealthOperation(name types.OperationName, description string, client types.
 	}
 }
 
+// DefaultSampleSize is the standard number of sample items returned in operation results
+const DefaultSampleSize = 5
+
 // TokenType indicates whether to use OAuth or API token extraction
 type TokenType int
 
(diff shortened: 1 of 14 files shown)


common/integrations/auth/authenticated_client.go
```diff
@@ -16,10 +15,9 @@ type AuthenticatedClient struct {
 
 // NewAuthenticatedClient builds an AuthenticatedClient with a cloned header map
 func NewAuthenticatedClient(bearerToken string, headers map[string]string) *AuthenticatedClient {
-	cloned := cloneHeaders(headers)
 	return &AuthenticatedClient{
-		BearerToken: strings.TrimSpace(bearerToken),
```
golanglemonade, line 21: just so I understand, why do you know longer want the spaces trimmed here?

golanglemonade (approved):

PR #2061, opened 2026-02-24
Directory + Integration + Identity Holder Schema Relationships + Refs

matoszz:
- Added identity holder links to support deduplicated cross-directory identity views across systems like Google Workspace, Slack, and GitHub and connected directory objects to integration + platforms so imported records have clear provenance
- Added directory source labeling and avatar metadata so UI can render per-source presence and profile imagery without requiring direct integration access, and extended platform relationships so the UI can clearly model system boundaries
- Linked findings to both directory accounts and identity holders so risk can be analyzed at both the raw account level and unified person level (relevant for user access review workflows)
- Enabled integration-mapping metadata on directory accounts so provider payloads can be upserted through the existing mapping pipeline


internal/ent/hooks/directoryaccount.go
```diff
@@ -0,0 +1,106 @@
...
+			}
+			canonicalEmail := *account.CanonicalEmail
+
+			// Run linking logic with privacy bypass so backend syncs can always reconcile identities
+			allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
+
+			holder, err := getOrCreateIdentityHolder(allowCtx, m.Client(), ownerID, canonicalEmail, account.DisplayName, account.JobTitle, account.Department)
+			if err != nil {
+				return nil, err
+			}
```
golanglemonade, line 51: I'd recommend logging errors for easier debugging
```suggestion
			if err != nil {
			    logx.FromContext(ctx).Error().Err(err).Msg("unable to get or create identity holder")
				return nil, err
			}
```

Same with the one update

internal/ent/hooks/directoryaccount.go
```diff
@@ -0,0 +1,106 @@
...
+			// Keep a direct link from directory account -> identity holder
+			update := m.Client().DirectoryAccount.UpdateOneID(account.ID)
+			if err := update.Mutation().SetField(directoryAccountIdentityHolderField, holder.ID); err != nil {
+				return nil, err
+			}
+
+			if err := update.Exec(allowCtx); err != nil {
+				return nil, err
+			}
+
```
golanglemonade, line 66: do you need this? I've ever used `SetField` but looks like its already doing the saving to the db?

internal/ent/schema/integration.go
```diff
@@ -71,6 +71,11 @@ func (Integration) Fields() []ent.Field {
 			Annotations(
 				entgql.OrderField("integration_type"),
 			),
+		field.String("platform_id").
+			Comment("optional platform associated with this integration for downstream inventory linkage").
+			Optional().
+			NotEmpty().
```
golanglemonade, line 77: The comment says optional, but there is a `NotEmpty` Validator?

internal/ent/schema/directorysyncrun.go
```diff
@@ -48,6 +48,11 @@ func (DirectorySyncRun) Fields() []ent.Field {
 			Comment("integration this sync run executed for").
 			NotEmpty().
 			Immutable(),
+		field.String("platform_id").
+			Comment("optional platform associated with this sync run").
+			Optional().
+			NotEmpty().
```
golanglemonade, line 54: Comment says optional but there is a `NotEmpty` Validator

internal/ent/schema/directorymembership.go
```diff
@@ -48,15 +48,22 @@ func (DirectoryMembership) Fields() []ent.Field {
 			Comment("integration that owns this directory membership").
 			NotEmpty().
 			Immutable(),
+		field.String("platform_id").
+			Comment("optional platform associated with this directory membership").
+			Optional().
+			NotEmpty().
```
golanglemonade, line 54: Comment says optional but there is a `NotEmpty` validator

internal/ent/schema/directorygroup.go
```diff
@@ -50,6 +50,11 @@ func (DirectoryGroup) Fields() []ent.Field {
 			Comment("integration that owns this directory group").
 			NotEmpty().
 			Immutable(),
+		field.String("platform_id").
+			Comment("optional platform associated with this directory group").
```
golanglemonade, line 54: Comment says optional but there is a NotEmpty validator

internal/ent/schema/directoryaccount.go
```diff
@@ -47,13 +49,34 @@ func (DirectoryAccount) PluralName() string {
 func (DirectoryAccount) Fields() []ent.Field {
 	return []ent.Field{
 		field.String("integration_id").
-			Comment("integration that owns this directory account").
+			Comment("optional integration that owns this directory account when sourced by an integration").
+			Optional().
 			NotEmpty().
```
golanglemonade, line 54: Should this have NotEmpty if is opional? Same with directory_sync_run_id and platform_id

internal/ent/schema/integration.go
matoszz: It can be optional and not empty?

golanglemonade: you're correct, I was thinking about cases where we use this to ensure a string field is set, but doesn't need to be provided by the user. this works fine.

internal/ent/hooks/directoryaccount.go
matoszz: Yes. `SetField` just delegates to individual setters; "SetField sets the value of a field with the given name. It returns an error if the field is not defined in the schema, or if the type mismatched the field type." Does not commit anything / alter the db, only the mutation in memory

PR #2063, opened 2026-02-25
Specs for exporting

Achiket123:
Hey @golanglemonade please review these and please tell me if there is any thing else i need to change.
Thankyou.


.entc/entc_github.com_theopenlane_core_internal_ent_schema_1771751453.go
```diff
@@ -0,0 +1,586 @@
+// Code generated by 'entc generate'. DO NOT EDIT.
```
golanglemonade: @Achiket123 this shouldn't be committed; it is part of the generate process but should be automatically removed when entc is complete

internal/ent/schema/export.go
```diff
@@ -112,13 +111,13 @@ func (e Export) Annotations() []schema.Annotation {
...
-			rule.AllowQueryIfSystemAdmin(),
+			policy.CheckOrgReadAccess(),
 		),
 		policy.WithMutationRules(
 			policy.AllowCreate(),
 		),
 		policy.WithOnMutationRules(ent.OpUpdate|ent.OpUpdateOne|ent.OpDelete|ent.OpDeleteOne,
-			rule.AllowMutationIfSystemAdmin(),
+			policy.CheckOrgWriteAccess(),
 		),
```
golanglemonade, line 134: The privacy rules shouldn't need to be changed for this; `csv` already works properly as is. 

The job is setup with a system-admin token and uses user `impersonation` to pull the data the user has access to.

PR #2074, opened 2026-02-26
Bulk evidence export

adelowo:
- added a new `export mode` to export schema
- added a new export metadata struct that can include other details. right now it only contains `keepOriginalFileName` in the exports

> will extend the job for the evidence export

### DIFF
--- a/internal/ent/hooks/export.go
+++ b/internal/ent/hooks/export.go
@@ -100,7 +100,7 @@ func handleExportCreate(ctx context.Context, m *generated.ExportMutation, next e
 		}
 	}
 
-	_, err = m.Job.Insert(ctx, args, nil)
+	err = enqueueJob(ctx, m.Job, args, nil)
 
 	return v, err
 }


common/enums/exportmode.go
```diff
@@ -0,0 +1,61 @@
...
+		return &ExportModeFlat
+	case ExportModeFolder.String():
+		return &ExportModeFolder
+	default:
+		return &ExportModeInvalid
+	}
+}
+
+// MarshalGQL implements the gqlgen Marshaler interface.
+func (r ExportMode) MarshalGQL(w io.Writer) {
```
golanglemonade: Does the template need to be updated? These now should be using the common functions like:

```

// MarshalGQL implements the gqlgen Marshaler interface.
func (r AssessmentType) MarshalGQL(w io.Writer) { marshalGQL(r, w) }

// UnmarshalGQL implements the gqlgen Unmarshaler interface.
func (r *AssessmentType) UnmarshalGQL(v any) error { return unmarshalGQL(r, v) }

...
```

internal/ent/hooks/export.go
```diff
@@ -84,11 +85,22 @@ func handleExportCreate(ctx context.Context, m *generated.ExportMutation, next e
...
+	if exportType == enums.ExportTypeEvidence {
+		mode, _ := m.Mode()
+		args.Mode = mode
+
+		if metadata, ok := m.ExportMetadata(); ok {
+			args.ExportMetadata = &metadata
+		}
+	}
+
+	_, err = m.Job.Insert(ctx, args, nil)
```
golanglemonade: can you update this to use `enqueueJob` like other places were updated? this must have been missed

PR #2075, opened 2026-02-27
Shuffle Integration Record Creation | Confirm Credentials Pre-store

matoszz:
- Moved integration record creation from the OAuth start flow to the callback so incomplete or abandoned OAuth flows no longer produce orphaned integration records
- Removed the IntegrationID requirement from BeginAuthorization since the integration record doesn't exist yet at that point in the flow, and inverted the configure flow to validate connectivity with the provider before persisting credentials, so we never store credentials we haven't confirmed work
- Added a MintPayload step before the health check so providers that require credential exchange (e.g. GCP workload identity) can produce a usable token from the raw input before validation runs
- Added an IntegrationProviders config override mechanism so OAuth client IDs and secrets are follow the same koanf structure as everything else and are merged with the integration provider config
- Fixed a bug where Active and Visible could not be overridden to false via the config merge mechanism because the zero-value pruning treated false the same as unset


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/750

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

PR #2077, opened 2026-02-27
OSCAL Support Mapping + entx plugin

matoszz:
This PR adds OSCAL-specific fields and 1 new schema (SystemDetail) which fill in the holes between our graph and the existing OSCAL specification model (actually quite close, unsurprisingly). I didn't re-use the existing Platform schema for the SystemDetail because it's a trust boundary construct that would conflict with what the OSCAL specification requires. Only schema changes + generate output were included in here.


internal/ent/schema/systemdetail.go
```diff
@@ -0,0 +1,187 @@
...
+// Policy of the SystemDetail
+func (s SystemDetail) Policy() ent.Policy {
+	return policy.NewPolicy(
+		policy.WithMutationRules(
+			policy.CanCreateObjectsUnderParents([]string{
+				Program{}.PluralName(),
+				Platform{}.PluralName(),
+			}),
+			policy.CheckCreateAccess(),
+			policy.CheckOrgWriteAccess(),
```
golanglemonade: I'd probably recommend doing this order, the `CheckOrgWriteAccess` doesn't need an fga check so its a cheaper operation than the `CanCreateObjectsUnderParents`
```suggestion
			policy.CheckOrgWriteAccess(),
			policy.CheckCreateAccess(),
			policy.CanCreateObjectsUnderParents([]string{
				Program{}.PluralName(),
				Platform{}.PluralName(),
			}),
```

internal/ent/schema/systemdetail.go
```diff
@@ -0,0 +1,187 @@
...
+
+// Mixin of the SystemDetail
+func (s SystemDetail) Mixin() []ent.Mixin {
+	return mixinConfig{
+		prefix: "SDT",
+		additionalMixins: []ent.Mixin{
+			newObjectOwnedMixin[generated.SystemDetail](s,
+				withParents(Program{}, Platform{}),
+				withOrganizationOwner(true),
+				withListObjectsFilter(),
```
golanglemonade: Why use `withListObjectsFilter`? This function is actually completely unused today (and nolinted because of that) I had left to see if we needed to switch anything to it but list objects is an expensive operation and only by default used on organizations because users will have very few and the complexity is cheap

internal/ent/schema/systemdetail.go
```diff
@@ -0,0 +1,187 @@
...
+func (SystemDetail) Modules() []models.OrgModule {
+	return []models.OrgModule{
+		models.CatalogComplianceModule,
+	}
+}
+
+// Annotations of the SystemDetail
+func (SystemDetail) Annotations() []schema.Annotation {
+	return []schema.Annotation{
+		entfga.SelfAccessChecks(),
```
golanglemonade, line 163: You need an update to the fga model to add this new type

PR #2287, opened 2026-04-10
Implement review date + frequencies for evidences

adelowo:


internal/ent/hooks/evidence.go
```diff
@@ -15,6 +16,63 @@ import (
...
+
+				if !frequencyMutated {
+					oldFrequency, err := m.OldReviewFrequency(ctx)
+					if err != nil {
+						return nil, err
+					}
+
+					frequency = oldFrequency
+				}
+			default:
```
golanglemonade, line 123: You are missing the `OpUpdate` case. If there is a bulk update of evidence to, for example, set the frequency, all the records need to be updated based on their previous creation date

internal/ent/schema/evidence.go
```diff
@@ -139,6 +139,22 @@ func (Evidence) Fields() []ent.Field {
...
 			Optional(),
+		field.Enum("review_frequency").
+			Comment("the cadence for reviewing the evidence").
+			GoType(enums.Frequency("")).
+			Default(enums.FrequencyYearly.String()).
+			Optional().
+			Annotations(
+				entgql.OrderField("REVIEW_FREQUENCY"),
+			),
+		field.Time("next_review_at").
```
golanglemonade: We already have `renewalDate`, this should be used instead of a new field

adelowo: ha seen. I supposed they would mean different things. will fix shortly

golanglemonade (dismissed):

golanglemonade (dismissed):

matoszz (approved):

PR #2297, opened 2026-04-13
DirectoryAccount -> IdentityHolder refactored logic, move to async listener, add tests

matoszz:
Refactors the previously simple directoryaccount hook to be executed via async listener(s) on create + update, and perform linking in the general order of:
- matching on identity holder email address
- matching on other directoryaccounts that have the email and a linked identity holder
- matching on full name (in the future we could add fuzzier matching logic and confidence % but kept it a direct match)

Adds hooks to perform clean up on delete of one of the related records (if identity holder is removed, unlinks directory accounts since the soft delete edges would still be there, if directory account is removed updates the email aliases).


internal/ent/hooks/directoryaccount.go
```diff
@@ -4,114 +4,49 @@ import (
...
-	if !generated.IsConstraintError(err) {
-		logx.FromContext(ctx).Error().Err(err).Msg("failed to create identity holder")
-
-		return nil, err
-	}
-
-	return client.IdentityHolder.Query().
-		Where(identityholder.OwnerID(ownerID), identityholder.Email(canonicalEmail)).
-		Only(ctx)
+	}, ent.OpDeleteOne)
```
matoszz, line 51: This schema doesn't have soft delete

golanglemonade, line 51: Does this need to support bulk delete and allow OpDelete?
```suggestion
	}, ent.OpDeleteOne|ent.OpDelete)
```

matoszz: Just adding OpDelete would require the hook to also be refactored to scan all of the deleted ID's to be able to handle a bulk operation - right now it's only pulling + querying for single ID's

golanglemonade (approved):

PR #2313, opened 2026-04-17
fixes: googleworkspace, githubapp integrations

golanglemonade:
- fixes issue `customer id missing from installation metadata` with googleworkspace by adding additional customer read only scope
- fixes issue `api request failed: Expected type 'number', but it was malformed: "-"."` by breaking the gql struct from full orgMemberNode struct
- Maps additional fields
- Sets `isOpenlaneUser` based on org membership
- Removes updatedAt default on update for avatarURL so its not updated on any update to the row

image


internal/ent/hooks/listeners_identityresolution.go
```diff
@@ -264,49 +267,100 @@ func createIdentityHolder(ctx context.Context, client *entgen.Client, account *e
...
+
 // applyPrimarySourceDefaults sets authoritative fields on a new identity holder create builder from a primary source directory account
 func applyPrimarySourceDefaults(create *entgen.IdentityHolderCreate, account *entgen.DirectoryAccount) {
 	create.SetStatus(mapDirectoryAccountStatus(account.Status))
 	create.SetIsActive(account.Status == enums.DirectoryAccountStatusActive)
 	create.SetNillableTitle(account.JobTitle)
 	create.SetNillableDepartment(account.Department)
+	create.SetEmailAliases(account.EmailAliases)
+	create.SetNillablePhoneNumber(account.PhoneNumber)
+	create.SetExternalUserID(account.ExternalID)
```
matoszz, line 297: You should not set these. The email aliases is set by the hook because its an array and needs to be appended to, this would overwrite. The external user ID should also not be set only if it's the primary directory.

internal/ent/hooks/listeners_identityresolution.go
```diff
@@ -264,49 +267,100 @@ func createIdentityHolder(ctx context.Context, client *entgen.Client, account *e
...
+
+	if account.AddedAt != nil {
```
matoszz, line 299: should only be set if it's the primary directory

internal/ent/hooks/listeners_identityresolution.go
```diff
@@ -264,49 +267,100 @@ func createIdentityHolder(ctx context.Context, client *entgen.Client, account *e
...
+		create.SetStartDate(models.DateTime(*account.AddedAt))
```
matoszz, line 300: should only be set if it's the primary directory

internal/ent/hooks/listeners_identityresolution.go
```diff
@@ -264,49 +267,100 @@ func createIdentityHolder(ctx context.Context, client *entgen.Client, account *e
...
 	accounts, err := client.DirectoryAccount.Query().
 		Where(directoryaccount.IdentityHolderID(holder.ID),
 			directoryaccount.CanonicalEmailNotNil()).
-		Select(directoryaccount.FieldCanonicalEmail).All(ctx)
+		Select(directoryaccount.FieldCanonicalEmail, directoryaccount.FieldEmailAliases).All(ctx)
 	if err != nil {
 		return err
 	}
 
-	aliases := lo.Uniq(lo.FilterMap(accounts, func(a *entgen.DirectoryAccount, _ int) (string, bool) {
```
matoszz, line 303: why is this being removed?

internal/ent/hooks/listeners_identityresolution.go
```diff
@@ -264,49 +267,100 @@ func createIdentityHolder(ctx context.Context, client *entgen.Client, account *e
...
 
 // syncEmailAliases rebuilds the identity holder's email_aliases from all linked
-// directory accounts' canonical emails, excluding the primary email
+// directory accounts' canonical emails and per-account aliases, excluding the primary email
 func syncEmailAliases(ctx context.Context, client *entgen.Client, holder *entgen.IdentityHolder) error {
 	accounts, err := client.DirectoryAccount.Query().
 		Where(directoryaccount.IdentityHolderID(holder.ID),
 			directoryaccount.CanonicalEmailNotNil()).
-		Select(directoryaccount.FieldCanonicalEmail).All(ctx)
+		Select(directoryaccount.FieldCanonicalEmail, directoryaccount.FieldEmailAliases).All(ctx)
```
matoszz, line 345: this goes against other matching logic in the hook...

internal/ent/schema/directoryaccount.go
```diff
@@ -113,6 +115,34 @@ func (DirectoryAccount) Fields() []ent.Field {
 				entx.IntegrationMappingField().UpsertKey(),
 				entgql.OrderField("canonical_email"),
 			),
+		field.Strings("email_aliases").
```
matoszz, line 118: directory accounts typically do not have email aliases

internal/ent/schema/identity_holder.go
```diff
@@ -86,6 +89,14 @@ func (IdentityHolder) Fields() []ent.Field {
 		field.Strings("email_aliases").
 			Comment("alternate email address for the identity holder in an array").
 			Optional().
+			Validate(func(emails []string) error {
```
matoszz, line 92: I don't think we should do this personally

internal/ent/schema/identity_holder.go
```diff
@@ -200,6 +211,12 @@ func (IdentityHolder) Fields() []ent.Field {
 		field.JSON("metadata", map[string]any{}).
 			Comment("additional metadata about the identity holder").
 			Optional(),
+		field.String("avatar_remote_url").
```
matoszz, line 214: Should only be set if we know it from the primary directory

internal/ent/hooks/listeners_identityresolution.go
```diff
@@ -237,17 +241,16 @@ func resolveIdentityHolder(ctx context.Context, client *entgen.Client, account *
...
 	canonicalEmail := *account.CanonicalEmail
+	exists := resolveIsOpenlaneUser(ctx, client, canonicalEmail)
 
 	create := client.IdentityHolder.Create().
 		SetOwnerID(account.OwnerID).
 		SetEmail(canonicalEmail).
+		SetIsOpenlaneUser(exists).
 		SetFullName(buildFullName(account.DisplayName, account.GivenName, account.FamilyName, canonicalEmail))
 
-	if account.PrimarySource {
```
matoszz, line 246: I don't think this logic should be changed; we only want to set the status field if we know the origin is the primary directory

internal/ent/schema/directoryaccount.go
golanglemonade: Google workspace does?

internal/ent/schema/identity_holder.go
golanglemonade: We shouldn't validate that emails are emails? Why not?

internal/ent/schema/identity_holder.go
golanglemonade: Right, the place that it is set is `enrichFromPrimarySource` - so it would only apply for the primary directory

internal/ent/hooks/listeners_identityresolution.go
golanglemonade: It's not, its updated to allow the combination with email aliases

internal/ent/hooks/listeners_identityresolution.go
golanglemonade: I assume this is about `create.SetEmailAliases(account.EmailAliases)`? Good catch, I realized that later and thought I removed it but must have had it in two places. 

The other two - phone number and external id - this is in the `applyPrimarySourceDefaults` function so that is already covered?

internal/ent/hooks/listeners_identityresolution.go
golanglemonade: this is in `applyPrimarySourceDefaults`

internal/ent/hooks/listeners_identityresolution.go
golanglemonade: this is in `applyPrimarySourceDefaults`

internal/ent/hooks/listeners_identityresolution.go
golanglemonade: I'm not sure I understand what you mean here?

internal/ent/hooks/listeners_identityresolution.go
golanglemonade: This was already reverted, not sure what github was showing you because this comment came in after I reverted 🤔

matoszz (approved):

PR #2321, opened 2026-04-22
feat: gsp scc integration, aws security hub, aws iam, cf groups 

golanglemonade:
Related UI changes: https://github.com/theopenlane/openlane-ui/pull/1762

- Cleans up mappings and sets up the GCP SCC Integration for risks, findings, and vulns
- Cleans up mappings and sets up the AWS security hub and aws iam integration for findings, vulns, and directories, configures assume role + tested service account credential setups
- Ensure private key data is not logged in body dumps
- Sets security_level off severity if score is not set in hook
- fixes directory name on existing directory-sync integrations using the persist functions
image

- Hides `SCIM` from available integrations for now
- Allows for only partial syncs per integration if it supports more than one sync. (e.g. only sync users from github, etc)

image

- Adds error logging when generic errors are returned to user
- Adds CF groups/roles
- Fixes CF health check to use account token instead of a user token
- Moves CF `AccountID` to credentials instead of config to align with other providers

image

- Adds new CheckResult table, this isn't used yet but will be in a follow-up PR for AWS Config ingest
- Auto creates vendor based on integration family if it doesnt already exist, otherwise updates the integration id to link to integration; I updated the family on the integration to be the matching name for creating the vendor, this wasn't used that I could tell besides a text representation/grouping between integrations.
image

- Adds new CheckResult schema; this will be used on a follow-up with the AWS Config integration
- Adds `directory_name` to the other directory tables
- Changes edge access to integrations to check Organization access, integrations inherit permissions from the org
- Adds `Permissions` to the the builder to show in the UI what scopes are required, particularly for those where user has to create a token with the correct access
- Removes hardcoded sources/directory names and sets these based on the integration name during persistence
- Adds tests to ensure valid `cel` in mappings
- Started adding `example/*.json` for some integrations and a few additional tests related to those

Tested Security Hub with  Sample Data:
image

Follow-ups
- Finish AWS Config (Checks + Check Results), Assets
- Link Finding Controls
- Other integrations, e.g. SCIM, Okta, Azure


common/enums/controlcheckstatus.go
```diff
@@ -0,0 +1,38 @@
+package enums
+
+import "io"
+
+// ControlCheckStatus is a custom type representing the various states of ControlCheckStatus.
+type ControlCheckStatus string
+
+var (
+	// ControlCheckStatusCompliant indicates the compliant.
+	ControlCheckStatusCompliant ControlCheckStatus = "COMPLIANT"
```
matoszz: We should definitely not use this language to describe a check status, it's either passing or failing not "compliant" or not

common/enums/exporttype.go
```diff
@@ -16,6 +16,8 @@ var (
 	ExportTypeContact ExportType = "CONTACT"
 	// ExportTypeControl indicates the control.
 	ExportTypeControl ExportType = "CONTROL"
+	// ExportTypeControlHealth indicates the controlhealth.
+	ExportTypeControlHealth ExportType = "CONTROL_HEALTH"
```
matoszz: also don't know that "control health" is a good way to describe this

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	{Key: integrationgenerated.IntegrationMappingFindingSeverity, Expr: `'severity' in payload && payload.severity != "SEVERITY_UNSPECIFIED" ? payload.severity : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingDescription, Expr: `'description' in payload ? payload.description : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingDisplayName, Expr: `'category' in payload && payload.category != "" ? ` + titleCaseCEL("payload.category") + ` : ('name' in payload ? payload.name : "")`},
+	{Key: integrationgenerated.IntegrationMappingFindingExternalURI, Expr: `'external_uri' in payload ? payload.external_uri : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingReportedAt, Expr: `'create_time' in payload ? payload.create_time : null`},
+	{Key: integrationgenerated.IntegrationMappingFindingSourceUpdatedAt, Expr: `'event_time' in payload ? payload.event_time : ('create_time' in payload ? payload.create_time : null)`},
+	{Key: integrationgenerated.IntegrationMappingFindingRawPayload, Expr: "payload"},
+	{Key: integrationgenerated.IntegrationMappingFindingSource, Expr: `"gcp-security-command-center"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	{Key: integrationgenerated.IntegrationMappingFindingResourceName, Expr: `'name' in payload ? payload.name : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingState, Expr: `'state' in payload ? payload.state : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingRecommendedActions, Expr: `'source_properties' in payload && 'recommendation' in payload.source_properties ? payload.source_properties.recommendation : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingStepsToReproduce, Expr: `'source_properties' in payload && 'explanation' in payload.source_properties ? [payload.source_properties.explanation] : []`},
+	{Key: integrationgenerated.IntegrationMappingFindingTargetDetails, Expr: `'kubernetes' in payload && size(payload.kubernetes) > 0 ? payload.kubernetes : 'containers' in payload && size(payload.containers) > 0 ? payload.containers : null`},
```
matoszz: what is this??

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+        ? payload.contextUris.mitreUri.url : null,
+      ('relatedFindingUri' in payload.contextUris && 'url' in payload.contextUris.relatedFindingUri)
+        ? payload.contextUris.relatedFindingUri.url : null,
+      ('virustotalIndicatorQueryUri' in payload.contextUris && size(payload.contextUris.virustotalIndicatorQueryUri) > 0)
+        ? payload.contextUris.virustotalIndicatorQueryUri[0].url : null
+    ].filter(u, u != null)
+  : []
+`},
+	{Key: integrationgenerated.IntegrationMappingFindingExploitability, Expr: `'attack_exposure_score' in payload ? payload.attack_exposure_score : 0.0`},
+	{Key: integrationgenerated.IntegrationMappingFindingScore, Expr: `'vulnerability' in payload && payload.vulnerability != null && 'cve' in payload.vulnerability && payload.vulnerability.cve != null && 'cvssv3' in payload.vulnerability.cve && payload.vulnerability.cve.cvssv3 != null ? payload.vulnerability.cve.cvssv3.base_score : 0.0`},
```
matoszz: this isn't right; we should be pulling from cvss 4 and not falling back to different values

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	// try best effort name matching
+	{Key: integrationgenerated.IntegrationMappingFindingProduction, Expr: `'name' in payload && payload.name.contains("prod") ? true: false`},
```
matoszz: we should not be doing this, this is configured via a cel expr by the user not by us

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+})
+
+// mapExprVuln is the CEL mapping expression for GCP Security Command Center vuln payloads
+var mapExprVuln = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingVulnerabilitySource, Expr: `"GCP Security Command Center"`},
+	// use CVE ID -> parse CVE from description -> fall back to category -> name
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityDisplayName, Expr: `'vulnerability' in payload && payload.vulnerability != null && 'cve' in payload.vulnerability && payload.vulnerability.cve != null && 'id' in payload.vulnerability.cve && payload.vulnerability.cve.id != "" ? payload.vulnerability.cve.id : ('description' in payload && payload.description.contains("CVE-") ? "CVE-" + payload.description.split("CVE-")[1].split(")")[0].split(" ")[0].split(".")[0].split(",")[0] : ('category' in payload && payload.category != "" ? ` + titleCaseCEL("payload.category") + ` : ('name' in payload ? payload.name : "")))`},
```
matoszz: same comments as above, take the field if it's in the payload or not all this fallback logic isn't appropriate

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	// use CVE ID -> parse CVE from description
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityCveID, Expr: `'vulnerability' in payload && payload.vulnerability != null && 'cve' in payload.vulnerability && payload.vulnerability.cve != null && 'id' in payload.vulnerability.cve && payload.vulnerability.cve.id != "" ? payload.vulnerability.cve.id : ('description' in payload && payload.description.contains("CVE-") ? "CVE-" + payload.description.split("CVE-")[1].split(")")[0].split(" ")[0].split(".")[0].split(",")[0] : "")`},
```
matoszz: same comment as above

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityExternalID, Expr: `'finding_id' in payload ? payload.finding_id : 'name' in payload ? payload.name : ""`},
 	{Key: integrationgenerated.IntegrationMappingVulnerabilityExternalOwnerID, Expr: "resource"},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityCategory, Expr: `'category' in payload ? payload.category : ""`},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityVulnerabilityStatusName, Expr: `'state' in payload ? payload.state : ""`},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityCategory, Expr: `'category' in payload ? ` + titleCaseCEL("payload.category") + ` : ""`},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityVulnerabilityStatusName, Expr: `'state' in payload && payload.state == "ACTIVE" ? "Open" : "Closed"`},
```
matoszz, line 73: We shouldn't be hard coding matching logic like this in the expressions CEL is for field extraction not a bunch of conditional logic like this

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityOpen, Expr: `'state' in payload && payload.state == "ACTIVE" ? true : false`},
 	{Key: integrationgenerated.IntegrationMappingVulnerabilitySeverity, Expr: `'severity' in payload && payload.severity != "SEVERITY_UNSPECIFIED" ? payload.severity : ""`},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilitySummary, Expr: `'category' in payload && payload.category != "" ? payload.category : ('canonical_name' in payload && payload.canonical_name != "" ? payload.canonical_name : ('name' in payload ? payload.name : ""))`},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilitySummary, Expr: `'description' in payload && payload.description != "" ? payload.description : ""`},
```
matoszz, line 76: duplicate of an expression above

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityCveID, Expr: `'vulnerability' in payload && payload.vulnerability != null && 'cve' in payload.vulnerability && payload.vulnerability.cve != null && 'id' in payload.vulnerability.cve ? payload.vulnerability.cve.id : ""`},
 	{Key: integrationgenerated.IntegrationMappingVulnerabilityExternalURI, Expr: `'external_uri' in payload ? payload.external_uri : ""`},
 	{Key: integrationgenerated.IntegrationMappingVulnerabilityDiscoveredAt, Expr: `'create_time' in payload ? payload.create_time : null`},
 	{Key: integrationgenerated.IntegrationMappingVulnerabilitySourceUpdatedAt, Expr: `'event_time' in payload ? payload.event_time : ('create_time' in payload ? payload.create_time : null)`},
 	{Key: integrationgenerated.IntegrationMappingVulnerabilityRawPayload, Expr: "payload"},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityFixAvailable, Expr: `'vulnerability' in payload && payload.vulnerability != null && 'cve' in payload.vulnerability && payload.vulnerability.cve != null && 'upstream_fix_available' in payload.vulnerability.cve ? payload.vulnerability.cve.upstream_fix_available : false`},
```
matoszz: this conditional logic shouldn't be necessary

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityFirstPatchedVersion, Expr: `'vulnerability' in payload && payload.vulnerability != null && 'cve' in payload.vulnerability && payload.vulnerability.cve != null && 'fixed_package' in payload.vulnerability.cve  && 'package_version' in payload.vulnerability.cve.fixed_package ? payload.vulnerability.cve.fixedPackage.package_version : ""`},
+})
+
+// mapExprRisk is the CEL mapping expression for GCP Security Command Center risk payloads
+var mapExprRisk = providerkit.CelMapExpr([]providerkit.CelMapEntry{
```
matoszz, line 92: is it a risk, or is it a finding? otherwise you're mapping this twice

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+	{Key: integrationgenerated.IntegrationMappingRiskExternalID, Expr: `'finding_id' in payload ? payload.finding_id : 'name' in payload ? payload.name : ""`},
+	{Key: integrationgenerated.IntegrationMappingRiskName, Expr: `'category' in payload ? payload.category : ""`},
+	{Key: integrationgenerated.IntegrationMappingRiskStatus, Expr: `'state' in payload && payload.state == "ACTIVE" ? "OPEN" : "CLOSED"`},
+	{Key: integrationgenerated.IntegrationMappingRiskImpact, Expr: `'severity' in payload && payload.severity != "SEVERITY_UNSPECIFIED" ? payload.severity : ""`},
```
matoszz: shouldn't be hard coding enum evaluation like this

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
+                ? (
+                    payload.toxic_combination.attack_exposure_score == 0 ? "UNLIKELY" :
+                    payload.toxic_combination.attack_exposure_score <= 1 ? "LIKELY" :
+                    "HIGHLY_LIKELY"
+                  )
+                : "UNLIKELY"
+            )
+      )
+`},
+	{Key: integrationgenerated.IntegrationMappingRiskScore, Expr: `
```
matoszz: don't hard code enum evaluation in CEL expressions

internal/integrations/definitions/gcpscc/mappings.go
```diff
@@ -1,35 +1,155 @@
...
 			Spec: types.MappingOverride{
-				FilterExpr: "true",
+				FilterExpr: `payload.finding_class == "VULNERABILITY" ? true : false`,
+				MapExpr:    mapExprVuln,
+			},
+		},
+		{
+			Schema: integrationgenerated.IntegrationMappingSchemaFinding,
+			Spec: types.MappingOverride{
+				FilterExpr: `((payload.finding_class == "VULNERABILITY") || (payload.parent_display_name == "Risk Engine")) ? false : true`,
```
matoszz: why would you be coding this like this?

internal/integrations/definitions/githubapp/mappings.go
```diff
@@ -142,7 +142,8 @@ var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountCanonicalEmail, Expr: `payload.CanonicalEmail`},
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDisplayName, Expr: `payload.Name != "" ? payload.Name : payload.Login`},
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountAvatarRemoteURL, Expr: `payload.AvatarURL`},
-	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `payload.Org`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `"Github"`},
```
matoszz: I really don't think we should be putting this in a CEL expression; if you want to hard code the name use the integration display name at the very least

internal/integrations/definitions/googleworkspace/mappings.go
```diff
@@ -20,7 +20,9 @@ var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDisplayName, Expr: `'name' in payload && 'fullName' in payload.name ? payload.name.fullName : ""`},
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountGivenName, Expr: `'name' in payload && 'givenName' in payload.name ? payload.name.givenName : ""`},
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountFamilyName, Expr: `'name' in payload && 'familyName' in payload.name ? payload.name.familyName : ""`},
-	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `'customerId' in payload ? payload.customerId : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `"Google Workspace"`},
```
matoszz: Same comment as github; this is an inappropriate place to hard code a value like this and should be derived from an already set name like the integration display name (if we set it that way at all)

internal/integrations/definitions/slack/mappings.go
```diff
@@ -17,7 +17,8 @@ var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountAvatarRemoteURL, Expr: `'avatar_url' in payload ? payload.avatar_url : ""`},
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountMfaState, Expr: `dyn('has_2fa' in payload && payload.has_2fa ? "ENABLED" : "DISABLED")`},
 	{Key: integrationgenerated.IntegrationMappingDirectoryAccountStatus, Expr: `dyn('deleted' in payload && payload.deleted ? "INACTIVE" : "ACTIVE")`},
-	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `'team_id' in payload ? payload.team_id : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `"Slack"`},
```
matoszz: same comment as the others, not appropriate to set here and in this way

internal/integrations/operations/dispatcher.go
```diff
@@ -80,6 +80,11 @@ func Dispatch(ctx context.Context, reg *registry.Registry, db *ent.Client, runti
 		}
 	}
 
+	tags := []string{}
+	if metadata.DefinitionID != "" {
```
matoszz: why would this be set in a tag?

internal/integrations/operations/ingest_control_persist.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+
+	ent "github.com/theopenlane/core/internal/ent/generated"
+	control "github.com/theopenlane/core/internal/ent/generated/control"
+)
+
+// persistControlInput upserts one Control record using the ingest lookup key fields
+func persistControlInput(ctx context.Context, db *ent.Client, integration *ent.Integration, createInput ent.CreateControlInput) error {
```
matoszz: what possible use case would we have for creating a control especially from an external system?

internal/integrations/operations/reconcile.go
```diff
@@ -47,9 +48,19 @@ func RegisterReconcileListener(runtime *gala.Gala, handle ReconcileHandler, sche
...
+					logx.FromContext(ctx.Context).Error().Err(execErr).Msg("integration not found, not queuing")
+					return river.JobCancel(execErr)
+				}
+
 				logx.FromContext(ctx.Context).Warn().Err(execErr).Str("integration_id", envelope.IntegrationID).Str("operation", envelope.Operation).Int("error_streak", envelope.Schedule.ErrorStreak+1).Msg("reconcile cycle failed, scheduling retry with backoff")
 			}
 
+			tags := []string{}
+			if envelope.DefinitionID != "" {
+				tags = append(tags, envelope.DefinitionID)
```
matoszz: same question as above curious as to the value of having the definition ID here

internal/integrations/operations/upsert.go
```diff
@@ -11,6 +13,8 @@ import (
 func roundTripUpdateInput[Create any, Update any](createInput Create) (Update, error) {
 	var updateInput Update
 	if err := jsonx.RoundTrip(createInput, &updateInput); err != nil {
+		log.Error().Err(err).Msg("integration: invalid ingest")
```
matoszz, line 16: isn't telling you much

internal/integrations/runtime/webhooks.go
```diff
@@ -153,13 +154,27 @@ func (r *Runtime) DispatchWebhookEvent(ctx context.Context, integration *ent.Int
 		DeliveryID:    event.DeliveryID,
 	}
 
+	tags := []string{}
+	if integration.Name != "" {
+		tags = append(tags, strcase.UpperSnakeCase(integration.Name))
+	}
+
+	if integration.Family != "" {
```
matoszz, line 162: not sure how this is valuable to have

internal/integrations/runtime/webhooks.go
```diff
@@ -153,13 +154,27 @@ func (r *Runtime) DispatchWebhookEvent(ctx context.Context, integration *ent.Int
...
+		tags = append(tags, strcase.UpperSnakeCase(integration.Family))
+	}
+
+	if metadata.DefinitionID != "" {
+		tags = append(tags, metadata.DefinitionID)
```
matoszz: same comment as above

internal/integrations/definitions/awssecurityhub/installation.go
```diff
@@ -4,27 +4,41 @@ import (
...
+	}
+
 	return InstallationMetadata{
-		RoleARN:       assumeRole.RoleARN,
-		HomeRegion:    assumeRole.HomeRegion,
-		AccountID:     assumeRole.AccountID,
-		AccountScope:  assumeRole.AccountScope,
-		AccountIDs:    assumeRole.AccountIDs,
-		LinkedRegions: assumeRole.LinkedRegions,
+		HomeRegion: serviceAccount.HomeRegion,
```
matoszz: why is this relevant to set for a service account?

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+func titleCaseCEL(expr string) string {
+	return fmt.Sprintf(
+		`(%s).split("_").map(w, w.substring(0, 1).upperAscii() + w.substring(1).lowerAscii()).join(" ")`,
+		expr,
+	)
+}
+
 // mapExprFinding is the CEL mapping expression for AWS Security Hub finding payloads
 var mapExprFinding = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingFindingSource, Expr: `"aws-security-hub"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+
+	{Key: integrationgenerated.IntegrationMappingFindingExternalID, Expr: `'Id' in payload ? payload.Id : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingDisplayName, Expr: `'Title' in payload ? payload.Title : ""`},
+	{Key: integrationgenerated.IntegrationMappingFindingOpen, Expr: `'Workflow' in payload && 'Status' in payload.Workflow ? (payload.Workflow.Status == "NEW" || payload.Workflow.Status == "NOTIFIED") : false`},
+
+	{Key: integrationgenerated.IntegrationMappingFindingFindingClass, Expr: `"THREAT"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityVulnerabilityStatusName, Expr: `'Workflow' in payload && payload.Workflow != null && 'Status' in payload.Workflow && payload.Workflow.Status != "" ? payload.Workflow.Status : ('RecordState' in payload ? payload.RecordState : "")`},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilitySeverity, Expr: `'Severity' in payload && payload.Severity != null && 'Label' in payload.Severity ? payload.Severity.Label : ""`},
+
+	// Open if Workflow.Status is NEW or NOTIFIED
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityOpen, Expr: `'Workflow' in payload && 'Status' in payload.Workflow ? (payload.Workflow.Status == "NEW" || payload.Workflow.Status == "NOTIFIED") : false`},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityVulnerabilityStatusName, Expr: `'Workflow' in payload && 'Status' in payload.Workflow ? (payload.Workflow.Status in ["NEW", "NOTIFIED"] ? "Open" : payload.Workflow.Status == "SUPPRESSED" ? "Suppressed" : payload.Workflow.Status == "RESOLVED" ? "Resolved" : "") : ""`},
```
matoszz: these both seem redundant mappings to ones above

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityExternalURI, Expr: `'Vulnerabilities' in payload && size(payload.Vulnerabilities) > 0 && payload.Vulnerabilities[0] != null && 'ReferenceUrls' in payload.Vulnerabilities[0] && size(payload.Vulnerabilities[0].ReferenceUrls) > 0 ? payload.Vulnerabilities[0].ReferenceUrls[0] : ""`},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityDiscoveredAt, Expr: `'CreatedAt' in payload ? payload.CreatedAt : null`},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilitySourceUpdatedAt, Expr: `'UpdatedAt' in payload ? payload.UpdatedAt : null`},
-	{Key: integrationgenerated.IntegrationMappingVulnerabilityOpen, Expr: `dyn('RecordState' in payload ? payload.RecordState == "ACTIVE" : false)`},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityExternalURI, Expr: `'Remediation' in payload && 'Recommendation' in payload.Remediation && 'Url' in payload.Remediation.Recommendation ? payload.Remediation.Recommendation.Url : ""`},
+
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityDiscoveredAt, Expr: `'FirstObservedAt' in payload ? payload.FirstObservedAt : ('CreatedAt' in payload ? payload.CreatedAt : null)`},
```
matoszz: createdAt and firstObservedAt are distinctly different things and shouldn't be mapped into the same field

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingVulnerabilitySourceUpdatedAt, Expr: `'UpdatedAt' in payload ? payload.UpdatedAt : ('LastObservedAt' in payload ? payload.LastObservedAt : null)`},
```
matoszz: same comment as above UpdatedAt and LastObservedAt are 2 totally distinctly different things

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityFirstPatchedVersion, Expr: `'Vulnerabilities' in payload && size(payload.Vulnerabilities) > 0 && 'VulnerablePackages' in payload.Vulnerabilities[0] && size(payload.Vulnerabilities[0].VulnerablePackages) > 0 && 'FixedInVersion' in payload.Vulnerabilities[0].VulnerablePackages[0] ? payload.Vulnerabilities[0].VulnerablePackages[0].FixedInVersion : ""`},
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityReferences, Expr: `'Vulnerabilities' in payload && size(payload.Vulnerabilities) > 0 && 'ReferenceUrls' in payload.Vulnerabilities[0] ? payload.Vulnerabilities[0].ReferenceUrls : []`},
+
+	// CVSS score from Vulnerabilities[0].Cvss
+	{Key: integrationgenerated.IntegrationMappingVulnerabilityScore, Expr: `'Vulnerabilities' in payload && size(payload.Vulnerabilities) > 0 && 'Cvss' in payload.Vulnerabilities[0] && size(payload.Vulnerabilities[0].Cvss) > 0 ? (payload.Vulnerabilities[0].Cvss.filter(c, 'Version' in c && c.Version.startsWith("3")).size() > 0 ? payload.Vulnerabilities[0].Cvss.filter(c, 'Version' in c && c.Version.startsWith("3"))[0].BaseScore : payload.Vulnerabilities[0].Cvss[0].BaseScore) : 0.0`},
```
matoszz, line 58: similar to other comments we should not be falling back to various cvss scores but mapping a single score into the field

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
-	{Key: integrationgenerated.IntegrationMappingFindingDescription, Expr: `'name' in payload && 'complianceType' in payload ? payload.name + " (" + payload.complianceType + ")" : ('name' in payload ? payload.name : "")`},
-	{Key: integrationgenerated.IntegrationMappingFindingFindingStatusName, Expr: `'status' in payload ? payload.status : ""`},
-	{Key: integrationgenerated.IntegrationMappingFindingFindingClass, Expr: `dyn("compliance_assessment")`},
-	{Key: integrationgenerated.IntegrationMappingFindingReportedAt, Expr: `'creationTime' in payload ? payload.creationTime : null`},
-	{Key: integrationgenerated.IntegrationMappingFindingSourceUpdatedAt, Expr: `'lastUpdated' in payload ? payload.lastUpdated : null`},
-	{Key: integrationgenerated.IntegrationMappingFindingRawPayload, Expr: "payload"},
+// mapExprConfigControl is the CEL mapping expression for enriched Config rule payloads mapped to controls
+// payload.rule holds the ConfigRule; payload.catalog holds the optional ControlSummary enrichment
+var mapExprConfigControl = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingControlSource, Expr: `"IMPORTED"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingControlSourceName, Expr: `"aws-config-rule"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingControlControlKindName, Expr: `'catalog' in payload && payload.catalog != null && 'Behavior' in payload.catalog ? ` + titleCaseCEL("payload.catalog.Behavior") + ` : ""`},
+	{Key: integrationgenerated.IntegrationMappingControlCategory, Expr: `'common' in payload && payload.common != null && 'Domain' in payload.common && payload.common.Domain != null && 'Name' in payload.common.Domain ? payload.common.Domain.Name : ""`},
+	{Key: integrationgenerated.IntegrationMappingControlSubcategory, Expr: `'common' in payload && payload.common != null && 'Objective' in payload.common && payload.common.Objective != null && 'Name' in payload.common.Objective ? payload.common.Objective.Name : ""`},
+})
+
+// mapExprConfigControlHealth is the CEL mapping expression for Config rule compliance health payloads
+var mapExprConfigControlHealth = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingControlHealthSource, Expr: `"aws-config-rule"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingControlHealthParentExternalID, Expr: `'rule' in payload && 'ConfigRuleId' in payload.rule && payload.rule.ConfigRuleId != null ? payload.rule.ConfigRuleId : ""`},
+	{Key: integrationgenerated.IntegrationMappingControlHealthCheckStatus, Expr: `'compliance' in payload && payload.compliance != null && 'ComplianceType' in payload.compliance ? (payload.compliance.ComplianceType == "COMPLIANT" ? "COMPLIANT" : payload.compliance.ComplianceType == "NON_COMPLIANT" ? "NOT_COMPLIANT" : "UNKNOWN") : "UNKNOWN"`},
```
matoszz: a really really bad idea to be using "Compliant" and "non-compliant" these should be the literal string status set by AWS not a weird overlay mapping into an enum we create

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingControlHealthExternalURI, Expr: `'rule' in payload && 'ConfigRuleArn' in payload.rule && payload.rule.ConfigRuleArn != null && payload.rule.ConfigRuleArn != "" && 'ConfigRuleName' in payload.rule && payload.rule.ConfigRuleName != null ? "https://" + payload.rule.ConfigRuleArn.split(":")[3] + ".console.aws.amazon.com/config/home?region=" + payload.rule.ConfigRuleArn.split(":")[3] + "#/rules/details?configRuleName=" + payload.rule.ConfigRuleName : ""`},
+})
+
+// mapExprDirectoryAccount maps AWS IAM user payloads to DirectoryAccount
+var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDirectoryName, Expr: `"AWS IAM"`},
```
matoszz: shouldn't be hard coded

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountExternalID, Expr: `'id' in payload ? payload.id : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDisplayName, Expr: `'userName' in payload && payload.userName != "" ? payload.userName : ('id' in payload ? payload.id : "")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountCanonicalEmail, Expr: `'tags' in payload && payload.tags != null && 'Email' in payload.tags ? payload.tags.Email : ('userName' in payload ? payload.userName : "")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountStatus, Expr: `dyn("ACTIVE")`},
```
matoszz: so just set everything to active?

internal/integrations/definitions/awssecurityhub/mappings.go
```diff
@@ -1,61 +1,211 @@
...
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountOrganizationUnit, Expr: `'path' in payload ? payload.path : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountProfile, Expr: "payload"},
+})
+
+// mapExprDirectoryGroup maps AWS IAM group payloads to DirectoryGroup
+var mapExprDirectoryGroup = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupExternalID, Expr: `'id' in payload ? payload.id : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupDisplayName, Expr: `'name' in payload && payload.name != "" ? payload.name : ('id' in payload ? payload.id : "")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupStatus, Expr: `dyn("ACTIVE")`},
```
matoszz: same comment as above

internal/integrations/definitions/awssecurityhub/operation_directory_sync.go
```diff
@@ -0,0 +1,298 @@
...
+type iamMembershipPayload struct {
+	// Group is the group side of the membership
+	Group iamEntityRef `json:"group"`
+	// Member is the user side of the membership
+	Member iamEntityRef `json:"member"`
+}
+
+// DirectorySync collects AWS IAM users, groups, and memberships for directory account ingest
+type DirectorySync struct {
+	cfg Config
```
matoszz: What config are we surfacing to the user that we surface in other directory sync operations? and why are you defining intermediary types like this? the point of using CEL and the json payload parsing is to avoid re-mapping all of the data we query in external systems into new type definitions, these shouldn't be required (and really would represent an antipattern)

internal/integrations/definitions/awssecurityhub/types.go
```diff
@@ -19,20 +20,28 @@ var (
...
 )
 
 // UserInput holds installation-specific configuration collected from the user
 type UserInput struct {
 	// FilterExpr limits imported records to envelopes matching the CEL expression
 	FilterExpr string `json:"filterExpr,omitempty" jsonschema:"title=Filter Expression,description=Optional CEL expression to apply to records before ingesting"`
+	// EnableGroupSync controls whether group and membership records are collected
+	EnableGroupSync bool `json:"enableGroupSync,omitempty" jsonschema:"title=Sync Groups"`
+	// EnableUserSync controls whether users are collected
+	EnableUserSync bool `json:"enableUserSync,omitempty" jsonschema:"title=Sync Users"`
```
matoszz: what purpose is this serving? We wouldn't sync groups without users, and the integrations existence is the trigger to perform the action why have a totally separate configuration you'd set?

internal/integrations/definitions/gcpscc/config.go
```diff
@@ -0,0 +1,5 @@
+package gcpscc
+
+// Config holds operator-level credentials for the GCP SCC definition
+type Config struct {
```
matoszz: why?

matoszz (changes requested):

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: Not sure what this comment means? It's parsing a risk?

internal/integrations/definitions/googleworkspace/mappings.go
golanglemonade: We had talked in slack about setting it; we cannot use the integration name afterwards so we need to set a directory name?

internal/integrations/operations/dispatcher.go
golanglemonade: so we can filter in the riverui on tags.

internal/integrations/operations/reconcile.go
golanglemonade: same comment; there is nothing to filter the river ui today to find the job I want, its just thousands of "gala_event_v1: with an envelope arg. I PR'ed riverui to allow filtering on tags, and set tags so I could actually find the one I'm looking for.

internal/integrations/operations/upsert.go
golanglemonade: It's logging the actually error - otherwise you just get `ErrIngestMappedDocumentInvalid` logged later with no idea what it means.

common/enums/exporttype.go
golanglemonade: well I tried "control status" but ran into issues with `controlStatuses` vs `controlStatusSlice` and gave up after fighting generation for an hour

internal/integrations/definitions/awssecurityhub/mappings.go
golanglemonade: Which? `IntegrationMappingVulnerabilityVulnerabilityStatusName` is only here once?

internal/integrations/definitions/awssecurityhub/mappings.go
golanglemonade: It's not falling back? It's using the `BaseScore` field from the `Vulnerabilities` object

internal/integrations/definitions/awssecurityhub/mappings.go
golanglemonade: Missed this, will fix

golanglemonade: Actually looks like AWS doesn't have a state, you just delete users. So to align with other records I have it set to active. There are other fields like last seen at I was going to add next.

internal/integrations/definitions/awssecurityhub/mappings.go
golanglemonade: removed.

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: Pulling resource from gcp that the finding is for?

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: I'm not sure what you mean, it's not falling back its using `payload.vulnerability.cve.cvssv3.base_score`

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: removed

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: IntegrationMappingVulnerabilitySummary? Its only here once?

internal/integrations/definitions/gcpscc/config.go
golanglemonade: cruft, removed

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: meaning just do ``` Expr: ` payload.vulnerability.cve.upstream_fix_available`},``` 
that breaks when the vulnerability field isn't there?

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: removed; added to the `persistDirectoryAccountInput` logic instead.

internal/integrations/definitions/gcpscc/mappings.go
matoszz: this, specifically: 

```
Expr: `'kubernetes' in payload && size(payload.kubernetes) > 0 ? payload.kubernetes : 'containers' in payload && size(payload.containers) > 0 ? payload.containers : null`}
```
We're filtering all `FindingTargetDetails` to needing to be kubernetes specific?

internal/integrations/definitions/gcpscc/mappings.go
matoszz: Sorry, to clarify:
- We should not default to a score of 0.0
- We should ideally map into CVSS v4 or alternatively update github / the other vuln upsert logic to go back to v3, or create multiple fields associated to the versioning
- The expression rule itself is needlessly large given all of the payloads are going to have `vulnerability.cve.cvssv3` not equal to null and always present; the check should be if `vulernability.cve.id != "" || vulnerability.cve.id != null`; GCP's payload comes across like this:
```
    "vulnerability": {
      "cve": {
        "id": "",
        "references": [],
        "cvssv3": {
          "baseScore": 0,
          "attackVector": "ATTACK_VECTOR_UNSPECIFIED",
          "attackComplexity": "ATTACK_COMPLEXITY_UNSPECIFIED",
          "privilegesRequired": "PRIVILEGES_REQUIRED_UNSPECIFIED",
          "userInteraction": "USER_INTERACTION_UNSPECIFIED",
          "scope": "SCOPE_UNSPECIFIED",
          "confidentialityImpact": "IMPACT_UNSPECIFIED",
          "integrityImpact": "IMPACT_UNSPECIFIED",
          "availabilityImpact": "IMPACT_UNSPECIFIED"
        },
        "upstreamFixAvailable": false,
        "impact": "RISK_RATING_UNSPECIFIED",
        "exploitationActivity": "EXPLOITATION_ACTIVITY_UNSPECIFIED",
        "observedInTheWild": false,
        "zeroDay": false,
        "exploitReleaseDate": "1970-01-01T00:00:00Z",
        "firstExploitationDate": "1970-01-01T00:00:00Z"
      },
```
- As far as I know the payload shape is `attack_exposure.score` not `attack_exposure_score` ex: 
```
    "attackExposure": {
      "score": 0,
      "latestCalculationTime": "1970-01-01T00:00:00Z",
      "attackExposureResult": "",
      "state": "STATE_UNSPECIFIED",
      "exposedHighValueResourcesCount": 0,
      "exposedMediumValueResourcesCount": 0,
      "exposedLowValueResourcesCount": 0
    },
```

internal/integrations/definitions/gcpscc/mappings.go
matoszz: ? the vulnerability field is always in the payload, and upstream fix available defaults to false already. https://securitycenter.googleapis.com/$discovery/rest?version=v2

internal/integrations/definitions/gcpscc/mappings.go
matoszz: What upstream object type are you querying that's reporting a risk which is separate from vulnerability and/or finding?

internal/integrations/definitions/googleworkspace/mappings.go
matoszz: what do you mean "afterwards" the integration definition is the thing performing the upsert, the metadata is definitely available to pass through

internal/integrations/definitions/awssecurityhub/mappings.go
matoszz: Why map it at all? the default of the ent schema is DirectoryAccountStatusActive

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: I added more and also updated the FindingTargets with the resource name; its no longer limited to just kubernetes or containers but also will check processes, databases, files, etc.

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: This is all from GetFindings - SCC returns all and you have to filter on finding_class or parent display name (source)

internal/integrations/definitions/googleworkspace/mappings.go
golanglemonade: resolved, will push an update shortly

internal/integrations/operations/ingest_control_persist.go
golanglemonade: removed

internal/integrations/definitions/awssecurityhub/installation.go
golanglemonade: removed

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: removed

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/883

### Changes Preview:

- 🔄 Merged Helm values.yaml

internal/integrations/definitions/gcpscc/mappings.go
golanglemonade: Switched to v3

matoszz (dismissed):

matoszz (approved):

PR #2325, opened 2026-04-23
Vendor Intake Questionnaire

adelowo:
- Added a new `transform_configuration` to `templates` schema. this will contain a json schema definition  that can convert a submitted assessment into an object/schema in the system ( only entity supported right now ).

there is also an `enabled` field there to determine if to temporarily disable the transforms while still keeping the config.

in this example below, the submitted data in `name` will map to the `displayName` of entity schema and the value in `owner.name` of the assessment response will map to `internal_owner`.

- Also made email optional when sending questionnaires out because for assessments like `vendor intake`, we want to have a link that can be shared as many times and would work.
- Added a new `accessURL` in assessment that will only be filled with the reusable link 

> Long term, we may want to put this on the assessments too directly just like we copy over `jsonconfig` to the assessments when cloning a template and allow users map themselves fields themselves when they edit the `jsonconfig` but for now since we are managing it, this is fine

```
{
  "target": "ENTITY",
  "enabled": true,
  "mappings": [
    {
      "to": "display_name",
      "from": "name",
      "required": true
    },
    {
      "to": "name",
      "from": "name",
      "required": true,
      "transform": "SLUGIFY"
    },
    {
      "to": "description",
      "from": "description",
      "transform": "STRING"
    },
    {
      "to": "contract_start_date",
      "from": "contract_start_date",
      "transform": "DATE"
    },
    {
      "to": "contract_end_date",
      "from": "contract_end_date",
      "transform": "DATE"
    },
    {
      "to": "has_soc2",
      "from": "has_soc2",
      "transform": "BOOL"
    },
    {
      "to": "annual_spend",
      "from": "cost_amount",
      "transform": "FLOAT"
    },
    {
      "to": "billing_model",
      "from": "cost_frequency",
      "transform": "STRING"
    },
    {
      "to": "links",
      "from": "url",
      "transform": "STRING_ARRAY"
    },
    {
      "to": "environment_id",
      "from": "environment",
      "resolver": "environment"
    },
    {
      "to": "internal_owner",
      "from": "owner.email",
      "resolver": "internal_owner"
    },
    {
      "to": "notes",
      "from": "additional_information",
      "transform": "STRING"
    },
    {
      "to": "status",
      "from": "status",
      "transform": "STRING"
    }
  ],
  "operation": "UPDATE"
}
```


internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,919 @@
...
+	AnnualSpend          *float64
+	BillingModel         string
+	Links                []string
+	Notes                string
+	EnvironmentName      string
+	EnvironmentID        string
+	InternalOwner        string
+	InternalOwnerUserID  string
+	InternalOwnerGroupID string
+}
```
golanglemonade: Why do we need to define this separately instead of just using the generated types?

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,919 @@
...
+	}
+
+	input, err := entityFromMapping(values)
+	if err != nil {
+		return err
+	}
+
+	if input.Status == "" {
+		input.Status = getEntityStatusFromContract(input.ContractStartDate)
+	}
```
golanglemonade: I dont think we should do this; we should take the user defined values only instead of trying to infer from other data.

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,919 @@
...
+		return fmt.Errorf("resolve environment: %w", err)
+	}
+
+	if entgen.IsNotFound(err) {
+		enum, err = client.CustomTypeEnum.Create().
+			SetName(environment).
+			SetField("environment").
+			SetObjectType("").
+			SetOwnerID(organizationID).
+			Save(ctx)
```
golanglemonade, line 399: We shouldn't create this, can we have the questionnaire contain the options as multiple choice instead?

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,919 @@
...
+
+	case strings.EqualFold(transform.String(), enums.TemplateProjectionTransformStringArray.String()):
+
+		return getStringArrayValue(value), nil
+	default:
+		return nil, &questionnaireValidationError{Message: fmt.Sprintf("unsupported transform %q", transform)}
+	}
+}
+
+func entityFromMapping(values map[string]any) (entityTransform, error) {
```
golanglemonade: I dont think we should be manually maintaining this switch statement. Take a look at how the integrations does ingest of data using `json.RawMessage` and converts to the typed input. We might even just be able to reuse the same functions and mimic this as an "integration" ingest except instead of from an integration, the "Questionnaire" is the integration with the mappings.

cmd/serve.go
```diff
@@ -167,6 +167,7 @@ func serve(ctx context.Context) error {
 		ent.SessionConfig(so.Config.Handler.SessionConfig),
 		ent.EntConfig(&so.Config.Settings.EntConfig),
 		ent.EntitlementManager(so.Config.Handler.Entitlements),
+		ent.QuestionnaireProductURL(so.Config.Settings.Integrations.Email.ProductURL),
```
golanglemonade: this should just go in the existing ent config

common/enums/templatekind.go
```diff
@@ -8,13 +8,19 @@ type TemplateKind string
 var (
 	// TemplateKindQuestionnaire indicates the questionnaire.
 	TemplateKindQuestionnaire TemplateKind = "QUESTIONNAIRE"
-	// TemplateKindTrustCenterNda indicates the trust center NDA.
+	// TemplateKindTrustCenterNda indicates the trustcenter nda.
 	TemplateKindTrustCenterNda TemplateKind = "TRUSTCENTER_NDA"
+	// TemplateKindVendorIntake indicates the vendor intake.
+	TemplateKindVendorIntake TemplateKind = "VENDOR_INTAKE"
```
golanglemonade: This is too specific; we don't want to have to create a new enum for every object type. If we need an enum here it should be something like `external` or `object` so that we can use for any type and not specific to just vendors
```suggestion
	TemplateKindExternalIntake TemplateKind = "EXTERNAL_INTAKE"
```

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/922

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

internal/ent/hooks/listeners_questionnaire_transform.go
adelowo: the template i have added to harmonize contains this which is why the custom enum bit was added. except I get rid of the custom text and leave only the radio group

image

internal/ent/hooks/listeners_questionnaire_transform.go
adelowo: embedded the generated type but kept the `Notes`

internal/ent/hooks/assessment.go
```diff
@@ -50,7 +50,8 @@ func HookQuestionnaireAssessment() ent.Hook {
 				return nil, err
 			}
 
-			if template.Kind != enums.TemplateKindQuestionnaire {
+			// longterm we want to use the transformconfig check on template instead
+			if template.Kind != enums.TemplateKindExternalIntake && template.Kind != enums.TemplateKindQuestionnaire {
```
golanglemonade: I think this check is to avoid `TRUSTCENTER_NDA` types, we should be able to switch it

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,809 @@
...
+	integrationtypes "github.com/theopenlane/core/internal/integrations/types"
+	"github.com/theopenlane/core/internal/workflows"
+	"github.com/theopenlane/core/pkg/gala"
+	"github.com/theopenlane/core/pkg/logx"
+)
+
+// RegisterGalaQuestionnaireTransformListeners registers listeners that transform
+// completed questionnaire document data into configured target schemas.
+// this supports only entities for now
+func RegisterGalaQuestionnaireTransformListeners(registry *gala.Registry) ([]gala.ListenerID, error) {
```
golanglemonade, line 41:
```suggestion
// RegisterGalaQuestionnaireTransformListeners registers listeners that transform
// completed questionnaire document data into configured target schemas.
// Supported types are defined in `TemplateProjectionTarget` enums
func RegisterGalaQuestionnaireTransformListeners(registry *gala.Registry) ([]gala.ListenerID, error) {
```

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,809 @@
...
+			query.WithTemplate()
+		}).
+		Only(allowCtx)
+	if err != nil {
+		if entgen.IsNotFound(err) {
+			return nil
+		}
+
+		return err
+	}
```
golanglemonade: Why are we silently ignoring the error when its not found? 

Can we log the error here when this fails?

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,809 @@
...
+
+	config := assessment.Edges.Template.TransformConfiguration
+	if !config.Enabled {
+		return nil
+	}
+
+	document := response.Edges.Document
+	if document == nil {
+		return nil
+	}
```
golanglemonade: Can this be pulled out in a `validateQuestionnaire` function to reduce the overall complexity of `handleAssessmentResponse`?

internal/ent/hooks/listeners_questionnaire_transform.go
```diff
@@ -0,0 +1,809 @@
...
+	if req.OrganizationID == "" {
+		return &questionnaireValidationError{Message: "missing transform organization id"}
+	}
+
+	switch {
+	case strings.EqualFold(req.Config.Target.String(), enums.TemplateProjectionTargetEntity.String()):
+		return handleEntityTransform(ctx, client, req)
+	default:
+		return &questionnaireValidationError{Message: fmt.Sprintf("unsupported transform target %q", req.Config.Target)}
+	}
```
golanglemonade: These are typed enums; shouldn't need to do a string comparison
```suggestion
	switch eq.Config.Target {
	case enums.TemplateProjectionTargetEntity:
		return handleEntityTransform(ctx, client, req)
	default:
		return &questionnaireValidationError{Message: fmt.Sprintf("unsupported transform target %q", req.Config.Target)}
	}
```

internal/ent/hooks/listeners_questionnaire_transform.go
adelowo: if for some reason the entity is deleted before gala execution runs. but will add a log there.

for the `err`, it is bubbled up which is why I left it out, but will add now

internal/ent/privacy/rule/invite.go
```diff
@@ -19,8 +20,13 @@ const (
 )
 
 // CanInviteUsers is a rule that returns allow decision if user has access to invite members or admins to the organization
-func CanInviteUsers() privacy.InviteMutationRuleFunc {
-	return privacy.InviteMutationRuleFunc(func(ctx context.Context, m *generated.InviteMutation) error {
+func CanInviteUsers() privacy.MutationRuleFunc {
```
golanglemonade: Why this change?

internal/ent/schema/entity.go
```diff
@@ -78,7 +78,7 @@ func (Entity) Fields() []ent.Field {
 			Optional().
 			Annotations(
 				entx.FieldSearchable(),
-				entgql.Skip(entgql.SkipWhereInput),
```
golanglemonade, line 81: Is this needed for integration code?

internal/httpserve/handlers/questionnaire.go
```diff
@@ -44,24 +44,30 @@ func (h *Handler) GetQuestionnaire(ctx echo.Context, openapi *OpenAPIContext) er
...
-	if err != nil && !generated.IsNotFound(err) {
-		logx.FromContext(reqCtx).Err(err).Msg("could not fetch assessment response")
-		return h.InternalServerError(ctx, ErrProcessingRequest, openapi)
+	var (
+		assessmentResponse *generated.AssessmentResponse
+		err                error
+	)
+
+	// email not required because we can generate anon links that can now be shared
+	// but if it exists, verify that it matche the response we created when we sent it out
```
golanglemonade:
```suggestion
	// but if it exists, verify that it matches the response we created when we sent it out
```

internal/ent/privacy/rule/invite.go
adelowo: There was a compilation/type/signature issue if I recall correctly. Can't recall the exact issue but other PRs pass actually now so can revert

internal/httpserve/handlers/questionnaire.go
```diff
@@ -44,24 +44,30 @@ func (h *Handler) GetQuestionnaire(ctx echo.Context, openapi *OpenAPIContext) er
...
-		Only(allowCtx)
-	if err != nil && !generated.IsNotFound(err) {
-		logx.FromContext(reqCtx).Err(err).Msg("could not fetch assessment response")
-		return h.InternalServerError(ctx, ErrProcessingRequest, openapi)
+	var (
+		assessmentResponse *generated.AssessmentResponse
+		err                error
+	)
+
+	// email not required because we can generate anon links that can now be shared
```
golanglemonade, line 57: I think the _response_ should still have an email; can that be a required field on the assessment itself?

internal/ent/schema/entity.go
adelowo: Yes it is. The "description" field of the integration wasn't getting generated with that annotation.

internal/httpserve/handlers/questionnaire.go
golanglemonade: chatted in slack; this comes in via the template/document data

golanglemonade (approved):

PR #2334, opened 2026-04-27
Update Edge Access Check for Identity Holder for directory edges

adelowo:


internal/ent/schema/directorymembership.go
```diff
@@ -60,6 +60,10 @@ func (DirectoryMembership) Fields() []ent.Field {
 			Comment("stable external workspace, tenant, or installation identifier used to correlate memberships across multiple integrations pointed at the same directory instance").
 			Optional().
 			Nillable(),
+		field.String("identity_holder_id").
```
golanglemonade: Why were these edges added here? Should have just needed to update the edge on identity holder?

golanglemonade (approved):

Issue #2342, opened 2026-04-30
[Integration] Authentik -  Directory Sync


golanglemonade:
Adds a `DirectorySync` integration for `Authentik`. This should ingest directory accounts, groups, and memberships. 

The config should: 
-  include a `DisableGroup` sync to allow for only account ingestion
- Include a `PrimaryDirectory` option to mark this integration as the authoritative source for the identity holder record

See the google workspace integration for an example of a working directory sync integraiton.

golanglemonade:
@manab-pr here you go

manab-pr:
@golanglemonade yep , assign it to me . : )

PR #2347, opened 2026-05-03
Directory Sync  : authentik integration 

manab-pr:
closes : Directory Sync #2342 

Directory Sync **Authentik** implemenation


manab-pr:
@golanglemonade  there are two things though : first , check the end points , i have put them in client.go , all the other integrations are using sdk so i couldn't find a reference where exactly should i put them , and second is : once you verify it and everything looks good , i'll write the documentation .

internal/integrations/definitions/authentik/client.go
```diff
@@ -0,0 +1,192 @@
...
+const (
+	// authentikRequestTimeout is the per-request timeout in seconds
+	authentikRequestTimeout = 30 * time.Second
+	// authentikRateLimitMaxRetries is the maximum number of retries on rate limit
+	authentikRateLimitMaxRetries = 3
+	// authentikRetryBaseDelay is the base delay between retries
+	authentikRetryBaseDelay = 2 * time.Second
+
+	// authentikUsersEndpoint is the path for the Authentik users list endpoint
+	authentikUsersEndpoint = "/api/v3/core/users/"
```
golanglemonade: Chatted in discord; this is going to be updated to use the go-client instead of using api endpoints directly

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,61 @@
+package authentik
+
+import (
+	"github.com/theopenlane/core/internal/ent/integrationgenerated"
+	"github.com/theopenlane/core/internal/integrations/providerkit"
+	"github.com/theopenlane/core/internal/integrations/types"
+)
+
+// mapExprDirectoryAccount is the CEL mapping expression for Authentik user payloads mapped to DirectoryAccount
+var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
```
golanglemonade, line 10: Other mappings you'll want to include: 

`date_joined`  -> `IntegrationMappingDirectoryAccountAddedAt`
`last_login` -> `IntegrationMappingDirectoryAccountLastSeenAt`
`last_updated` -> `IntegrationMappingDirectoryAccountObservedAt`
`uuid` -> `IntegrationMappingDirectoryAccountExternalID` (instead of `pk`)
`type` -> `IntegrationMappingDirectoryAccountAccountType`
`attributes` -> `IntegrationMappingDirectoryAccountMetadata`

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,61 @@
...
+	{Key: integrationgenerated.IntegrationMappingDirectoryMembershipMetadata, Expr: "payload"},
+})
+
+// authentikMappings returns the built-in Authentik ingest mappings
+func authentikMappings() []types.MappingRegistration {
+	return []types.MappingRegistration{
+		{
+			Schema: integrationgenerated.IntegrationMappingSchemaDirectoryAccount,
+			Spec: types.MappingOverride{
+				FilterExpr: `'type' in payload && (payload.type == "internal" || payload.type == "external")`,
```
golanglemonade: This should be true, users can filter on type if they wanted using the `FilterExpr` in user input
```suggestion
				FilterExpr: "true",
```

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,61 @@
...
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDisplayName, Expr: `'name' in payload && payload.name != null && payload.name != "" ? payload.name : ('username' in payload ? payload.username : "")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountStatus, Expr: `dyn('is_active' in payload ? (payload.is_active ? "ACTIVE" : "INACTIVE") : "INACTIVE")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountProfile, Expr: "payload"},
+})
+
+// mapExprDirectoryGroup is the CEL mapping expression for Authentik group payloads mapped to DirectoryGroup
+var mapExprDirectoryGroup = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupExternalID, Expr: `'pk' in payload ? payload.pk : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupDisplayName, Expr: `'name' in payload && payload.name != null && payload.name != "" ? payload.name : ('pk' in payload ? payload.pk : "")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupEmail, Expr: `""`},
```
golanglemonade:
```suggestion
```

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,61 @@
...
+var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountExternalID, Expr: `'pk' in payload ? string(payload.pk) : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountCanonicalEmail, Expr: `'email' in payload && payload.email != null && payload.email != "" ? payload.email : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountDisplayName, Expr: `'name' in payload && payload.name != null && payload.name != "" ? payload.name : ('username' in payload ? payload.username : "")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountStatus, Expr: `dyn('is_active' in payload ? (payload.is_active ? "ACTIVE" : "INACTIVE") : "INACTIVE")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountProfile, Expr: "payload"},
+})
+
+// mapExprDirectoryGroup is the CEL mapping expression for Authentik group payloads mapped to DirectoryGroup
+var mapExprDirectoryGroup = providerkit.CelMapExpr([]providerkit.CelMapEntry{
```
golanglemonade, line 24: Additional mappings:
`attributes` -> `IntegrationMappingDirectoryGroupMetadata`
`name` is required so you shouldn't need a pk fallback here

internal/integrations/definitions/authentik/client.go
manab-pr: yeah, i'm done with that , only testing is pending 🥳

internal/integrations/definitions/authentik/mappings.go
manab-pr: uuid is only available on the full User model from the list endpoint. PartialUser (embedded in group responses) and UserSelf (me endpoint) both expose uid but not uuid. If uuid is preferred as ExternalID, we could use uid for both accounts and memberships since it's consistently available across all three models , but uid and uuid are different values. Alternatively we stay with pk which is consistent everywhere. What's Say ? should i use uid then ?

internal/integrations/definitions/authentik/client.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// extractScheme extracts the scheme from a base URL
+func extractScheme(baseURL string) string {
+	if strings.HasPrefix(baseURL, "https://") {
+		return "https"
+	}
+
+	return "http"
+}
```
golanglemonade: These should use the `url` lib, e.g for schema:

```go
	u, err := url.Parse(rawURL)
	if err != nil {
		return "", err
	}

	return u.Scheme, nil
```

internal/integrations/definitions/authentik/operation_directory_sync.go
```diff
@@ -0,0 +1,171 @@
...
+			_ = jsonx.UnmarshalIfPresent(request.Integration.Config.ClientConfig, &cfg)
+		}
+
+		return d.Run(ctx, c, cfg)
+	})
+}
+
+// Run collects Authentik directory users, groups, and memberships
+func (DirectorySync) Run(ctx context.Context, c *authentikSDK.APIClient, cfg UserInput) ([]types.IngestPayloadSet, error) {
+	users, err := listDirectoryUsers(ctx, c)
```
golanglemonade: Either in the API call if possible, or in the loop to get the users below, you need to filter on `last_updated` using `request.LastRunAt`

internal/integrations/definitions/authentik/builder.go
```diff
@@ -0,0 +1,93 @@
...
+					Handle:       HealthCheck{}.Handle(),
+				},
+				{
+					Name:         directorySyncOperation.Name(),
+					Description:  "Collect Authentik directory users, groups, and memberships as directory accounts",
+					Topic:        definitionID.OperationTopic(directorySyncOperation.Name()),
+					ClientRef:    authentikClient.ID(),
+					ConfigSchema: directorySyncSchema,
+					Policy:       types.ExecutionPolicy{Reconcile: true},
+					Ingest: []types.IngestContract{
```
golanglemonade, line 77: This needs `SkipDefaultLookback: true,`

internal/integrations/definitions/authentik/mappings_test.go
```diff
@@ -0,0 +1,26 @@
+package authentik
+
+import (
```
golanglemonade, line 3: Can you add mappings tests based on some example API results. You should be able to add `examples/*.json` and follow the pattern here: https://github.com/theopenlane/core/blob/main/internal/integrations/definitions/awssecurityhub/mappings_test.go#L83

internal/integrations/definitions/authentik/types.go
```diff
@@ -0,0 +1,59 @@
...
+	// Token is the Authentik API token
+	Token string `json:"token" jsonschema:"required,title=API Token"`
+}
+
+// UserInput holds installation-specific configuration collected from the user
+type UserInput struct {
+	// DisableGroupSync when true only syncs users, skipping groups and memberships
+	DisableGroupSync bool `json:"disableGroupSync,omitempty" jsonschema:"title=Disable Group Sync,description=Only sync users disable group and membership sync operations"`
+	// PrimaryDirectory marks this installation as the authoritative source for identity holder sync
+	PrimaryDirectory bool `json:"primaryDirectory,omitempty" jsonschema:"title=Primary Directory,description=Mark this as the authoritative source for identity holder enrichment and lifecycle"`
```
golanglemonade: Can you switch the order of these two; the frontend pulls this via JsonSchema definition and we'd want the primary directory toggle first
```suggestion
	// PrimaryDirectory marks this installation as the authoritative source for identity holder sync
	PrimaryDirectory bool `json:"primaryDirectory,omitempty" jsonschema:"title=Primary Directory,description=Mark this as the authoritative source for identity holder enrichment and lifecycle"`
		// DisableGroupSync when true only syncs users, skipping groups and memberships
	DisableGroupSync bool `json:"disableGroupSync,omitempty" jsonschema:"title=Disable Group Sync,description=Only sync users disable group and membership sync operations"`
```

golanglemonade (commented):
Looks good; Some updates needed around last updated so we don't pull the full records on every sync and add some additional tests. 

Any reason why you removed the compose file? It would be nice to have not only the compose file but seed some demo data to enable local testing of this integration.

internal/integrations/definitions/authentik/mappings_test.go
manab-pr: actually i followed Okta, okay will do that .

internal/integrations/definitions/authentik/types.go
manab-pr: sure will do that .

manab-pr:
> Looks good; Some updates needed around last updated so we don't pull the full records on every sync and add some additional tests.
> 
> Any reason why you removed the compose file? It would be nice to have not only the compose file but seed some demo data to enable local testing of this integration.

yeah no reason , i thought i was not supposed to push that file , as i can't see any other test files in root specifically , that's why after testing i removed it , will add that back.

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"github.com/theopenlane/core/internal/ent/integrationgenerated"
+	"github.com/theopenlane/core/internal/integrations/providerkit"
+	"github.com/theopenlane/core/internal/integrations/types"
+)
+
+// mapExprDirectoryAccount is the CEL mapping expression for Authentik user payloads mapped to DirectoryAccount
+var mapExprDirectoryAccount = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+    {Key: integrationgenerated.IntegrationMappingDirectoryAccountExternalID, Expr: `'uid' in payload ? payload.uid : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountCanonicalEmail, Expr: `'email' in payload && payload.email != null && payload.email != "" ? payload.email : ""`},
```
golanglemonade: You don't need `payload.email != ""` since that will be the same as the fallback here
```suggestion
	{Key: integrationgenerated.IntegrationMappingDirectoryAccountCanonicalEmail, Expr: `'email' in payload && payload.email != null ? payload.email : ""`},
```

golanglemonade:
> yeah no reason , i thought i was not supposed to push that file , as i can't see any other test files in root specifically , that's why after testing i removed it , will add that back.

Yeah, thats because other ones are SaaS services; whereas authentik is OSS and we'd want to be able to test locally.

docker/docker-compose-authentik.yml
```diff
@@ -0,0 +1,67 @@
+services:
```
golanglemonade, line 1: Can you put this in the docker directory + add a task command for it?

manab-pr: sure

docker/docker-compose-authentik.yml
```diff
@@ -0,0 +1,67 @@
...
+  postgresql:
+    env_file:
+      - authentik.env
```
golanglemonade: Since this is a dev setup, we can provide this. Add it to `docker/configs/authentik` and add the required env vars, e.g.: 
```
AUTHENTIK_BOOTSTRAP_PASSWORD=akadmin123
```

That way it will just work without additional user setup to test this.

docker/docker-compose-authentik.yml
```diff
@@ -0,0 +1,67 @@
...
+    environment:
+      POSTGRES_DB: ${PG_DB:-authentik}
+      POSTGRES_PASSWORD: ${PG_PASS:?database password required}
```
golanglemonade: Same here, this is a dev env so we can set this up. 

Alternatively, can we use the same postrgres container is is already coming up as part of the main stack and just create the db in there?

docker/docker-compose-authentik.yml
```diff
@@ -0,0 +1,67 @@
...
+      postgresql:
+        condition: service_healthy
+    env_file:
+      - authentik.env
+    environment:
+      AUTHENTIK_POSTGRESQL__HOST: postgresql
+      AUTHENTIK_POSTGRESQL__NAME: ${PG_DB:-authentik}
+      AUTHENTIK_POSTGRESQL__PASSWORD: ${PG_PASS}
+      AUTHENTIK_POSTGRESQL__USER: ${PG_USER:-authentik}
+      AUTHENTIK_SECRET_KEY: ${AUTHENTIK_SECRET_KEY:?secret key required}
```
golanglemonade: This should also be set and not required

internal/integrations/definitions/authentik/installation.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	if cred.Token == "" || cred.BaseURL == "" {
+		return InstallationMetadata{}, false, nil
+	}
+
+	client, err := Client{}.Build(ctx, types.ClientBuildRequest{Credentials: req.Credentials})
+	if err != nil {
+		return InstallationMetadata{}, false, err
+	}
```
golanglemonade, line 27: Log these errors, this was helpful for me when trying to set it up the first time
```suggestion
	if err != nil {
	    logx.FromContext(ctx).Error().Err(err).Msg("error  during installation")
		
		return InstallationMetadata{}, false, err
	}
```

internal/integrations/definitions/authentik/installation.go
```diff
@@ -0,0 +1,42 @@
...
+	apiClient := client.(*authentikSDK.APIClient)
+
+	info, resp, err := apiClient.AdminApi.AdminSystemRetrieve(ctx).Execute()
+	if resp != nil {
+		_ = resp.Body.Close()
+	}
+
+	if err != nil {
+		return InstallationMetadata{}, false, ErrHealthCheckFailed
+	}
```
golanglemonade, line 40:
```suggestion
	if err != nil {
		logx.FromContext(ctx).Error().Err(err).Msg("error checking admin")
		
		return InstallationMetadata{}, false, ErrHealthCheckFailed
	}
```

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,66 @@
...
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountStatus, Expr: `dyn('is_active' in payload ? (payload.is_active ? "ACTIVE" : "INACTIVE") : "INACTIVE")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountAccountType, Expr: `dyn('type' in payload && payload.type != null ? (payload.type == "internal" ? "USER" : (payload.type == "external" ? "GUEST" : (payload.type == "service_account" ? "SERVICE" : (payload.type == "internal_service_account" ? "SERVICE" : "USER")))) : "USER")`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountAddedAt, Expr: `'date_joined' in payload && payload.date_joined != null ? payload.date_joined : null`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountLastSeenAt, Expr: `'last_login' in payload && payload.last_login != null ? payload.last_login : null`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountObservedAt, Expr: `'last_updated' in payload && payload.last_updated != null ? payload.last_updated : null`},
```
golanglemonade:
```suggestion
	{Key: integrationgenerated.IntegrationMappingDirectoryAccountAddedAt, Expr: `'date_joined' in payload ? payload.date_joined`},
	{Key: integrationgenerated.IntegrationMappingDirectoryAccountLastSeenAt, Expr: `'last_login' in payload ? payload.last_login`},
	{Key: integrationgenerated.IntegrationMappingDirectoryAccountObservedAt, Expr: `'last_updated' in payload ? payload.last_updated`},
```
I don't think you need the null check here if thats the fallback

internal/integrations/definitions/authentik/mappings.go
```diff
@@ -0,0 +1,66 @@
...
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountMetadata, Expr: `'attributes' in payload ? payload.attributes : {}`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryAccountProfile, Expr: "payload"},
+})
+
+// mapExprDirectoryGroup is the CEL mapping expression for Authentik group payloads mapped to DirectoryGroup
+var mapExprDirectoryGroup = providerkit.CelMapExpr([]providerkit.CelMapEntry{
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupExternalID, Expr: `'pk' in payload ? payload.pk : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupDisplayName, Expr: `'name' in payload && payload.name != null ? payload.name : ""`},
+	{Key: integrationgenerated.IntegrationMappingDirectoryGroupClassification, Expr: `dyn('is_superuser' in payload && payload.is_superuser ? "SECURITY" : "TEAM")`},
```
golanglemonade:
```suggestion
```

I would just remove this and use the default; `SECURITY` isn't really the right thing here

internal/integrations/definitions/authentik/types.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// UserInput holds installation-specific configuration collected from the user
+type UserInput struct {
+	// PrimaryDirectory marks this installation as the authoritative source for identity holder sync
+	PrimaryDirectory bool `json:"primaryDirectory,omitempty" jsonschema:"title=Primary Directory,description=Mark this as the authoritative source for identity holder enrichment and lifecycle"`
+	// DisableGroupSync when true only syncs users, skipping groups and memberships
+	DisableGroupSync bool `json:"disableGroupSync,omitempty" jsonschema:"title=Disable Group Sync,description=Only sync users disable group and membership sync operations"`
+	// FilterExpr limits imported records to envelopes matching a CEL expression
+	FilterExpr string `json:"filterExpr,omitempty" jsonschema:"title=Filter Expression,description=Optional CEL expression to apply to records before ingesting"`
```
golanglemonade: Add a relevant example, e.g.: 
```suggestion
	FilterExpr string `json:"filterExpr,omitempty" jsonschema:"title=Filter Expression,description=Optional CEL expression to apply to records before ingesting,example=Example: payload.type == 'internal'"`
```

internal/integrations/definitions/authentik/operation_health.go
```diff
@@ -0,0 +1,51 @@
...
+// Run executes the Authentik health check
+func (HealthCheck) Run(ctx context.Context, c *authentikSDK.APIClient) (json.RawMessage, error) {
+	me, resp, err := c.CoreApi.CoreUsersMeRetrieve(ctx).Execute()
+	if resp != nil {
+		_ = resp.Body.Close()
+	}
+
+	if err != nil {
+		return nil, ErrHealthCheckFailed
+	}
```
golanglemonade, line 40: Same here, very helpful for debugging
```suggestion
	if err != nil {
	     logx.FromContext(ctx).Error().Err(err).Msg("error checking health")
	     
		return nil, ErrHealthCheckFailed
	}
```

internal/integrations/definitions/authentik/operation_directory_sync.go
```diff
@@ -0,0 +1,184 @@
...
+			PageSize(directoryDefaultPageSize).
+			IncludeUsers(true).
+			Execute()
+		if resp != nil {
+			_ = resp.Body.Close()
+		}
+
+		if err != nil {
+			return nil, ErrDirectoryGroupsFetchFailed
+		}
```
golanglemonade, line 177:
```suggestion
		if err != nil {
		    logx.FromContext(ctx).Error().Err(err).Msg("error listing groups")
		    
			return nil, ErrDirectoryGroupsFetchFailed
		}
```

internal/integrations/definitions/authentik/operation_directory_sync.go
```diff
@@ -0,0 +1,184 @@
...
+
+		result, resp, err := req.Execute()
+		if resp != nil {
+			_ = resp.Body.Close()
+		}
+
+		if err != nil {
+			return nil, ErrDirectoryUsersFetchFailed
+		}
```
golanglemonade, line 140:
```suggestion
		if err != nil {
		    logx.FromContext(ctx).Error().Err(err).Msg("error listing users")
		    
			return nil, ErrDirectoryUsersFetchFailed
		}
```

golanglemonade (commented):
nice work, @manab-pr. this is great. 

I tested locally and all seems to work:

image
image
image

Just a few final comments. Once we get docs in we can PR this again to set to active + visible. 🚀

docker/docker-compose-authentik.yml
manab-pr: yes, we can do that actually , i have modified the yml file , and it is now using main stack postgres container and added few explicit params for redis as well .

golanglemonade (commented):
all good, just need to fix the test

golanglemonade (approved):

PR #2354, opened 2026-05-05
Email Integration, Runtime Providers, Campaigns E2E

matoszz:
- Rebuilt all transactional and branded email sending as a first-class integration definition, replacing scattered ent hook email logic with a unified runtime-dispatched email operation; collapses / entirely removes both branded-emails + emailtemplates repos and libraries
- Introduced callout sections, coloring configuration, and many other useful knobs to the underlying email build and definition constructs  
- Added a template catalog and variable system that exposes available template variables to the GraphQL API for UI tooling
- Added a campaign dispatch pipeline that composes branded emails with template interpolation, recipient targeting, and recurring schedule support                                                                                                                                       
- Extracted a generic scheduled listener abstraction for Gala that handles adaptive scheduling, error classification, and per-operation schedule overrides - used by both reconcile and recurring campaign cycles                                                                                                                                                                                                                  
- Added campaign test email support so operators can preview rendered emails before launch
- Ported system Slack notifications (new user, new subscriber, demo request, GitHub app install) into the Slack integration definition, eliminating the standalone slacknotify/slacktemplates packages

Additionally the workflow engine code was updated / refactored to account for the new email templating constructs, and the integration reconciliation was inverted so that we're querying for not only the integration definition but the associated templates to perform related actions. Additional work is needed to finalize the UI display associated with the workflow definition builder and associated actions but given workflows is not yet in use yet, those updates will be added in a follow on PR.


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/900

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

config/config-dev.example.yaml
```diff
@@ -170,26 +154,59 @@ objectstorage:
...
+      clientid: local-core-oidc
+      clientsecret: local-core-oidc-secret
+      discoveryurl: http://localhost:5557/dex
+      enabled: false
+      redirecturl: http://localhost:17608/v1/integrations/auth/callback
+    email:
+        api_key: ""
+        company_address: 5150 Broadway St San Antonio, TX 78209
+        company_name: Openlane
+        corporation: theopenlane, Inc.
```
golanglemonade: These look like they got removed? 
```suggestion
```

go.mod
```diff
@@ -55,7 +55,7 @@ require (
 	github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674
 	github.com/gqlgo/gqlgenc v0.35.1
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/invopop/jsonschema v0.14.0
+	github.com/invopop/jsonschema v0.13.0
```
golanglemonade: Any reason for the downgrade or just merge conflicts missed?

internal/ent/hooks/email_context_helpers.go
```diff
@@ -0,0 +1,33 @@
...
+func organizationDisplayNameByID(ctx context.Context, client *generated.Client, orgID string) (string, error) {
+	org, err := client.Organization.Query().
+		Where(organization.ID(orgID)).
+		Select(organization.FieldDisplayName).
+		Only(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return org.DisplayName, nil
```
golanglemonade: nit, doesn't really matter but when you just select one field you can do:
```suggestion
	orgDispalyName, err := client.Organization.Query().
		Where(organization.ID(orgID)).
		Select(organization.FieldDisplayName).
		String(ctx)
	if err != nil {
		return "", err
	}

	return orgDispalyName, nil
```

internal/ent/hooks/invite.go
```diff
@@ -71,12 +69,66 @@ func HookInvite() ent.Hook {
...
+			if err != nil {
+				return nil, err
+			}
+
+			recipientRaw, recipientOK := m.Recipient()
+			emailAddress, err := requiredMutationString("recipient", recipientRaw, recipientOK)
+			if err != nil {
+				return nil, err
+			}
+
```
golanglemonade: these are already set + validated by `setRecipientAndToken` and by the schema itself

internal/ent/hooks/invite.go
```diff
@@ -71,12 +69,66 @@ func HookInvite() ent.Hook {
...
+			case auth.UserSubjectType:
+				requestor, reqErr := m.Client().User.Query().
+					Where(user.ID(reqID)).
+					Select(user.FieldFirstName).
+					Only(ctx)
+				if reqErr != nil {
+					return retValue, reqErr
+				}
+
+				inviterName = requestor.FirstName
```
golanglemonade, line 98: FirstName is not a required field, it should fall back to either DisplayName which is always auto-set or organization display name to prevent empty values: I realize you didnt change this logic though, so old bug :)

internal/ent/hooks/invite.go
matoszz: sure, but the variables aren't accessible directly inline in the function without these accesors (e.g. i can't just use m.RequestorID(), so this is extracting the values set in the mutation object)

internal/ent/hooks/listeners_campaign_recurring.go
```diff
@@ -0,0 +1,133 @@
...
+		return clearNextRunAt(ctx.Context, client, camp.ID)
+	default:
+		return nil
+	}
+}
+
+// recomputeNextRunAt computes next_run_at from now and persists it
+func recomputeNextRunAt(ctx context.Context, client *entgen.Client, camp *entgen.Campaign) error {
+	now := time.Now()
+	nextRun := operations.NextCampaignRunAt(now, camp.RecurrenceFrequency, camp.RecurrenceInterval, camp.RecurrenceTimezone)
```
golanglemonade: should this be based on last run instead of now? I would expect if my last run was 5/1 and my frequency was monthly, and I change it to every 2 months, it would run on 7/1, not 7/5?

internal/ent/hooks/invite.go
matoszz: the query above is selecting Only, no fallback is required; it's also a user query not an organization query so displayname isn't relevant to use

internal/ent/hooks/notificationtemplate.go
```diff
@@ -17,27 +21,140 @@ import (
...
+			if len(vars) == 0 && !defaultsSet {
+				return next.Mutate(ctx, m)
+			}
+
+			var jsonconfig map[string]any
+
+			if v, exists := m.Jsonconfig(); exists {
+				jsonconfig = v
+			} else if !m.Op().Is(ent.OpCreate) {
+				if old, err := m.OldJsonconfig(ctx); err == nil {
```
golanglemonade, line 84: `OldJsonconfig` will fail on `ent.OpUpdate` but that is a operation the hook supports, This also silently ignores any errors so you would have an empty jsonconfig in some cases and possibly wipe the jsonsconfig

internal/ent/hooks/notificationtemplate.go
```diff
@@ -17,27 +21,140 @@ import (
...
+const expectedRegexpGroups = 2
+
+// templateVarPattern matches dot-prefixed Go template variable references, e.g. {{ .Foo }} or {{ .User.Name }}.
+var templateVarPattern = regexp.MustCompile(`{{\s*\.([A-Za-z][A-Za-z0-9_.]*)\s*}}`)
+
+// extractTemplateVarNames scans the provided template strings for {{ .VarName }} references
+// and returns a map of top-level variable name to JSON Schema type. Dotted paths like
+// {{ .User.Name }} yield "User" with type "object". Direct references like {{ .Name }}
+// yield type "string"
+func extractTemplateVarNames(templates ...string) map[string]string {
```
golanglemonade, line 114: can you add a unit test for this, regex is always bug-prone in my experience

internal/ent/hooks/listeners_campaign_recurring.go
matoszz: I suppose it *could* be but it's a lot more convoluted than just changing that comparison, because the calculation for next run based on last run could be a time in the past rather than in the future, and while we could run it immediately given that's the case the reconcile scheduler isn't checking for next run dates in the past but not yet executed.

internal/ent/hooks/onboarding.go
```diff
@@ -59,6 +62,28 @@ func HookOnboarding() ent.Hook {
...
+			if err := sendSystemSlack(ctx, m.Client(), slackdef.DemoRequestOp.Name(), slackdef.DemoRequestMessage{
+				CompanyName:    companyName,
+				Email:          callerEmail,
+				Domains:        domains,
+				CompanyDetails: companyDetails,
+				UserDetails:    userDetails,
+				Compliance:     compliance,
+				DemoRequested:  demoRequested,
+			}); err != nil {
+				return nil, err
```
golanglemonade: We should log the error instead of returning the error if this fails; otherwise we could fail user sign-up due to failure to send a slack message when that shouldn't be in the critical path

internal/ent/hooks/organizationsetting.go
```diff
@@ -222,37 +218,24 @@ func HookBillingEmailChange() ent.Hook {
...
+			ChangedAt:       changedAt,
+		}); err != nil {
+			logx.FromContext(ctx).Error().Err(err).Msg("failed to send billing email change notification")
 
-		if _, err = m.Job.Insert(ctx, jobs.EmailArgs{
-			Message: *email,
-		}, nil); err != nil {
-			logx.FromContext(ctx).Error().Err(err).Msg("failed to queue billing email change notification")
-			continue
+			return err
```
golanglemonade: This is a change in behavior where now we will roll back the change if `sendSystemEmail` errors, whereas before the email would still be updated. That seems ok to me, but just want to make sure that is intentional?

internal/ent/hooks/subscriber.go
```diff
@@ -63,7 +63,40 @@ func HookSubscriberCreate() ent.Hook {
...
+			emailRaw, emailOK := m.Email()
+			emailAddress, err := requiredMutationString("email", emailRaw, emailOK)
+			if err != nil {
+				return nil, err
+			}
+
+			orgIDValue, ownerOK := m.OwnerID()
+			orgID, err := requiredMutationString("owner_id", orgIDValue, ownerOK)
+			if err != nil {
+				return nil, err
```
golanglemonade: Same here as in the invites; these are already require values and this is after the mutation is complete so we shouldn't need to check again.

matoszz: they aren't being checked, the variables are being extracted to be re-used

internal/ent/hooks/usersettings.go
```diff
@@ -139,46 +140,29 @@ func HookUserSettingEmailConfirmation() ent.Hook {
...
+			}); err != nil {
 				logx.FromContext(ctx).Error().Err(err).Msg("could not send welcome email")
+
+				return nil, err
+			}
+
+			if err := sendSystemSlack(ctx, m.Client(), slackdef.NewUserOp.Name(), slackdef.NewUserMessage{
+				Email: user.Email,
+			}); err != nil {
+				return nil, err
```
golanglemonade: I think we should log instead of returning the error on the slack messages; this feels more like a nice to have vs. in the user creation path

internal/ent/schema/campaign.go
```diff
@@ -218,30 +218,33 @@ func (Campaign) Fields() []ent.Field {
...
-			Optional(),
 		field.String("email_template_id").
 			Comment("the email template associated with the campaign").
 			Optional(),
+		field.String("integration_id").
+			Comment("the email integration used for campaign dispatch").
+			Optional(),
+		field.String("email_branding_id").
+			Comment("the email branding associated with the campaign").
+			Optional(),
```
golanglemonade, line 247: This schema + edge is removed, should this field be removed too?
```suggestion
```

matoszz: Im not doing any more surgery in this branch, removing the other fields was hard enough, there's a ton more that needs to be deleted that I can do at another time

internal/graphapi/campaignextended.resolvers.go
```diff
@@ -65,6 +67,20 @@ func (r *mutationResolver) LaunchCampaign(ctx context.Context, input model.Launc
 		action = campaignDispatchActionResend
 	}
 
+	caller, ok := auth.CallerFromContext(ctx)
+	if !ok {
+		logx.FromContext(ctx).Error().Msg("no caller in context")
+
+		return nil, errors.New("internal error")
```
golanglemonade: should return a defined error, e.g. 
```suggestion
		return nil, rout.ErrPermissionDenied
```

internal/integrations/definitions/email/operation_send_campaign.go
```diff
@@ -0,0 +1,231 @@
...
+				return sendCampaignIndividual(ctx, db, client, messages, targetIDs, nil)
+			}
+
+			logx.FromContext(ctx).Error().Err(err).Msg("batch send failed")
+
+			return sent, len(messages) - sent
+		}
+
+		for _, id := range batchIDs {
+			if err := markCampaignTargetSent(ctx, db, id); err != nil {
```
golanglemonade, line 181: Instead of looping, can this function just take all ids and do a bulk update so it can update 1 -> many at a time

internal/integrations/definitions/email/types.go
```diff
@@ -0,0 +1,198 @@
...
+	// ProductURL is the product home URL
+	ProductURL string `json:"productURL" koanf:"productURL" jsonschema:"description=Product home URL" default:"https://console.theopenlane.io"`
+	// DocsURL is the documentation URL
+	DocsURL string `json:"docsURL" koanf:"docsURL" jsonschema:"description=Documentation URL" default:"https://docs.theopenlane.io"`
+	// CompanyName is the display name of the sending company
+	CompanyName string `json:"companyName" jsonschema:"description=Company display name" default:"Openlane"`
+	// CompanyAddress is the mailing address of the company
+	CompanyAddress string `json:"companyAddress" jsonschema:"description=Company mailing address" default:""`
+	// Corporation is the legal corporation name
+	Corporation string `json:"corporation" jsonschema:"description=Legal corporation name" default:"theopenlane, Inc."`
```
golanglemonade, line 64: These don't have the koanf tag, which explains why they weren't in the config anymore; is that intentional?

internal/integrations/definitions/slack/builder.go
```diff
@@ -11,34 +12,19 @@ import (
 	"github.com/theopenlane/core/pkg/jsonx"
 )
 
-var chatScopes = []string{
-	"channels:read",
-	"chat:write",
-	"chat:write.public",
-	"chat:write.customize",
-}
-var directoryScopes = []string{
```
golanglemonade, line 20: I had split these up to actually show the correct scopes required in `RequiredPermissions` per operation but somehow I missed actually putting that in the builder def 🤦‍♀️

internal/integrations/definitions/slack/builder.go
```diff
@@ -11,34 +12,19 @@ import (
...
-				ID:          definitionID.ID(),
-				Family:      "Slack",
+				ID:          DefinitionID.ID(),
+				Family:      "slack",
 				DisplayName: "Slack",
 				Description: "Integrate with Slack to verify workspace posture and send operational or compliance notifications.",
 				Category:    "collaboration",
 				DocsURL:     "https://docs.theopenlane.io/docs/platform/integrations/slack/overview",
-				Tags:        []string{"messaging", "directory"},
+				Tags:        []string{"messaging", "directory-sync"},
```
golanglemonade: These are all without `-sync` on the types, should be consistent and either leave this off here or update all of the tags

internal/integrations/definitions/slack/builder.go
```diff
@@ -11,34 +12,19 @@ import (
...
-func Builder(cfg Config) registry.Builder {
+// Builder returns the Slack definition builder with the supplied operator and runtime config applied
+// When runtime.Provisioned() is true, a RuntimeIntegration is included for system-send
+func Builder(cfg Config, runtime *RuntimeSlackConfig) registry.Builder {
 	return registry.Builder(func() (types.Definition, error) {
-		return types.Definition{
+		def := types.Definition{
 			DefinitionSpec: types.DefinitionSpec{
-				ID:          definitionID.ID(),
-				Family:      "Slack",
```
golanglemonade, line 36: Family was updated to be captial on all integrations because it is what is used to create the `Vendor` record on when its setup. Any reason you switched this?

internal/integrations/definitions/slack/system_messages.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// NewUserMessage is the input for the new user Slack notification
+type NewUserMessage struct {
+	// Email is the new user email address
+	Email string `json:"email" jsonschema:"required,description=New user email address"`
+}
+
+// GitHubAppInstalledMessage is the input for the GitHub App installation Slack notification
+type GitHubAppInstalledMessage struct {
```
golanglemonade, line 40: Why only github and not any integratoin install?

internal/integrations/definitions/email/types.go
matoszz: yes; having them present flooded the configuration for attributes we effectively never change, and each individual email can override any of the runtime configuration set attributes so it didn't seem terribly relevant to have them all propagated through

internal/integrations/definitions/slack/builder.go
```diff
@@ -133,33 +119,33 @@ func Builder(cfg Config) registry.Builder {
...
 					Ref:            slackClient.ID(),
 					CredentialRefs: []types.CredentialSlotID{slackCredential.ID(), slackBotTokenCredential.ID()},
-					Description:    "Slack Web API client",
+					Description:    "Unified Slack client wrapping the Web API and system-notification transports",
 					Build:          Client{}.Build,
 				},
 			},
-			Operations: []types.OperationRegistration{
-				{
+			Operations: append(AllSlackSystemMessages(),
```
golanglemonade, line 126: These all show up in the user-slack setup when they should be hidden as they aren't user-facing operations, 
image

golanglemonade: I dont feel like scrolling up...so same with email :-D 
image

internal/integrations/runtime/integration.go
```diff
@@ -49,6 +49,36 @@ func (r *Runtime) ResolveIntegration(ctx context.Context, lookup IntegrationLook
...
+			if prefer[0](inst) {
+				return inst.ID, nil
+			}
+		}
+	}
+
+	if len(integrations) == 1 {
+		return integrations[0].ID, nil
+	}
+
```
golanglemonade: This  should be switched, because if there is only 1, it will always return that not matter the prefer? What happens in the case of more than one but not prefer? This will currently return no id. 
```suggestion
	if len(integrations) == 1 {
		return integrations[0].ID, nil
	}
	
	if len(prefer) > 0 {
		for _, inst := range integrations {
			if prefer[0](inst) {
				return inst.ID, nil
			}
		}
	}

```

internal/ent/hooks/subscriber.go
golanglemonade: Right, I just mean they can be like this, the requiredMutationString is checking if they are set, but thats already done. 

```
			emailRaw, _ := m.Email()

```

golanglemonade (approved):

PR #2365, opened 2026-05-06
fix: return early for nil creds schema

golanglemonade:
- returns early for nil creds schema (e.g. Google Workspace because it is done via oauth)
- Fixes log lines for CF
- adds integration family name to logs to help diagnose issues in the future
- adds tag for Campaign events to help with UI filtering later


golanglemonade:
pulling into another PR, closing this one out

PR #2367, opened 2026-05-06
Centralize integration log context and simplify durable codec

matoszz:
- Centralized integration identity and operation metadata onto the context logger so every downstream log line automatically includes consistent fields without manual repetition
- Simplified the gala durable context codec by removing a legacy composite codec that duplicated auth propagation already handled by the generic key codec
- Added nil and empty schema guards in credential and config resolution to prevent panics for OAuth-based integrations that have no credential schema 
- Fixed the operation client resolver to return resolved credentials to the caller and corrected the email template input to match its updated schema


internal/integrations/definitions/cloudflare/operation_health.go
```diff
@@ -32,15 +32,15 @@ func (h HealthCheck) Handle() types.OperationHandler {
...
 		return nil, err
 	}
 
 	res, err := c.Accounts.Tokens.Verify(ctx, accounts.TokenVerifyParams{
 		AccountID: cf.F(meta.AccountID),
 	})
 	if err != nil {
-		logx.FromContext(ctx).Error().Err(err).Msg("githubapp: healthcheck failed on token verification")
+		logx.FromContext(ctx).Error().Err(err).Msg("healthcheck failed on token verification")
 		return nil, ErrTokenVerificationFailed
```
golanglemonade, line 44: Can you add `cloudflare: ` to these logs, I've been using them to easily figure out which integration that is for (these were just labeled wrong 🙃

matoszz: its already embedded in the context, kind of the point of the construct :)

golanglemonade (approved):

PR #2377, opened 2026-05-07
Enrich logo of entities from system owned subprocessors

adelowo:


internal/ent/hooks/entity.go
```diff
@@ -128,7 +130,87 @@ func HookEntityCreate() ent.Hook {
...
+	if len(names) == 0 {
+		return nil
+	}
+
+	defaultPredicates := make([]predicate.Subprocessor, 0, len(names))
+	for _, name := range names {
+		defaultPredicates = append(defaultPredicates, subprocessor.NameEqualFold(name))
+	}
+
+	processors, err := m.Client().Subprocessor.Query().
```
matoszz, line 153: you'd presumably need to do a privacy bypass here?

internal/ent/hooks/entity.go
```diff
@@ -128,7 +130,87 @@ func HookEntityCreate() ent.Hook {
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	if description, ok := m.Description(); !ok || strings.TrimSpace(description) == "" {
+		if strings.TrimSpace(processors.Description) != "" {
+			m.SetDescription(processors.Description)
+		}
```
matoszz, line 175: why would we ever want to trim spaces on a description?

internal/ent/hooks/entity.go
```diff
@@ -128,7 +130,87 @@ func HookEntityCreate() ent.Hook {
...
+	}
+
+	return nil
+}
+
+func getNameMatches(m *generated.EntityMutation) []string {
+	var names []string
+
+	fn := func(name string) {
+		name = strings.TrimSpace(name)
```
matoszz, line 193: should not be trimming spaces on a name that could be 2 words

internal/graphapi/entity_test.go
```diff
@@ -315,6 +316,51 @@ func TestMutationCreateEntity(t *testing.T) {
...
+
+	resp, err := suite.client.api.CreateEntity(testUser1.UserCtx, testclient.CreateEntityInput{
+		Name: lo.ToPtr(strings.ToUpper(name)),
+		Tier: lo.ToPtr(enums.VendorTierStandard),
+	}, nil, nil, nil, nil, nil)
+
+	assert.NilError(t, err)
+	assert.Assert(t, resp != nil)
+
+	assert.Check(t, is.Equal(description, *resp.CreateEntity.Entity.Description))
```
matoszz, line 341: confused how this test would pass given the description definition above has spaces in it and the hook would have trimmed them out in the created entity

internal/ent/hooks/entity.go
golanglemonade: They are system-owned subprocessor, not privacy bypass needed, user would have access to these. They work the same way with trust-center subprocessors being able to be created based on these

matoszz (approved):

PR #2396, opened 2026-05-13
Dev mode for email/slack integration providers, Unify Webhook Handlers, Ported NDA attestation 

matoszz:
- Unified the webhook infrastructure so that email (Resend), GitHub App, and campaign webhooks all flow through a single portable, runtime-configurable webhook pipeline instead of each having its own bespoke handler and route
- Added a dev mode for both email and Slack integrations so the platform can start and process dispatches without live provider credentials, removing a hard dependency on third-party services during local development
- Built out the NDA attestation workflow end-to-end (ported from corejobs): when a trust center NDA is signed, the system now downloads the original PDF, appends a signature certification page with signer metadata, uploads the attested copy to object storage, and records the hash for tamper detection
- Ported object storage operations for NDA uploads to use the native storage client directly instead of going through presigned URLs; switched PDF processing entirely to pdfcpu, dropping the previously used deprecated library
- Made the Slack notification for integration installs generic rather than GitHub-specific, so any new integration type gets a Slack notification without custom plumbing
- Added an email webhook handler within the email integration definition so Resend delivery events (sent, delivered, opened, clicked, bounced) are tracked through the same integration pipeline as everything else
- Simplified the integration registry by removing concurrency locks that were unnecessary given the registry is built once at startup and read-only thereafter
- Added a payment reminder operation to the scheduled jobs framework
- Added an org deletion notice email template


internal/integrations/definitions/email/types.go
```diff
@@ -40,10 +42,14 @@ const (
 // at startup. Branding and presentation fields carry struct-tag defaults only and
 // are overridable per-send via UserInput or per-operation Config functions
 type RuntimeEmailConfig struct {
+	// TestDir is the directory where dev-mode email files are written
+	TestDir string `json:"testDir,omitempty" koanf:"testDir" jsonschema:"description=Directory for dev-mode email output" default:"fixtures/email"`
+	// ResendSecret is the Resend webhook signing secret for verifying inbound delivery events
+	ResendSecret string `json:"resendSecret,omitempty" koanf:"resendSecret" jsonschema:"description=Resend webhook signing secret" sensitive:"true"`
```
golanglemonade: Must have missed this on a previous PR, this whole section has incorrect casing for koanf configs
```suggestion
	ResendSecret string `json:"resendsecret,omitempty" koanf:"resendsecret" jsonschema:"description=Resend webhook signing secret" sensitive:"true"`
```

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/926

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

internal/integrations/definitions/email/compose.go
```diff
@@ -33,19 +33,21 @@ type CampaignDispatchResult struct {
 // loadEmailTemplate resolves an active email template by ID for the given owner, eager-loading
 // the Files edge so static attachments can be included in the dispatched message
 func loadEmailTemplate(ctx context.Context, client *generated.Client, ownerID string, emailTemplateID string) (*generated.EmailTemplate, error) {
+	systemCtx := privacy.DecisionContext(ctx, privacy.Allow)
```
golanglemonade, line 36: Why is this now needed but wasn't needed before?

internal/integrations/operations/payment_reminder.go
```diff
@@ -0,0 +1,69 @@
...
+// PaymentReminderConfig contains the configuration for the payment reminder scheduled listener
+type PaymentReminderConfig struct {
+	// PaymentMethodInterval is the number of days after org creation before
+	// an org without a payment method is marked for deletion
+	PaymentMethodInterval uint8 `json:"payment_method_interval" koanf:"paymentmethodinterval" jsonschema:"default=30,description=Days after org creation before marking for deletion"`
+	// DeletionDays is the number of days between marking an org for deletion
+	// and the actual deletion date set on pending_deletion_at
+	DeletionDays uint8 `json:"deletion_days" koanf:"deletiondays" jsonschema:"default=7,description=Days between marking and actual deletion"`
+	// Enabled controls whether the payment reminder polling loop is seeded at startup
+	Enabled bool `json:"enabled" koanf:"enabled" jsonschema:"description=Whether the payment reminder listener is enabled"`
```
golanglemonade, line 33: This is missing `DryRun` mode, where this is setup to just log the orgs that would be marked for deletion instead of sending the emails:

https://github.com/theopenlane/riverboat/blob/main/pkg/jobs/organization_payment_reminder.go#L156-L162

internal/ent/hooks/nda_attestation.go
```diff
@@ -0,0 +1,560 @@
...
+	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
+
+	templateFile, err := fetchNDATemplateFile(allowCtx, client, templateID)
+	if err != nil {
+		return nil, err
+	}
+
+	var ndaMetadata signedNDADocumentData
+	if err := jsonx.RoundTrip(docData.Data, &ndaMetadata); err != nil {
+		return nil, ErrFailedToUnmarshalNDAMetadata
```
golanglemonade, line 84: Can these errors that are returned as more generic errors be logged so we know the actual error that occurred; this helps a lot with debugging; theres a few others in this file too that should also log the error. 
```suggestion
        logx.FromContext(ctx).Error().Err(err).Msg("failed to unmarshal nda metadata")
         
		return nil, ErrFailedToUnmarshalNDAMetadata
```

internal/integrations/operations/payment_reminder.go
```diff
@@ -0,0 +1,69 @@
...
+	PaymentReminderMinInterval = 6 * time.Hour
+	// PaymentReminderMaxInterval is the maximum polling interval for payment reminders
+	PaymentReminderMaxInterval = 24 * time.Hour
+)
+
+// PaymentReminderConfig contains the configuration for the payment reminder scheduled listener
+type PaymentReminderConfig struct {
+	// PaymentMethodInterval is the number of days after org creation before
+	// an org without a payment method is marked for deletion
+	PaymentMethodInterval uint8 `json:"payment_method_interval" koanf:"paymentmethodinterval" jsonschema:"default=30,description=Days after org creation before marking for deletion"`
```
golanglemonade:
```suggestion
	PaymentMethodInterval uint8 `json:"paymentmethodinterval" koanf:"paymentmethodinterval" jsonschema:"default=30,description=Days after org creation before marking for deletion"`
```

internal/integrations/definitions/email/compose.go
matoszz: for the file attachment to work (hence the comment noting limited support being removed, the new context being used)

internal/integrations/operations/payment_reminder.go
matoszz: The PR adds a "devmode" setup which would not actually send any of the system emails, and this individual listener also has an enabled and disabled; what's the use case for having all other system emails enabled, enabling this but then not having it actually process?

golanglemonade: dry run was also to not actually mark the records as pending deletion so we could turn it on and make sure its functioning correctly and not accidentally start incorrectly deleting full organizations

golanglemonade (approved):

PR #2398, opened 2026-05-14
feat: permissions revamp

golanglemonade:
⚠️  **Breaking Changes**

This PR contains revamped changes to permissions constructs and requires data migration.

- FGA Model - this switches to using a `fga.mod` file in order to break up permissions into more manageable files. These are grouped based on functionality (e.g. `compliance`, `trust center`, `regsitry`, etc.
- All base `crud` permissions are now generated based on the ent schema and are located in `fga/model/generated/crud.fga`. This contains all the `can_<action>_<object>` permissions that are used for both API Token scopes as well as base sets of permissions
- New `roles.fga` file that contains more defined role types, e.g. `compliance_manager`, `policy_manager`, `risk_manager` - this will allow for easier designation of roles to user. These can be added on top of organization roles. For example, a `member` can be given `policy_manger` role which will give them full access to manage all policies in the organization. 
    - This file uses special annotations to allow for easier management without duplication:
        - `@inherit`: This inherits all the same permissions as the specified roles, e.g. `compliance_manager` inherits the permissions given to `can_manage_compliance, can_manage_policies, can_manage_registry, can_manage_risk`
        - `@crud`: This specifies which objects the role provides crud access to
- New Roles:
    - `Super Admin`: Users with this role will have the same permissions as the owner with the exception of being able to delete the organization. Organizations can have 0->many super admins
    - `Auditor`: This will be a mostly read-only role users can give to their auditor and in the future be able to specifiy what they have access to
 - Scoped API Tokens - The update will allow API Tokens to be scoped to specific objects + operations allowing for more fined grained access on the tokens. This deprecates the old `can_view`, `can_edit`, `can_delete`, `group_manager` scopes in favor of new scopes. **THE UI NEEDS TO BE UPDATED TO ACCOMODATE THIS CHANGE**
  - Adds new `parent_context` relation which replaces the `parent` relation for `organization` level checks for all objects with the exception of `files`.  This requires a migration of `parent` relations for `organization` objects to `parent_context`. We cannot use contextual tuples because of parent hierarchy, e.g. this fails when you get to something like `control_implementation` where the access is based on the control and the control is based on the organization - we'd have to pass the full hierachy for this to work with the structure we have today. 
  - Removes adding user tuples when objects are created; previously this meant if you changed roles your access wasn't removed for any objects you created. Now, it is based on the roles - so if you can manage policies, you can see the policy you created and do not need a user specific tuple. 

Other Changes:
- Logs `email` when `checkAllowedEmailDomain` fails so we know the user attempted to be added to the organization
- Changes the `entitlement reconciliation failed` logs when a constraint error to an warn log so test logs aren't full of these errors but we still see the warning in production. All other errors are kept as error level logs.
- Updates default org update logic to order by personal org desc so we don't set the default org to the personal org if there are other orgs available to the user
- Adds helpers for creating a fresh org, cleaning up org after
- Moves all tests that create fresh orgs and do not do river checks (most of the trust center tests) to use `t.Parallel()`

Deps:
- [Migration Guide](https://github.com/theopenlane/core/blob/5dcbde602274126f10faea65fd054bf7734244b9/db/MIGRATION.md)
Depends on: https://github.com/theopenlane/iam/pull/508
Depends on: https://github.com/theopenlane/entx/pull/278


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/928

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

internal/ent/hooks/apitoken.go
```diff
@@ -91,6 +93,24 @@ func HookCreateAPIToken() ent.Hook {
...
+			var scopesModified bool
+
+			// Only query old scopes if scopes are being modified and this is an UpdateOne operation
+			_, scopesModified = m.Scopes()
+			if !scopesModified {
+				// check appended
+				_, scopesModified = m.AppendedScopes()
+			}
+
+			if scopesModified && m.Op().Is(ent.OpUpdateOne) {
```
matoszz, line 106: the hook triggers on OpUpdate and OpUpdateOne but you only have logic for OpUpdateOne

internal/ent/hooks/objectownedtuples.go
```diff
@@ -39,7 +37,9 @@ func HookObjectOwnedTuples(parents []string, ownerRelation string, skipCreateUse
 
 			var addTuples []fgax.TupleKey
 
-			if skip := skipCreateUserPermissions(ctx, m); !skip {
```
matoszz, line 42: is something changing with the order of operations that allows us to safely remove this?

internal/ent/hooks/organization.go
```diff
@@ -500,6 +501,13 @@ func checkAndUpdateDefaultOrg(ctx context.Context, userID string, oldOrgID strin
 		newDefaultOrgID, err := client.
 			Organization.
 			Query().
+			Where(
+				organization.IDNEQ(oldOrgID),
+			).
+			Order(
+				// order by personal orgs last so that if there is another org available it will be set as the default instead of the personal org
+				organization.ByPersonalOrg(sql.OrderAsc()),
```
matoszz, line 486: why would we need to perform ordering when the Where condition already precludes the previously set organization ID ?

internal/ent/privacy/rule/group.go
```diff
@@ -2,64 +2,83 @@ package rule
...
+	if err := CheckSubjectScope(ctx, m.Type(), "", &op); err != nil {
+		if !errors.Is(err, privacy.Skip) {
+			return err
 		}
+	}
 
-		caller, ok := auth.CallerFromContext(ctx)
-		if !ok || caller == nil || caller.IsAnonymous() {
-			logx.FromContext(ctx).Info().Msg("unable to get caller from context")
+	caller, ok := auth.CallerFromContext(ctx)
```
matoszz, line 41: it might be easier to do `GetAuthzSubjectType` or one of the other helpers here

internal/ent/privacy/rule/invite.go
```diff
@@ -100,11 +102,17 @@ func getRelationToCheck(ctx context.Context, m *generated.InviteMutation) (strin
 		}
 	}
 
-	// allow member to invite members
-	if strings.EqualFold(role.String(), fgax.MemberRelation) {
+	switch strings.ToLower(role.String()) {
+	case strings.ToLower(fgax.MemberRelation):
```
matoszz: the fgax constants are all already lower, no real reason to have strings.ToLower prefixing all these

internal/ent/privacy/rule/invite.go
```diff
@@ -100,11 +102,17 @@ func getRelationToCheck(ctx context.Context, m *generated.InviteMutation) (strin
...
+	case strings.ToLower(fgax.SuperAdminRelation):
+		return inviteSuperAdminRelation, nil
+	case strings.ToLower(fgax.AuditorRelation):
+		return inviteAuditors, nil
 	}
 
-	// default to admin
-	return inviteAdminRelation, nil
+	// default to super admin
+	return inviteSuperAdminRelation, nil
```
matoszz: why not use `default` in the switch / case statement?

internal/ent/privacy/rule/organization.go
```diff
@@ -33,9 +33,20 @@ func CheckCurrentOrgAccess(ctx context.Context, m ent.Mutation, relation string)
 		return privacy.Allow
 	}
 
+	// Check API token scope first if applicable
+	genericMut, ok := m.(utils.GenericMutation)
+	if ok {
+		op := genericMut.Op()
+		if err := CheckSubjectScope(ctx, genericMut.Type(), relation, &op); err != nil {
```
matoszz, line 41: what's the rationale behind filtering out the return here? It seems slightly strange to have `if !errors.Is(err, privacy.Skip` (since its either privacy skip or privacy allow) since all the callers of CheckCurrentOrgAccess are similarly performing an errors.Is check for privacy allow?

internal/ent/privacy/rule/organization.go
```diff
@@ -33,9 +33,20 @@ func CheckCurrentOrgAccess(ctx context.Context, m ent.Mutation, relation string)
...
+			if !errors.Is(err, privacy.Skip) {
+				return err
+			}
+		}
+	}
+
 	caller, ok := auth.CallerFromContext(ctx)
 	if ok && caller != nil && caller.OrganizationID != "" {
-		if relation == fgax.CanView {
+		if relation == fgax.CanView && !auth.IsAPITokenAuthentication(ctx) {
```
matoszz, line 50: why filter the check to exclude API token auth?

internal/ent/privacy/rule/orgowner.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/theopenlane/core/internal/ent/generated/trustcenter"
+	"github.com/theopenlane/core/internal/ent/privacy/utils"
+	"github.com/theopenlane/core/pkg/logx"
+	"github.com/theopenlane/iam/auth"
+	"github.com/theopenlane/iam/fgax"
+)
+
+// skipperType are types that are generally user owned where the owner_id is a user
+// not an organization
+var skipperType = map[string]struct{}{
```
matoszz: why not use the same `Caller` and `Key` constructs for these vs. stringly typed map?

internal/ent/privacy/rule/orgowner.go
```diff
@@ -0,0 +1,286 @@
...
+	trustCenterID := getTrustCenterIDFromMutation(ctx, m)
+	if trustCenterID == "" {
+		return privacy.Skip
+	}
+
+	// the function above already checks this is a trust
+	// center mutation, no need to check again
+	tcMutation, _ := m.(utils.TrustCenterMutation)
+
+	ownerID, err := tcMutation.Client().TrustCenter.Query().Where(
```
matoszz, line 284: don't you need a privacy.Allow for this to work ?

internal/ent/privacy/rule/orgowner.go
```diff
@@ -0,0 +1,286 @@
...
+
+// checkOrganizationMutation checks to see the user has access to the organization mutation
+// based on mutation type and parent organization
+func checkOrganizationMutation(ctx context.Context, m ent.Mutation, orgID string) error {
+	mut := m.(*generated.OrganizationMutation)
+
+	if m.Op().Is(ent.OpCreate) {
+		parentOrgID, ok := mut.ParentOrganizationID()
+		if !ok || parentOrgID == "" {
+			// if there is no parent org, allow
```
matoszz: skip != allow; might want to change the comment or confirm the return shouldn't instead be privacy.Allow?

internal/ent/privacy/rule/orgowner.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	// check view access to the organization instead if the object is an organization
+	if strings.EqualFold(objectType, organization.Label) {
+		if objectID != "" && orgID != objectID {
+			return privacy.Denyf("user does not have access to the requested organization")
+		}
+
+		if err := CheckCurrentOrgAccess(ctx, m, fgax.CanView); errors.Is(err, privacy.Allow) {
+			return nil
```
matoszz: for consistency i think this should return skip instead of nil

internal/ent/privacy/rule/orgowner.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+
+		return privacy.Denyf("user does not have access to the requested organization")
+	}
+
+	if strings.EqualFold(objectType, orgmembership.Label) {
+		if err := CheckCurrentOrgAccess(ctx, m, fgax.CanView); errors.Is(err, privacy.Allow) {
+			return nil
```
matoszz: same comment as above

internal/ent/privacy/rule/orgowner.go
```diff
@@ -0,0 +1,286 @@
...
+		return privacy.Denyf("user does not have access to the requested organization")
+
+	}
+
+	// check if the object is in the organization
+	pluralObjectType := pluralize.NewClient().Plural(objectType)
+	tableName := strcase.SnakeCase(pluralObjectType)
+
+	// files are not org owned, they rely on parents
+	if tableName == "files" {
```
matoszz, line 244: what about the files table is unique from every other schema we have that would necessitate it being structured this way?

internal/ent/privacy/rule/scopes.go
```diff
@@ -0,0 +1,161 @@
...
+		// strip history suffix for history tables
+		objectType = strings.TrimSuffix(objectType, "History")
+
+		op := m.Op()
+		return CheckSubjectScope(ctx, objectType, "", &op)
+	})
+}
+
+// CheckSubjectScope enforces that the authorized subject has the required scope for the given object type, relation, and operation.
+// Returns nil if the rule should be skipped (no scoped relation), privacy.Allow if access is granted, or an error if denied
```
matoszz: returns privacy.Skip not nil, right?

internal/ent/privacy/rule/scopes_test.go
```diff
@@ -0,0 +1,32 @@
...
+func TestScopedRelation(t *testing.T) {
+	tests := []struct {
+		name       string
+		objectType string
+		relation   string
+		op         ent.Op
+		expected   string
+	}{
+		{name: "view query", objectType: "Control", relation: "can_view", expected: "can_view_control"},
+		{name: "update op", objectType: "Task", relation: "can_view", op: ent.OpUpdate, expected: "can_edit_task"},
```
matoszz: update operation checks can_view but on OpUpdate and expected can edit?

internal/ent/privacy/rule/scopes_test.go
```diff
@@ -0,0 +1,32 @@
...
+		{name: "edit relation", objectType: "Evidence", relation: "can_edit", expected: "can_edit_evidence"},
+		{name: "delete op", objectType: "File", relation: "can_edit", op: ent.OpDeleteOne, expected: "can_delete_file"},
```
matoszz: shouldn't this be can_delete?

internal/ent/privacy/rule/scopes.go
```diff
@@ -0,0 +1,161 @@
...
+		logx.FromContext(ctx).Debug().Err(err).Interface("check", ac).Msg("failed scope check, unable to determine access")
+
+		return privacy.Skip
+	}
+
+	if hasAccess {
+		return privacy.Allow
+	}
+
+	if auth.IsAPITokenAuthentication(ctx) {
```
matoszz, line 153: isn't this duplicative of the check that's at the top of the function already?

internal/ent/privacy/rule/scopes.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	CanCreatePrefix = "can_create_"
+	CanEditPrefix   = "can_edit_"
+	CanDeletePrefix = "can_delete_"
+	CanViewPrefix   = "can_view_"
+)
+
+// scopedRelation returns the scoped relation based on the object type, relation, and operation. A operation is checked for for create, update, delete. If instead a specific relation should be checked, that should be passed instead of the operation
+func scopedRelation(objectType string, relation string, op *ent.Op) string {
+	object := strcase.SnakeCase(objectType)
```
matoszz, line 30: why manipulate the object type input like this?

internal/ent/privacy/rule/scopes.go
```diff
@@ -0,0 +1,161 @@
...
+	return objectType
+}
+
+// AllowIfTokenHasMutationScope is a rule that allows mutation if the api token has the appropriate scope
+// for the object type and operation
+// this is used on the base mutation policy to enforce api token scope checks
+func AllowIfTokenHasMutationScope() privacy.MutationRuleFunc {
+	return privacy.MutationRuleFunc(func(ctx context.Context, m ent.Mutation) error {
+		objectType := m.Type()
+		if objectType == "" {
```
matoszz: is it possible for m.Type() to return empty?

internal/graphapi/workflow_resolvers_extended_test.go
```diff
@@ -129,6 +129,7 @@ func createWorkflowProposal(t *testing.T, ctx context.Context, ownerID string, i
 
 func TestRequestChangesWorkflowAssignment(t *testing.T) {
 	ensureWorkflowEngine(t)
+	t.Parallel()
```
matoszz, line 132: any particular reason these parallels were added?

matoszz (dismissed):

internal/ent/hooks/objectownedtuples.go
golanglemonade: We are no longer creating tuples based on who created the objects; this was actually problematic because a user that created an object would always have permissions to the object even if their role changed and didn't need it

internal/ent/hooks/organization.go
golanglemonade: so that we always return the non-personal org if there is one; we take the first result so this will make sure the personal org is last

internal/ent/privacy/rule/group.go
golanglemonade: need the organization id as well, so this gives me both

internal/ent/privacy/rule/organization.go
golanglemonade: because if we have an explicit deny or allow (or any other error) we want to return it, if that check returned a skip, we want to continue in the function

internal/ent/privacy/rule/organization.go
golanglemonade: because api tokens we need to check scopes

internal/ent/privacy/rule/orgowner.go
golanglemonade: no, this should be based on the users context, if the ctx can't get the trust center than they dont have access

internal/ent/privacy/rule/orgowner.go
golanglemonade: done

internal/ent/privacy/rule/orgowner.go
golanglemonade: done

internal/ent/privacy/rule/orgowner.go
golanglemonade: There is a comment in the fga schema, files still have the parent `organization` instead of parent_context. This is because files are much more complex with parent/child relations and can be owned by a user _or_ an organization so we can't follow the same pattern. Everything else is one or the other.

internal/ent/privacy/rule/scopes_test.go
golanglemonade: because op takes precedence, I'll update the test to be more clear

internal/ent/privacy/rule/scopes_test.go
golanglemonade: same, op takes precedence. Removing from test so its clear

internal/ent/privacy/rule/scopes.go
golanglemonade: No, the top allows only `objectType == generated.TypeAPIToken `

internal/ent/privacy/rule/scopes.go
golanglemonade: objects are ent objects so are `InternalPolicy` and fga relations are snake case, so `internal_policy`

internal/ent/privacy/rule/scopes.go
golanglemonade: shouldn't be, only if manually crafted a mutation

internal/graphapi/workflow_resolvers_extended_test.go
golanglemonade: yeah, because I had removed them before because we were getting errors on conflicting tests, but I cleaned them all up - a lot of it was trust center things - and was able to add these back.

matoszz (approved):

PR #2400, opened 2026-05-14
ISS-2196: policy word doc management

papistacoding:
Adds a `DocumentManagementMode` enum (`OPENLANE_MANAGED` |
  `EXTERNAL_REFERENCE`) on the shared document mixin so `InternalPolicy`,
  `Procedure`, and `ActionPlan` can flag how the user wants to edit them —
  inline in Openlane's Plate editor, or as an uploaded Word file they
  manage in their native editor and only preview here. Backfills existing
  rows to `OPENLANE_MANAGED`.

  `managementMode` is a **presentation hint for the frontend**, not a
  backend parsing toggle: the document import hook parses every upload
  unchanged so search, AI suggestions, and downstream automation keep
  working regardless of how the user views the document. The frontend
  reads the mode to decide between rendering the rich-text editor or the
  file preview + "Replace document" UI.

  `CreateUpload*` mutations take an optional `managementMode` so the
  frontend can persist the user's choice on creation. Resolvers expose
  `presignedURL` on the file edge so the preview can fetch the raw file.

  Hardening on the new enum surface:
  - `UnmarshalGQL` validates membership so bogus values can't slip into
    Go memory; adds `IsValid()`, `ToDocumentManagementModeOrDefault()`,
    and focused tests.
  - `parseDocx` errors are now wrapped with a new `ErrDOCXParseFailed`
    sentinel.
  - The MIME validator now relies on `normalizeMimeType` in
    `pkg/objects/validators.go` to strip `; charset=utf-8` and similar
    parameter suffixes, so the duplicate entries (`text/plain` and
    `text/plain; charset=utf-8`) are removed from the allowlists.


go.sum
```diff
@@ -440,6 +440,8 @@ github.com/labstack/echo/v4 v4.15.2 h1:nnh2sCzGCVYnU+wCisMPiYapEg/QVo/gcI9ePKg5/
 github.com/labstack/echo/v4 v4.15.2/go.mod h1:Xzp1Ns1RA2c9fY7nSgUJkpkUZGNbEIVHZbtbOMPktBI=
 github.com/labstack/gommon v0.5.0 h1:6VSQ2NOzsnEJ5W6+84E0RbcaDDmgB6NIAzWCczTEe6c=
 github.com/labstack/gommon v0.5.0/go.mod h1:Rzlg7HHy1maLfzBYGg9NZcVuz1sA68HHhLjhcEllYE0=
+github.com/ledongthuc/pdf v0.0.0-20250511090121-5959a4027728 h1:QwWKgMY28TAXaDl+ExRDqGQltzXqN/xypdKP86niVn8=
```
golanglemonade: We should use `github.com/pdfcpu/pdfcpu` for pdf processing - but also I'm not sure why this is needed. If its a pdf why not just view the actual PDF?

golanglemonade: Ah,  see the why. We didnt previously support upload via pdf - only export. I think we should leave this off for now then; the formatting here isn't great so I'd rather just now support it unless we need to later.

internal/ent/hooks/documents.go
```diff
@@ -107,6 +108,20 @@ func HookImportDocument() ent.Hook {
...
+				// into details/name/tags — that would silently overwrite the
+				// metadata the user is managing externally.
+				if doc, ok := mut.(documentMutation); ok && managementModeFor(ctx, doc) == enums.DocumentManagementModeExternalReference {
+					if err := attachUploadedFile(ctx, mut); err != nil {
+						return nil, err
+					}
+
+					break
+				}
+
```
golanglemonade: Why do you need this? The upload process should work the same as it does today

internal/ent/hooks/documents.go
```diff
@@ -124,6 +139,34 @@ func mutationToFileKey(m importSchemaMutation) string {
 	return fmt.Sprintf("%sFile", strcase.LowerCamelCase(m.Type()))
 }
 
+// attachUploadedFile points the mutation's FileID at the freshly uploaded file
+// without parsing it. This is the EXTERNAL_REFERENCE path: Openlane stores and
```
golanglemonade: We do want to parse it; that allows search to still work

internal/ent/hooks/documents.go
```diff
@@ -225,20 +268,35 @@ var client = &http.Client{
...
+
+// allowedImportSchemes restricts importURLToSchema to remote HTTP(S) origins. Schemes like
+// file://, gopher://, dict://, etc. are SSRF-adjacent and should never reach our fetcher.
+var allowedImportSchemes = map[string]struct{}{"http": {}, "https": {}}
+
 // importURLToSchema is a helper that fetches content from a URL, detects its MIME type, parses it and writes
 // the sanitized content into the mutation details, recording the URL used
-func importURLToSchema(m importSchemaMutation) error {
+//
+// TODO: This is still SSRF-exposed to private network ranges (10/8, 192.168/16, 172.16/12,
```
golanglemonade: What is all this?

internal/ent/hooks/revision.go
```diff
@@ -63,6 +64,13 @@ func HookRevisionUpdate() ent.Hook {
 // If the revision is set, it does nothing
 // If the revision is not set, it retrieves the current revision from the database and bumps the version based on the revision bump
 // If there is no revision bump set, it bumps the patch version
+//
+// TODO: known read-modify-write race — OldRevision + SetRevision is not serialized against
+// concurrent updaters on the same row, so two parallel writes can compute the same vN+1 and
+// the audit log loses one of them. The proper fix is to either (a) run this hook inside a
+// transaction with SELECT ... FOR UPDATE on the target row before reading OldRevision, or
+// (b) push the bump into SQL via "WHERE revision = $old" with retry on row-count == 0.
+// Both require infrastructure changes beyond this hook. Track in a dedicated PR.
```
golanglemonade:
```suggestion
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +130,62 @@ const (
 	detailsFieldName     = "details"
 )
 
+// documentMutation is the subset of revision-bearing mutations that carry a document file
+// and a management mode. Action plans, internal policies, and procedures satisfy it; other
+// revision-bearing mutations (standards, templates, control objectives) do not.
```
golanglemonade: this is kind of uncessary and would require updating the list of types that do or dont when that mixin was added
```suggestion
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +130,62 @@ const (
...
+// mutations behave as managed documents.
+func managementModeFor(ctx context.Context, mut documentMutation) enums.DocumentManagementMode {
+	if v, ok := mut.ManagementMode(); ok && v.IsValid() {
+		return v
+	}
+
+	if v, err := mut.OldManagementMode(ctx); err == nil && v.IsValid() {
+		return v
+	} else if err != nil {
+		logx.FromContext(ctx).Debug().Err(err).Msg("could not read old management_mode; defaulting to OPENLANE_MANAGED")
```
golanglemonade:
```suggestion
		logx.FromContext(ctx).Info().Err(err).Msg("could not read old management_mode; defaulting to OPENLANE_MANAGED")
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +130,62 @@ const (
...
+
+// managementModeFor returns the mode that will be in effect after this mutation.
+// Defaults to OPENLANE_MANAGED so pre-existing rows (NULL column) and create-path
+// mutations behave as managed documents.
+func managementModeFor(ctx context.Context, mut documentMutation) enums.DocumentManagementMode {
+	if v, ok := mut.ManagementMode(); ok && v.IsValid() {
+		return v
+	}
+
+	if v, err := mut.OldManagementMode(ctx); err == nil && v.IsValid() {
```
golanglemonade: `OldManagementMode` only works on `OpUpdateOne`, so I would add a check for that before and if its:
- OpCreate and not set, `DocumentManagementModeOpenlaneManaged` would be accurate
- OpUpdate (meaning many are updating) we won't know, but a user would never update the details in a bulk update so we would just want to default to the patch revision bump

internal/ent/hooks/revision.go
```diff
@@ -117,6 +130,62 @@ const (
...
+	cleared := mut.FileIDCleared()
+	if !set && !cleared {
+		return false
+	}
+
+	oldID, err := mut.OldFileID(ctx)
+	if err != nil {
+		logx.FromContext(ctx).Debug().Err(err).Msg("could not read old file_id; treating as file change")
+		return true
+	}
```
golanglemonade, line 164: same here; this would end up as "file changed" on a bulk update because `mut.OldFileID(ctx)` is not valid except on OpUpdateOne

golanglemonade: It's also probably worth nothing that `ProcessFilesForMutation` already processes the file; it might be better to just use that, add a context key and check that; because if we processed a file for this, we know its updated.

internal/objects/validators/mime.go
```diff
@@ -28,8 +30,8 @@ var sharedMimeTypes = []string{
 	"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
 	"application/x-vnd.oasis.opendocument.spreadsheet",
 	"text/csv",
-	"application/x-yaml", "application/x-yaml; charset=utf-8", "text/yaml",
-	"application/json", "application/json; charset=utf-8",
```
golanglemonade, line 32: Why are a bunch of these removed?

internal/objects/validators/mime.go
```diff
@@ -44,7 +46,7 @@ var validMimeTypes = map[string][]string{
 	"logoFile":           {"image/jpeg", "image/png", "image/svg+xml", "image/webp"},
 	"faviconFile":        {"image/jpeg", "image/png", "image/x-icon"},
 	"evidenceFiles":      append(sharedMimeTypes, evidenceMimeTypes...),
-	"exportFiles":        {"text/csv", "text/plain; charset=utf-8", "text/plain", "application/json", "application/json; charset=utf-8", "application/zip", "application/pdf", "application/vnd.openxmlformats-officedocument.wordprocessingml.document", "text/markdown", "text/x-markdown"},
```
golanglemonade, line 47: Same here  - why removed?

internal/objects/validators/mime.go
```diff
@@ -6,10 +6,13 @@ import (
...
 
 // importSchemaMimeTypes is a list of mime-types that are accepted for files uploaded as part of the import schema process (e.g. procedure, internal policy, action plan).
+// Parameter suffixes (e.g. "; charset=utf-8") are normalized away by MimeTypeValidator, so each
+// media type only needs to be listed once here.
 var importSchemaMimeTypes = []string{
-	"text/plain; charset=utf-8", "text/plain",
+	"text/plain",
 	"text/markdown",
 	"text/x-markdown",
+	"text/html",
```
golanglemonade: I dont think we should do this right now either; html import requires a lot of validation + scrubbing and should be tracked in a different set of work. This was only to support word docs.

internal/objects/validators/mime.go
papistacoding: pkg/objects/validators.go now has normalizeMimeType

internal/ent/hooks/documents.go
papistacoding: I see, removed

internal/ent/hooks/documents.go
papistacoding: got it, helper deleted, now parsiing

internal/ent/hooks/documents.go
papistacoding: removed

internal/objects/validators/mime.go
papistacoding: pkg/objects/validators.go now has normalizeMimeType

internal/ent/hooks/documents.go
```diff
@@ -225,20 +226,29 @@ var client = &http.Client{
...
-func importURLToSchema(m importSchemaMutation) error {
+// ErrInvalidImportURL is returned when an import URL is rejected before any network call.
+var ErrInvalidImportURL = errors.New("invalid import URL")
+
+var allowedImportSchemes = map[string]struct{}{"http": {}, "https": {}}
+
+// importURLToSchema fetches content from a URL, parses it by MIME type, and writes the
+// sanitized result into the mutation details. The scheme allowlist blocks file://, etc.;
+// it does not block private IPs — callers must trust the URL source.
+func importURLToSchema(parentCtx context.Context, m importSchemaMutation) error {
```
golanglemonade: nit: go convention would have this named `ctx`
```suggestion
func importURLToSchema(ctx context.Context, m importSchemaMutation) error {
```

golanglemonade: this is an internal function, that isn't really necessary 
```suggestion
// importURLToSchema fetches content from a URL, parses it by MIME type, and writes the
// sanitized result into the mutation details. The scheme allowlist blocks file://, etc.
func importURLToSchema(parentCtx context.Context, m importSchemaMutation) error {
```

internal/ent/hooks/documents.go
```diff
@@ -225,20 +226,29 @@ var client = &http.Client{
...
 	}
 
-	_, err := url.Parse(downloadURL)
+	u, err := url.Parse(downloadURL)
 	if err != nil {
-		return err
+		return fmt.Errorf("%w: %s", ErrInvalidImportURL, err.Error())
+	}
+	if _, ok := allowedImportSchemes[strings.ToLower(u.Scheme)]; !ok {
+		return fmt.Errorf("%w: scheme %q not allowed", ErrInvalidImportURL, u.Scheme)
```
golanglemonade: this should be a validator on the schema, not within this function

internal/ent/hooks/revision.go
```diff
@@ -78,7 +78,6 @@ func SetNewRevision(ctx context.Context, mut MutationWithRevision) error {
 
 	revisionBump, ok := models.VersionBumpFromRequestContext(ctx)
 	if !ok {
-		// derive based on if there were meaningful updates to the details of the document
 		revisionBump = &models.Patch
 		if detailsUpdated(ctx, mut) {
 			revisionBump = &models.Minor
```
golanglemonade, line 90: shouldn't this still have the logic to not check detailsUpdated on `EXTERNAL_MANAGED`?

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,58 @@ const (
...
+	if err != nil {
+		logx.FromContext(ctx).Debug().Err(err).Msg("could not read old file_id; treating as file change")
+		return true
+	}
+
+	oldStr := ""
+	if oldID != nil {
+		oldStr = *oldID
+	}
+
```
golanglemonade: the `Old` functions _only_ work on `OpUpdateOne`, they should not be checked if `OpUpdate` or `OpCreate`

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,65 @@ const (
 	detailsFieldName     = "details"
 )
 
+// documentMutation is the subset of revision-bearing mutations that carry a document file
+// and a management mode (action plans, internal policies, procedures — not standards/templates).
+type documentMutation interface {
+	Op() ent.Op
```
golanglemonade: use the common interface 
```suggestion
	utils.GenericMutation
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,65 @@ const (
...
+	if v, ok := mut.ManagementMode(); ok && v.IsValid() {
+		return v
+	}
+
+	if mut.Op().Is(ent.OpUpdateOne) {
+		if v, err := mut.OldManagementMode(ctx); err == nil && v.IsValid() {
+			return v
+		} else if err != nil {
+			logx.FromContext(ctx).Debug().Err(err).Msg("could not read old management_mode; defaulting to OPENLANE_MANAGED")
+		}
```
golanglemonade: `v.IsValid()` isn't really necessary when using the typed functions. 
```
IsValid reports whether v represents a value. It returns false if v is the zero Value. If [Value.IsValid](https://pkg.go.dev/reflect#Value.IsValid) returns false, all other methods except String panic. Most functions and methods never return an invalid Value. If one does, its documentation states the conditions explicitly.
```

Use `info` so we'd see these in production logs; this isn't really a case we'd want to just keep falling bac to

```suggestion
		if v, err := mut.OldManagementMode(ctx); err == nil {
			return v
		}
		
		logx.FromContext(ctx).Info().Err(err).Msg("could not read old management_mode; defaulting to OPENLANE_MANAGED")
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,65 @@ const (
...
+		oldStr = *oldID
+	}
+
+	if cleared {
+		return oldStr != ""
+	}
+
+	return oldStr != newID
+}
+
```
golanglemonade, line 172: I wouldn't just commit t his as the indentation probably is correct and you'll get a linter error, but I dont think the logic is quite right as is. 
```suggestion
	newID, set := mut.FileID()
	cleared := mut.FileIDCleared()
	if !set && !cleared {
		return false
	}
	
	switch mut.Op() {
		case ent.OpCreate:
		     if set {
					return true
			} else {
					return false
			}
		case ent.OpUpdate, ent.OpDelete, ent.OpDeleteOne:
		   return false
	}
		   
    // check the UpdateOne case
	oldID, err := mut.OldFileID(ctx)
	if err != nil {
		logx.FromContext(ctx).Debug().Err(err).Msg("could not read old file_id; treating as file change")
		return true
	}

	if oldID == nil {
			if set {
					// we have no old id, and we have a new one set
					return true
			} else {
			   return false
			}
	}

	return *oldStr != newID
}

```

golanglemonade: ughhh ignore all this. The hook only runs on `OpUpdateOne`; let me simplify this.

internal/ent/hooks/revision.go
golanglemonade: ignore; SetNewRevision is only called on `hook.HasOp(ent.OpUpdateOne),`

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,65 @@ const (
...
+	if oldID != nil {
+		oldStr = *oldID
+	}
+
+	if cleared {
+		return oldStr != ""
+	}
+
+	return oldStr != newID
+}
```
golanglemonade:
```suggestion
// fileChanged reports whether an OpUpdateOne mutation replaces, sets, or clears the document's file.
func fileChanged(ctx context.Context, mut documentMutation) bool {
	newID, set := mut.FileID()
	cleared := mut.FileIDCleared()
	if !set && !cleared {
		return false
	}

	oldID, err := mut.OldFileID(ctx)
	if err != nil {
		logx.FromContext(ctx).Debug().Err(err).Msg("could not read old file_id; treating as file change")
		return true
	}

	if oldID == nil {
		if set {
			return true
		}
		
		return false
	}


	return *oldStr != newID
}
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,65 @@ const (
...
+	if mut.Op().Is(ent.OpUpdateOne) {
+		if v, err := mut.OldManagementMode(ctx); err == nil && v.IsValid() {
+			return v
+		} else if err != nil {
+			logx.FromContext(ctx).Debug().Err(err).Msg("could not read old management_mode; defaulting to OPENLANE_MANAGED")
+		}
+	}
+
+	return enums.DocumentManagementModeOpenlaneManaged
+}
```
golanglemonade:
```suggestion
```suggestion
// managementModeFor returns the mode that will be in effect after this OpUpdateOne mutation,
// defaulting to OPENLANE_MANAGED for pre-existing rows (NULL column) and create-path mutations.
func managementModeFor(ctx context.Context, mut documentMutation) enums.DocumentManagementMode {
	if v, ok := mut.ManagementMode(); ok {
		return v
	}

	if v, err := mut.OldManagementMode(ctx); err == nil {
			return v
	} 
			
	logx.FromContext(ctx).Info().Msg("could not determine management mod; defaulting to OPENLANE_MANAGED")
	
	return enums.DocumentManagementModeOpenlaneManaged
}
```

internal/ent/hooks/revision.go
```diff
@@ -117,6 +123,65 @@ const (
 	detailsFieldName     = "details"
 )
 
+// documentMutation is the subset of revision-bearing mutations that carry a document file
+// and a management mode (action plans, internal policies, procedures — not standards/templates).
+type documentMutation interface {
+	utils.GenericMutation
```
golanglemonade: Sorry, I missed the hook only checked OpUpdateOne already and couldn't tell from the helper functions in the code github was showing me. This isn't needed at all. 
```suggestion
```

internal/ent/hooks/revision.go
papistacoding: markign this one as resolved then

golanglemonade (approved):

PR #2403, opened 2026-05-15
fix(files): derive providedFileExtension from OriginalName

papistacoding:
this is to support : https://github.com/theopenlane/openlane-ui/pull/1859


internal/objects/store/persistence.go
```diff
@@ -104,6 +104,15 @@ func createFile(ctx context.Context, f pkgobjects.File) (*ent.File, error) {
 	return entFile, nil
 }
 
+// resolveProvidedExtension returns the file extension derived from the client-provided
+// filename. OriginalName is the canonical source — the separate ProvidedExtension field
+// on pkgobjects.File is not populated by the multipart parser, so deriving from
+// OriginalName avoids drift and works for filenames containing spaces, multiple dots,
+// or other special characters.
+func resolveProvidedExtension(f pkgobjects.File) string {
+	return filepath.Ext(f.OriginalName)
```
golanglemonade, line 116: It should still use it if its set, but fall back to the original name
```suggestion
    if f.ProvidedExtension != "" {
        return f.ProvidedExtension
    }
    
	return filepath.Ext(f.OriginalName)
```

golanglemonade (approved):

PR #2405, opened 2026-05-15
fix(storage): serve previewable MIMEs inline; re-detect octet-stream

papistacoding:


common/storagetypes/disposition.go
```diff
@@ -0,0 +1,61 @@
+package storagetypes
```
golanglemonade: why is this in common? This module is only for things that are needed by other repos to avoid a cyclical depdency on core. This looks like it can go in pkg/objects?

internal/httpserve/handlers/filedownload.go
```diff
@@ -137,9 +137,24 @@ func (h *Handler) FileDownloadHandler(ctx echo.Context, openapi *OpenAPIContext)
 		return h.InternalServerError(ctx, ErrProcessingRequest, openapi)
 	}
 
+	// Format the disposition through mime.FormatMediaType so the filename is
+	// properly quoted/escaped; ProvidedFileName is user-supplied and could
+	// otherwise break out of the quoted-string or split headers via CR/LF.
+	// FormatMediaType returns "" for unrepresentable filenames; fall back to
+	// the bare disposition rather than emitting nothing.
```
golanglemonade: this comment does not need to be so verbose

internal/httpserve/handlers/filedownload.go
```diff
@@ -137,9 +137,24 @@ func (h *Handler) FileDownloadHandler(ctx echo.Context, openapi *OpenAPIContext)
...
+		disposition = dispBase
+	}
+
 	headers := ctx.Response().Header()
 	headers.Set(echo.HeaderContentType, downloadFile.DetectedContentType)
-	headers.Set(echo.HeaderContentDisposition, fmt.Sprintf("attachment; filename=\"%s\"", downloadFile.ProvidedFileName))
+	headers.Set(echo.HeaderContentDisposition, disposition)
+	// Prevent the browser from MIME-sniffing the body to a more dangerous type
+	// than DetectedContentType claims — critical now that this path may serve
+	// content with inline disposition.
```
golanglemonade:
```suggestion
```

golanglemonade (approved):

PR #2429, opened 2026-05-25
Cloudflare security insights as findings

adelowo:
- updated sdk to v7
-  sdk format does not match api requests here again so added a `option.WithResponseBodyInto(&response)`. see https://gist.github.com/adelowo/9efee681baf63c3a02d5913807c3bf4d
- security insights are now stored as findings.
- removed `dedicatedDb` from graphql
- added a new `SkipFirstReconcile`. enabled for cloudflare syncing. this is because unlike others that go through the oauth2 flow to get connected. the ctx here that would be used to insert the findings would be from directly from  the http request https://github.com/theopenlane/core/blob/main/internal/httpserve/handlers/integration_config.go#L32-L57

CleanShot 2026-05-25 at 20 12 12

CleanShot 2026-05-25 at 20 12 00


internal/integrations/runtime/credentials.go
```diff
@@ -296,11 +296,13 @@ func (r *Runtime) reconcileCredential(ctx context.Context, installation *ent.Int
 		return err
 	}
 
-	if wasFirstConnection {
+	if wasFirstConnection && !connection.SkipFirstReconcile {
```
matoszz: this feels problematic and affects much more than just cloudflare - have you tested with all other available integrations?

internal/httpserve/handlers/integration_config.go
```diff
@@ -36,6 +36,12 @@ func (h *Handler) ConfigureIntegrationProvider(ctx echo.Context, openapiCtx *Ope
 		return h.Unauthorized(ctx, auth.ErrNoAuthUser, openapiCtx)
 	}
 
+	requestCtx = auth.WithCaller(
```
golanglemonade: Can we name this `systemCtx` and make sure we aren't using it for things like the `UpdateOne`. Thats a highly privileged context and it should only be used for system operations. We can also remove the duplicate on line 94 

```
	systemCtx := privacy.DecisionContext(requestCtx, privacy.Allow)
```

I would consider putting it in `Reconcile` instead of in the handler directly as well so its not accidentally used when it shouldn't be

golanglemonade (commented):
can you add an example json response to `integrations/definitions/cloudflare/examples` and a test for the mappings?

matoszz (approved):

PR #2433, opened 2026-05-26
Map domain registrations on cloudflare to assets

adelowo:


internal/integrations/definitions/cloudflare/builder.go
```diff
@@ -110,6 +112,28 @@ func Builder() registry.Builder {
...
+					Ingest: []types.IngestContract{
+						{
+							Schema: integrationgenerated.IntegrationMappingSchemaAsset,
+						},
+					},
+					IngestHandle:        AssetCollect{}.IngestHandle(),
+					SkipDefaultLookback: true,
+					RequiredPermissions: []string{"Registrar Domains Read"},
+					ReconcileSchedule: gala.NewFullFetchSchedule(
+						gala.WithMinInterval(24*time.Hour),
```
golanglemonade: Why change this? this would mean if it fails it won't retry again for 24 hours?

internal/integrations/definitions/cloudflare/operation_asset_sync.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/theopenlane/core/pkg/logx"
+)
+
+// AssetCollect collects Cloudflare domain registrations and stores them as assets
+type AssetCollect struct{}
+
+// IngestHandle adapts asset collection to the ingest operation registration boundary
+func (a AssetCollect) IngestHandle() types.IngestHandler {
+	return providerkit.WithClientRequestConfig(cloudflareClient, assetSyncOperation, ErrOperationConfigInvalid, func(ctx context.Context, request types.OperationRequest, client *cf.Client, _ AssetSync) ([]types.IngestPayloadSet, error) {
+		return a.Run(ctx, request.Credentials, client)
```
golanglemonade: If you don't need the config you can use `WithClientRequest` instead of `WithClientRequestConfig`
```suggestion
	return providerkit.WithClientRequest(cloudflareClient, assetSyncOperation, ErrOperationConfigInvalid, func(ctx context.Context, request types.OperationRequest, client *cf.Client) ([]types.IngestPayloadSet, error) {
		return a.Run(ctx, request.Credentials, client)
```

internal/integrations/definitions/cloudflare/operation_asset_sync.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func fetchRegistrarRegistrations(ctx context.Context, client *cf.Client, accountID string) ([]registrar.Registration, error) {
+	domains := make([]registrar.Registration, 0)
+
+	for cursor := ""; ; {
+		var response cloudflareRegistrationsResponse
+		params := registrar.RegistrationListParams{
+			AccountID: cf.F(accountID),
+			PerPage:   cf.F(int64(50)),
```
golanglemonade: probably going to have a linter error here for magic numbers

internal/integrations/definitions/cloudflare/builder.go
adelowo: domain registrations won't change often I suppose but I didn't think so much of retries 🤔

internal/integrations/definitions/cloudflare/operation_asset_sync.go
adelowo: copy pasta from seciurity insights :))

internal/integrations/definitions/cloudflare/builder.go
golanglemonade: yeah I think the max interval is correct, and it increases over time when new records are not pulled; but I guess in this case theres no timestamp update so we pull them every time?

adelowo: yes correct. no timestamp update here

golanglemonade (commented):
failing test, `TestDomainRegistrationsAssetMapping` otherwise LGTM

golanglemonade (approved):

PR #2440, opened 2026-05-29
prevent owner direct transfers

adelowo:
- Owner role can never be changed directly - only transferred
- a user cannot grant a higher permissions than what they have or demote. e.g member cannot take admin to member or take a member to admin. same way


internal/ent/hooks/orgmembers.go
```diff
@@ -141,6 +142,50 @@ func HookUpdateManagedGroups() ent.Hook {
...
+				return nil, ErrOrgOwnerCannotBeUpdated
+			}
+
+			return next.Mutate(ctx, m)
+		})
+	}, ent.OpUpdate|ent.OpUpdateOne)
+}
+
+func getUserMembershipRole(ctx context.Context, m *generated.OrgMembershipMutation) (enums.Role, error) {
+	id, ok := m.ID()
```
golanglemonade: the hook allows for `OpUpdate` but this only checks `m.ID` and not `m.IDs()`. If we allow for bulk update of org members you need to check both + update your query

internal/ent/privacy/rule/membership.go
```diff
@@ -36,3 +41,68 @@ func AllowSelfOrgMembershipDelete() privacy.OrgMembershipMutationRuleFunc {
...
+
+// AllowOrgMemberRoleUpdate allows role updates using the same ceiling as invites.
+func AllowOrgMemberRoleUpdate() privacy.OrgMembershipMutationRuleFunc {
+	return privacy.OrgMembershipMutationRuleFunc(func(ctx context.Context, m *generated.OrgMembershipMutation) error {
+		newRole, ok := m.Role()
+		if !ok {
+			return privacy.Skip
+		}
+
+		id, ok := m.ID()
```
golanglemonade: same here, this only works on single update, not bulk

golanglemonade: this is also basically the same code written in the hooks; can we reuse the same thing instead of writing it twice? OR can we just have the rule and not also a hook? Seems like we just need one?

internal/ent/schema/orgmembership.go
```diff
@@ -146,6 +147,10 @@ func (OrgMembership) Policy() ent.Policy {
 			ent.OpDelete|ent.OpDeleteOne,
 			rule.AllowSelfOrgMembershipDelete(),
 		),
+		policy.WithOnMutationRules(
+			ent.OpUpdate|ent.OpUpdateOne,
+			rule.AllowOrgMemberRoleUpdate(),
```
golanglemonade, line 152: Do we prevent the deletion of the owner org member?

internal/ent/privacy/rule/membership.go
golanglemonade: ignore second comment

internal/ent/schema/orgmembership.go
adelowo: yes, owner is restricted from deletion

golanglemonade (approved):

PR #2441, opened 2026-05-29
fix: org owned objects with access via api token, return error, confi…

golanglemonade:
- fixes a bug where org owned objects (e.g. not object owned) were still returned even if the api token didn't have the scope
- now returns an error instead of empty results if the token doesn't have all scopes required for the request, info log logged with object type + token id for debugging
- pulls in iam v0.32.0 to update default config for module file
- removes authentik db creation from main setup, moves to it's own init script; removes windmill db as its not used


adelowo (approved):

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/960

### Changes Preview:

- 🔄 Merged Helm values.yaml

PR #2453, opened 2026-06-03
NDA Approval Email

matoszz:
Adds new NDA approval email that gets sent to owners, super admins, and admins - the construction uses a single email with multiple recipients rather than a single email per user. Easily changed if the behavior is undesirable. Since you can't link to a specific approval request (they are all in a table) the URL behind the button is static.


golanglemonade (dismissed):
LGTM but you have some failing tests - probably related to sending out two emails now instead of 1 and its string checking the email text

golanglemonade (approved):

PR #2467, opened 2026-06-10
Allow for dynamic assignment of roles to users

adelowo:
- added `/account/organization-roles` ( post, delete and get ) to 
  - fetch roles the currently authenticated user has 
  - assign a new role to users/groups
  - delete a role from users/groups
- extend fga parser with new  `@role`


common/openapi/models.go
```diff
@@ -2561,9 +2564,92 @@ var ExampleRolesRequest = RolesRequest{}
...
+		},
+		{
+			ID:          "risk_manager",
+			Name:        "Risk Manager",
+			Description: "Can manage risks, vulnerabilities, and findings",
+		},
 	},
 }
+
+// OrganizationRolesRequest contains roles that can be assigned to users or groups.
```
golanglemonade:
```suggestion
// OrganizationRolesRequest contains functional roles that can be assigned to users or groups in addition to their base level org role such as member, admin, etc. 
```

internal/httpserve/handlers/roles.go
```diff
@@ -2,33 +2,235 @@ package handlers
...
+
+	ok, err = fgamodel.IsOrganizationRole(in.Role)
+	if err != nil {
+		logx.FromContext(reqCtx).Error().Err(err).Msg("error retrieving organization roles")
+		return h.InternalServerError(ctx, ErrProcessingRequest, openapi)
+	}
+
+	if !ok {
+		return h.BadRequest(ctx, ErrInvalidInput, openapi)
+	}
```
golanglemonade: Should this be in the `Validate` function instead?

golanglemonade (commented):
LGTM. Can you add a few graphapi tests that assign some of these functional roles and make sure the user can then edit. I dont think I added tests originally when making these. A good example would be a new member user in the policy tests, make sure they can't edit the policy by default, give them the policy manager role and now they should be able to edit/delete any policy.

golanglemonade (approved):

PR #2480, opened 2026-06-13
feat: azure entra id and one drive integrations

golanglemonade:
- Tested + updated the Azure EntraID integration for directory sync
   - When testing in dev against an account that is not used as a primary IDP (e.g. our emails are not coming from MSFT so you cannot login correctly) you must specifiy the `defaulttentant`. This should only be used in dev, in production this should be empty
   
image

- Added onedrive integration for document policy view, needs the updated UI: https://github.com/theopenlane/openlane-ui/pull/1933
   - This uses an iframe to view in console - non-personal onedrive accounts do not export to HTML, so needed to either do iframe, PDF, or try to get the HTML ourselves. Went with support for iframe in the UI and have PDF in order to allow bulk PDF export (will be a future PR) from our UI. 
   - Added a common interface for both exports (google and onedrive) so that usage outside the integration (e.g. in the policy resolver) can use the same functions and not a long switch statement


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/992

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

adelowo (approved):

PR #2483, opened 2026-06-15
fix: confirm SSO emails before session creation

Little709:
## Summary

SSO/OIDC callback provisioning currently creates or loads the user, then immediately builds the authenticated session. Unlike the OAuth registration path, it does not mark the OIDC-verified email address as confirmed first.

That means hooks tied to email confirmation, including matching-domain organization auto-join and pending invite acceptance, do not run before session generation.

This change marks the user's email as confirmed after successful OIDC token exchange and user provisioning, before building the authenticated session. The existing email-confirmation hooks can then perform the expected organization membership setup.

## Why

OIDC identity providers have already verified the email claim in the returned ID token. Treating that email as confirmed keeps SSO behavior aligned with the existing organization auto-join flow.

## Testing

Not run locally. Verified by code inspection against the existing `setEmailConfirmed` helper and user-setting email-confirmation hook behavior.


golanglemonade (approved):
thanks @Little709!

golanglemonade:
@Little709 the test failure looks unrelated to your change, I'm taking a look.

golanglemonade:
> @Little709 the test failure looks unrelated to your change, I'm taking a look.

@Little709 Try a rebase, it should pass now

matoszz:
@Little709 I'm incorporating your proposed changed into https://github.com/theopenlane/core/pull/2520 , so going to close this one 🙏  Thank you for your contribution and the valid addition; this was a good catch 👍

PR #2510, opened 2026-06-17
Auditor fixes

adelowo:


internal/ent/schema/review.go
```diff
@@ -294,7 +294,6 @@ func (Review) Annotations() []schema.Annotation {
 func (r Review) Policy() ent.Policy {
 	return policy.NewPolicy(
 		policy.WithMutationRules(
-			policy.CheckOrgWriteAccess(),
```
golanglemonade, line 297: I think this just needs to be re-ordered so its after CheckCreateAccess instead of deleted, it allows admins+ access to all reviews - I think it might just need to come after policy.CheckCreateAccess because it returns access denied

golanglemonade (approved):

PR #2516, opened 2026-06-17
disable integrations execution if org does not have a valid subscription

adelowo:
also includes the check and disabling on the billing reconciliation process too


internal/entitlements/reconciler/reconciler_test.go
```diff
@@ -84,3 +86,48 @@ func TestReconcileResult(t *testing.T) {
...
+		{
+			name: "subscription is active",
+			sub: &ent.OrgSubscription{
+				Active:                   true,
+				StripeSubscriptionStatus: string(stripe.SubscriptionStatusActive),
+			},
+			want: true,
+		},
+		{
+			name: "subscription is inactive",
```
golanglemonade, line 109: can you add one more test case for active = false and status = trialing so we make sure we don't disable integrations for tirals?

golanglemonade (approved):

PR #2520, opened 2026-06-18
Configurable SSO Exclusion (Org level + User level), Organization Consent + Support Access Capabilities

matoszz:
- Organizations can exempt individual members or entire email domains from enforced SSO
- Invitations can grant a member an SSO exemption the moment they accept
- A domain can never sit in both the SSO-exempt and allowed lists at once
- Multi-factor enforcement still applies to exempt users
- SSO enforcement no longer disables auto-join or allowed-domain membership controls
- Personal access tokens skip SSO authorization when their owner is exempt
- Adds the ability for Openlane support to enter a consenting organization via a shared password plus a domain-restricted identity provider (support access should not be predicated on a literal Openlane user, so this was made to be purely environment based)
- Support sessions stay scoped to the one consented organization and attribute every action to the support identity while recording the real engineer behind it
- Support and impersonation sessions can be ended and revoked at logout, not just left to expire


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/1010

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

common/go.mod
```diff
@@ -91,12 +91,12 @@ require (
...
 	golang.org/x/exp v0.0.0-20260508232706-74f9aab9d74a // indirect
 	golang.org/x/mod v0.37.0 // indirect
-	golang.org/x/net v0.56.0 // indirect
+	golang.org/x/net v0.55.0 // indirect
 	golang.org/x/sync v0.21.0 // indirect
 	golang.org/x/sys v0.46.0 // indirect
 	golang.org/x/text v0.38.0 // indirect
-	golang.org/x/tools v0.46.0 // indirect
-	google.golang.org/genproto/googleapis/api v0.0.0-20260526163538-3dc84a4a5aaa // indirect
+	golang.org/x/tools v0.45.0 // indirect
```
golanglemonade: Can you fix these downgrades when you rebase 🙏

internal/httpserve/handlers/support_test.go
```diff
@@ -0,0 +1,270 @@
...
+	assert.Equal(t, org.ID, cbOut.OrganizationID)
+	assert.Equal(t, "engineer@theopenlane.io", cbOut.Impersonator)
+
+	// the minted token must carry both identities: the virtual support user and the individual
+	claims, err := suite.h.TokenManager.ValidateImpersonationToken(context.Background(), cbOut.Token)
+	require.NoError(t, err)
+	assert.Equal(t, "01JSPPRT000000000000000000", claims.UserID, "target is the virtual support identity")
+	assert.Equal(t, "engineer@theopenlane.io", claims.ImpersonatorID, "impersonator is the individual from the IdP")
+	assert.Equal(t, "support", claims.Type)
+	assert.Equal(t, org.ID, claims.OrgID)
```
golanglemonade, line 145: Can you add a few simple checks to make sure support can query data in the org + make updates?

internal/httpserve/handlers/support.go
```diff
@@ -0,0 +1,296 @@
...
+	token, err := h.TokenManager.CreateImpersonationToken(reqCtx, tokens.CreateImpersonationTokenOptions{
+		ImpersonatorID:    individualEmail,
+		ImpersonatorEmail: individualEmail,
+		TargetUserID:      cfg.SubjectID,
+		TargetUserEmail:   cfg.Email,
+		OrganizationID:    orgCookie.Value,
+		Type:              string(auth.SupportImpersonation),
+		Reason:            reason,
+		Duration:          duration,
+		Scopes:            []string{"read", "write"},
```
golanglemonade: What do these scopes actually do? Seems like this might be part of a dead-code path?

golanglemonade (commented):
LGTM; just a few questions/comments

internal/ent/hooks/orgsetting_sso.go
```diff
@@ -11,29 +13,80 @@ import (
...
+// hasIdentityProviderConfigFields reports whether all identity provider configuration fields are part of
+// the mutation, matching the original trigger so the enforcement reset and validation only run for IdP changes
+func hasIdentityProviderConfigFields(m *generated.OrganizationSettingMutation) bool {
+	_, provider := m.IdentityProvider()
+	_, clientID := m.IdentityProviderClientID()
+	_, clientSecret := m.IdentityProviderClientSecret()
+	_, discovery := m.OidcDiscoveryEndpoint()
+	_, enforced := m.IdentityProviderLoginEnforced()
+
+	return provider && clientID && clientSecret && discovery && enforced
```
golanglemonade: Shouldn't this be an _or_? This would require all the fields to be set to reset the status but couldn't you just set one?

internal/ent/hooks/orgsetting_sso.go
```diff
@@ -11,29 +13,80 @@ import (
...
+}
+
+// validateExemptDomainsDisjoint ensures a domain is never both in allowed_email_domains and
+// sso_exempt_domains; a domain cannot be allowed for access and excluded from SSO enforcement at once
+func validateExemptDomainsDisjoint(ctx context.Context, m *generated.OrganizationSettingMutation) error {
+	exempt := resolvedDomains(ctx, m.SSOExemptDomains, m.OldSSOExemptDomains)
+	allowed := resolvedDomains(ctx, m.AllowedEmailDomains, m.OldAllowedEmailDomains)
```
golanglemonade: Doesn't allowedDomains prevent users without those domains from logging in at all? So if I had:

AllowedDomains: `acme.io`, `myauditor.com` - but I want to disabled sso for `myauditor.com` this would prevent that? 

I think if we want this we need to change AllowedDomains to only be used for the auto-join functionality and not access to the org?

internal/ent/privacy/rule/scopes.go
```diff
@@ -120,6 +120,13 @@ func CheckSubjectScope(ctx context.Context, objectType string, relation string,
 		return privacy.Skip
 	}
 
+	// org-scoped support sessions hold every scope for their organization, equivalent to a fully scoped
+	// token; delete is excluded so support cannot remove organizations or their objects
+	if caller, ok := auth.CallerFromContext(ctx); ok && caller != nil && caller.OrganizationID != "" &&
+		caller.Has(auth.CapOrgSupport) && !strings.HasPrefix(scopedRelation, CanDeletePrefix) {
+		return privacy.Allow
+	}
```
golanglemonade, line 135: We should short circuit the `delete` requests so they don't go and check FGA next, can either just do prviacy.Deny, or use the same `ErrRequiredScopeNotSet` we use for api tokens. 


```suggestion
	if caller, ok := auth.CallerFromContext(ctx); ok && caller != nil && caller.OrganizationID != "" &&
		caller.Has(auth.CapOrgSupport) {
		if !strings.HasPrefix(scopedRelation, CanDeletePrefix) {
			return privacy.Allow
		}
		
		logx.FromContext(ctx).Info().Str("scope", scopedRelation).Str("user_id", caller.SubjectID).Msg("support attempting to perform action thats not allowed")
		
		return ErrRequiredScopeNotSet
	}
```

internal/ent/hooks/orgsetting_sso.go
matoszz: AllowedDomains IS only used for auto join functionality, you don't have to set allowed domains to have SSO enforcement on or off, and AllowedDomains is mutually exclusive with disabled for SSO

internal/ent/hooks/orgsetting_sso.go
matoszz: Not entirely sure what you mean

golanglemonade: this needs an or `return provider || clientID || clientSecret || discovery || enforced`  so that if _any_ of these fields are changed SSO tested is reset. This matches the previous behavior.

golanglemonade (dismissed):

adelowo (approved):

PR #2525, opened 2026-06-18
Add Trust Center Subscribers, Email Send, Subprocessor Update Notifications

matoszz:
- Enable external users to subscribe to trust center updates
- Unsubscribe via a tokenized link, with per-recipient unsubscribe URLs in emails
- Send subscriber emails through campaigns, branded automatically from the trust center's configuration (or fallback if not configured)
- Email subscribers when a trust center post is published, only when the author opts in
- Email subscribers when subprocessors change, only when the trust center has that option enabled
- Debounce and coalesce update emails so subscribers get one notification instead of a burst
- Track who was notified and which subprocessors were included in each notification
- Automatically link new subscribers to existing contacts and users


internal/ent/schema/subscriber.go
```diff
@@ -114,7 +127,10 @@ func (s Subscriber) Mixin() []ent.Mixin {
 	return mixinConfig{
 		additionalMixins: []ent.Mixin{
 			newOrgOwnedMixin(s,
-				withSkipTokenTypesObjects(&token.VerifyToken{}, &token.SignUpToken{}), withSkipForSystemAdmin()),
+				withSkipTokenTypesObjects(&token.VerifyToken{}, &token.SignUpToken{}),
+				withSkipForSystemAdmin(),
+				withAllowAnonymousTrustCenterAccess(true),
```
golanglemonade, line 132: Why do anon trust center users need to access the subscriber schema? This would allow someone to use a anon token and pull all subscribers to a trust center?

internal/ent/schema/subscriber.go
```diff
@@ -136,10 +169,18 @@ func (Subscriber) Hooks() []ent.Hook {
...
 func (Subscriber) Indexes() []ent.Index {
 	return []ent.Index{
+		// legacy organization-level subscribers (no trust center) remain unique per org
 		index.Fields("email", ownerFieldName).
 			Unique().
 			Annotations(
-				entsql.IndexWhere("deleted_at is NULL and unsubscribed = false"),
+				entsql.IndexWhere("deleted_at is NULL and unsubscribed = false and trust_center_id IS NULL"),
+			),
+		// trust center subscribers are unique per trust center, allowing the same email to
```
golanglemonade, line 178: Organizations can only have one trust center though? https://github.com/theopenlane/core/blob/ea94b3492ce6abf9b4798504a7c96f3ee1c916b8/internal/ent/hooks/trustcenter.go#L41

golanglemonade (approved):

PR #2553, opened 2026-06-29
Store MD5 hash on every file upload and validate trustcenter submissions

adelowo:


golanglemonade (approved):
Looks good, can you remove the no-longer needed fields from the client side (trust center ui repo) and mark the PR as dependent of this one?

PR #2554, opened 2026-06-30
Auditor should be able to view org avatars

adelowo:


fga/model/roles/roles.fga
```diff
@@ -14,7 +14,7 @@ extend type organization
     define owner: [user] or owner from parent
 
     # @create: comment, discussion, task
-    # @view: control, subcontrol, program, evidence, internal_policy, procedure, review, platform
+    # @view: control, subcontrol, program, evidence, file, internal_policy, procedure, review, platform
```
golanglemonade: This would give them access to _all_ files in the org, I dont think we want that, the parent_viewer on fiel you have there should do the trick?

fga/tests/tests.yaml
```diff
@@ -830,6 +830,18 @@ tests:
           can_view: true
           can_edit: true
           can_delete: true
+      - user: user:ulid-auditor # organization auditor should be able to view organization files
+        object: file:file-org-context
+        assertions:
+          can_view: true
```
golanglemonade, line 836: Auditor should _only_ be able to see files based on the parent, not all files. 
```suggestion
          can_view: false
```

golanglemonade (approved):

PR #2559, opened 2026-07-04
Objects created by integration runs do not have audit fields set

adelowo:
also makes sure we do not run subscription checks when entitlements is not enabled.

this is really useful for local dev work e.g or for core installations without stripe payments. As integrations will not work without this fix except orgs have stripe/subscriptions


internal/integrations/operations/ingest.go
```diff
@@ -65,6 +67,21 @@ type mappedIngestRecord struct {
 	Payload json.RawMessage
 }
 
+func setIntegrationAuthCaller(ctx context.Context, integration *ent.Integration) context.Context {
+	if integration == nil || integration.ID == "" {
+		return ctx
+	}
+
+	caller := &auth.Caller{
```
golanglemonade: As we chatted in slack; this should be updated to just set the subject, not recreate a new caller

internal/integrations/operations/dispatcher.go
```diff
@@ -20,7 +21,7 @@ import (
 // Dispatch validates and enqueues one operation execution request. When
 // DispatchRequest.Runtime is true, no DB integration lookup is performed and
 // the client is resolved from the registry at execution time
-func Dispatch(ctx context.Context, reg *registry.Registry, db *ent.Client, runtime *gala.Gala, req types.DispatchRequest) (types.DispatchResult, error) {
+func Dispatch(ctx context.Context, reg *registry.Registry, db *ent.Client, runtime *gala.Gala, integrationActorConfig identity.Config, req types.DispatchRequest) (types.DispatchResult, error) {
```
matoszz, line 24: the context would carry the attributes, the config shouldn't need to be passed

internal/integrations/operations/dispatcher.go
```diff
@@ -80,7 +81,7 @@ func Dispatch(ctx context.Context, reg *registry.Registry, db *ent.Client, runti
 	var runID string
 
 	if installation != nil && !operation.Policy.SkipRunRecord {
-		runRecord, err := CreatePendingRun(ctx, db, installation, types.DispatchRequest{
+		runRecord, err := CreatePendingRun(ctx, db, installation, integrationActorConfig, types.DispatchRequest{
```
matoszz, line 84: the context would carry the caller, the config shouldn't need to be passed

internal/integrations/operations/dispatcher.go
```diff
@@ -100,6 +101,8 @@ func Dispatch(ctx context.Context, reg *registry.Registry, db *ent.Client, runti
 	oc := types.NewOperationContext(ownerID, req.Operation, src)
 
 	emitCtx := gala.WithOperationContext(ctx, oc)
+	emitCtx = identity.WithIntegrationCaller(emitCtx, installation, integrationActorConfig)
```
matoszz, line 104: you should set the caller prior to gala.WithOperationContext

internal/integrations/runtime/runtime.go
```diff
@@ -39,6 +40,8 @@ type Config struct {
 	RedisClient *redis.Client
 	// CatalogConfig supplies operator-level credentials for all built-in definitions
 	CatalogConfig catalog.Config
+	// IntegrationActor holds the integration virtual user value used when creating db objects by the integration
```
matoszz, line 43: this doesn't belong here

internal/integrations/runtime/runtime.go
```diff
@@ -61,6 +64,8 @@ type Runtime struct {
 	defaultLookback time.Duration
 	// devMode indicates the server is running in development mode
 	devMode bool
+	// integrationActor supplies the virtual actor values used for integration-created records
```
matoszz, line 67: this does not belong on Runtime

internal/integrations/runtime/runtime.go
```diff
@@ -129,6 +134,11 @@ func (r *Runtime) Redis() *redis.Client {
 	return do.MustInvoke[*redis.Client](r.injector)
 }
 
+// IntegrationActorConfig returns the runtime-owned virtual integration actor config.
+func (r *Runtime) IntegrationActorConfig() identity.Config {
```
matoszz, line 138: this shouldn't be accessed via Runtime

internal/integrations/runtime/runtime.go
```diff
@@ -209,9 +219,10 @@ func New(config Config) (*Runtime, error) {
 
 	injector := do.New()
 	rt := &Runtime{
-		injector:        injector,
-		defaultLookback: lookback,
-		devMode:         config.DevMode,
+		injector:          injector,
+		defaultLookback:   lookback,
+		devMode:           config.DevMode,
+		integrationActor:  config.IntegrationActor,
```
matoszz, line 225: this should not be set on Runtime

matoszz (changes requested):

PR #2570, opened 2026-07-08
Create automated tasks based off onboarding metadata

adelowo:
- Added a new definition file that allows for programmatic task creation 
- Added db level index to idempotency key for tasks


internal/ent/hooks/definitions/onboarding_tasks.yaml
```diff
@@ -0,0 +1,32 @@
+# Onboarding Task definition
+#
+# placeholders such as {{ }} and will be replaced at task creation time.
+# this is to make sure we set the idempotency key correctly
+#
+# `if_key_value` should match the metadata from the frontend and the action you want performed
+# after the evaluation
+- key: onboarding.policies.import_existing
```
golanglemonade, line 8: I like the yaml format for this here; I think we are going to want and end state thats allows for more configurable + flow but I like where this is headed. 

Notes specifically on the info here:
- These are going to need links, a user looking at this task isn't going to know what "using Upload" means. 
- The url to docs is going to need to be more specific to land them in the right place - but maybe we don't even need that here if the UI already has the docs links
- We want a decision tree of "are you using integrations -> do this" or are you doing in openlane -> do this


This will play really well with the onboarding flow updates I'm making off your UI PR; Let me get a little further before you make these updates so we can make sure these are in-sync but wanted to write some initial notes here.

golanglemonade: I also think that we seed the current "suggested actions" as "always included onboarding tasks" so that users can mark those as complete and not be there forever

golanglemonade: One thing I'm thinking about is the disconnect between the front and backend - what if we had a jsonschema of the onboarding questions that the UI could pull so that what we have here always lines up + task creation is more cohesive based on that?

internal/ent/hooks/listeners_onboarding_policy_tasks.go
golanglemonade, line 1: this won't just be "policy" tasks - so the file name needs to be more generic but think this is the right direction generally

PR #2595, opened 2026-07-13
Add `snake_case` JSON tags to mutation input types, entx and entityops additions

matoszz:
- Updated the ent codegen templates so all generated mutation-input types carry `snake_case` JSON tags, letting us marshal/unmarshal payloads directly into the generated inputs instead of going through map translation; includes a rename of the scan vulnerability-IDs field whose case collision made it undecodable over the wire
- Marked assessment, assessment response, finding, remediation, risk, task, vulnerability) as workflow-eligible and added the workflow instance / object-ref schema edges and migrations
- Pulled in latest `entx` and regenerated to create new entityops registry as (hopefully) the single generated surface for cross-object operations schema lookup, target selection, and create-time link injection. Runtime consumers will be added in follow on PR's
- Reworked CEL selector evaluation to bind entities as native typed structs from generated projections so expressions type-check against real field shapes at compile time and evaluate without a decode-to-map hop


internal/ent/schema/assessment.go
```diff
@@ -108,6 +110,12 @@ func (a Assessment) Edges() []ent.Edge {
 		defaultEdgeFromWithPagination(a, IdentityHolder{}),
 		defaultEdgeToWithPagination(a, AssessmentResponse{}),
 		defaultEdgeToWithPagination(a, Campaign{}),
+		edgeFromWithPagination(&edgeDefinition{
+			fromSchema: a,
+			edgeSchema: WorkflowObjectRef{},
+			name:       "workflow_object_refs",
+			ref:        "assessment",
+		}),
```
golanglemonade, line 118: the overrides for name and ref should be the defaults, did this not work the same?  Same comment on every schema the from edge was added
```suggestion
		defaultEdgeFromWithPagination(a, WorkflowObjectRef{}),
```

golanglemonade (approved):

PR #2597, opened 2026-07-14
Auditors should be able to create reviews

adelowo:
```
"error": "ERROR: relation \"reviewers\" does not exist (SQLSTATE 42P01)",
```

we do not have a reviewers table, we also cannot do `accessmap.EdgeView(User{}.Name)` as users table does have
`owner_id` which is why we are doing `NoAuth` here

- updated tests too

- also updated `accessmap.EdgeNoAuth` to `accessmap.EdgeViewCheck(User{}.Name()),` where required. simplified hooks that had to check for the user presence earlier since it will now be handled by access map instead


internal/ent/privacy/policy/checks.go
```diff
@@ -298,6 +299,10 @@ func checkEdgesEditAccess(ctx context.Context, m ent.Mutation, edges []string, a
 
 			// check api token scope first, as api tokens will have full access to object types they have scope for
 			if confirmedInOrganization {
+				if edgeMap.ObjectType == user.Label && edgeMap.CheckViewAccess {
```
golanglemonade, line 305: Why is this skipped? Can you add a comment?

internal/ent/privacy/rule/orgowner.go
```diff
@@ -236,6 +239,34 @@ func EnsureObjectInOrganization(ctx context.Context, m ent.Mutation, objectType,
...
+		defer rows.Close()
+
+		if rows.Next() {
+			var exists bool
+			if err := rows.Scan(&exists); err == nil && exists {
+				return nil
+			}
+		}
+
+		return privacy.Denyf("requested object not in organization")
```
golanglemonade: Can you pull this out into a function to keep the complexity of this (already complex) function down?

internal/ent/schema/notification.go
```diff
@@ -132,6 +134,7 @@ func (n Notification) Edges() []ent.Edge {
 			field:      "user_id",
 			immutable:  true,
 			annotations: []schema.Annotation{
+				accessmap.EdgeViewCheck(User{}.Name()),
```
golanglemonade: Shouldn't need this; theres no external resolvers to create notifications (see skip on next line)

internal/ent/schema/notificationpreference.go
```diff
@@ -156,7 +157,7 @@ func (n NotificationPreference) Edges() []ent.Edge {
 			required:   true,
 			immutable:  true,
 			annotations: []schema.Annotation{
-				accessmap.EdgeNoAuthCheck(),
```
golanglemonade, line 159: This shouldn't be needed; This is the user's notification preferences? Isn't it automatically set and not changeable?

internal/ent/schema/notification.go
adelowo: removed it now but just wanted to have the edge checks still

internal/ent/privacy/rule/orgowner.go
adelowo: fixed

golanglemonade (approved):

PR #2603, opened 2026-07-15
Fix openAPI spec generation and completeness, update type names + examples consistently throughout 

matoszz:
The published OpenAPI spec is now derived automatically from handler source code; request bodies, response shapes, status codes, and parameters are discovered from what each handler actually binds and writes. The prior version naively assumed that a "registration context" could get you all possible responses a handler would emit, but took no account of things like 429 middleware rate limit, or code-branches only reachable based on presence (or not) of db records that were never present during registration.

- The documented API surface is now opt-in per route
- Served spec is pre-generated versioned artifact; gutted all runtime behavior
- Models declare their own named examples inline
- All error responses reference one shared error schema
- BREAKING CHANGE: API response types now follow a single *Response naming convention (previously a Reply/Response mix), making the spec schemas and Go client surface uniform


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/1077

### Changes Preview:

- 🔄 Merged Helm values.yaml

golanglemonade (dismissed):

golanglemonade (approved):

PR #2604, opened 2026-07-15
feat: domain scan updates

golanglemonade:
-  Moves enrich calls to the initial create so we don't wait for url scanner and then wait again for these, cutting full time in half
- Fixes issue where the poll would fail on unmarshal of payload when some fields were nil; adds logger to validate to make debugging easier
- Adds system owned option for the scan schema
- Adds rate limiting, internal definition, payment required to the integrations runtime handler
- Pulls in vendor for domain, will allow for future release to map back to a system as well 
- Changes default status of scan to `pending`
- Adds an import job, allow the UI to send back a single import request and have all the data created via the backend, this will allow reuse of the linking of objects in the future, more flexibility, and quick response back to the user as the data is imported. 
- New vendor enrich package that attempts to enrich vendor data with system owned entities 
- Most of the new files in pkg/domainscan are just broken up functions that were previously in `report.go`
- fixes cve with x/image
- pulls in @matoszz change to restructure dispatch request, add runtime services interface, collapse reconcile and scheduled listener structures, init system integration definition


internal/vendorenrich/enrich.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+
+		candidates = append(candidates, candidate{vendor: vendor, name: name, domain: domain})
+	}
+
+	return candidates
+}
+
+// vendorDomain returns the lowercased hostname parsed from vendor's url
+func vendorDomain(vendor map[string]any) string {
```
matoszz, line 90: `pkg/domain` -> `NormalizeHostname`

internal/vendorenrich/enrich.go
```diff
@@ -0,0 +1,260 @@
...
+
+	"github.com/theopenlane/core/internal/ent/generated"
+	"github.com/theopenlane/core/internal/ent/generated/entity"
+	"github.com/theopenlane/core/internal/ent/generated/entitytype"
+	"github.com/theopenlane/core/internal/ent/generated/predicate"
+	"github.com/theopenlane/core/pkg/jsonx"
+)
+
+// systemVendorEntityType is the EntityType.Name that marks an Entity as a vendor
+const systemVendorEntityType = "vendor"
```
matoszz, line 20: is this not an enum?

golanglemonade: No; its another table. We did this before we setup custom enums :(

matoszz (approved):

PR #2606, opened 2026-07-16
Onboarding questionnaire API

adelowo:
onboarding steps are now defined in a `yaml` file. I have incorporated the logic from the catalog system into this too.
hence, there is a new task command `task onboarding:all` to generate the files as needed

Added rest apis for fetching the questions so the frontend can dynamically build the onboarding UI

Merged https://github.com/theopenlane/core/pull/2616 into this

Also added observation period edges for programs


common/models/onboarding.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type Step struct {
+	Key            string               `json:"key" yaml:"key"`
+	Title          string               `json:"title" yaml:"title"`
+	Description    string               `json:"description,omitempty" yaml:"description,omitempty"`
+	Order          int                  `json:"order" yaml:"order"`
+	Hidden         bool                 `json:"hidden" yaml:"hidden"`
+	Questions      []Question           `json:"questions" yaml:"questions"`
+	Modules        []Module             `json:"modules,omitempty" yaml:"modules,omitempty"`
```
golanglemonade: This doesn't follow my latest mock; This should be a more generic `Cards` and doesn't need to be specific to modules

db/migrations-goose-postgres/20260720220333_onboarding_history.sql
```diff
@@ -0,0 +1,7 @@
+-- +goose Up
```
golanglemonade: Can you squash your two different migrations into one?

internal/ent/schema/task.go
```diff
@@ -143,6 +146,32 @@ func (Task) Fields() []ent.Field {
 				entgql.OrderField("is_template"),
 			).
 			Default(false),
```
golanglemonade, line 148: Can you add the metadata field so we can inlcude links/docs like we discussed in slack?

internal/onboarding/genyaml/yamlgen.go
```diff
@@ -0,0 +1,402 @@
...
+	"sort"
+
+	"github.com/dave/jennifer/jen"
+	yaml "github.com/goccy/go-yaml"
+	"github.com/urfave/cli/v3"
+)
+
+const modelsPkg = "github.com/theopenlane/core/common/models"
+
+type questionnaire struct {
```
golanglemonade: Why is this defined here and in models?

internal/onboarding/onboarding.yaml
```diff
@@ -0,0 +1,219 @@
...
+  - key: trial
+    title: Your free 30-day trial
+    order: 6
+    dynamicModules: true
+    tasks:
+      - key: setup-payment-method
+        title: Set up payment method
+        details: Add a payment method before the trial ends to keep paid modules active.
+        priority: 90
+        availableAfterDays: 21
```
golanglemonade: I dont think we need this; I think we just create it as-is
```suggestion
```

internal/ent/schema/task.go
```diff
@@ -143,6 +146,32 @@ func (Task) Fields() []ent.Field {
...
+			).
+			Default(0),
+		field.Time("available_at").
+			GoType(models.DateTime{}).
+			Comment("the time when the task should become available to users").
+			Annotations(
+				entgql.OrderField("available_at"),
+			).
+			Nillable().
+			Optional(),
```
golanglemonade:
```suggestion
```

I think we can leave this out. We can do a due date, but I don't see much use in not showing the the task

internal/onboarding/catalog.go
```diff
@@ -0,0 +1,137 @@
...
+	options := make([]models.QuestionOption, 0, len(standards))
+	seen := map[string]struct{}{}
+
+	for _, std := range standards {
+		value := std.Framework
+		if value == "" {
+			value = std.ID
+		}
+
+		if _, ok := seen[value]; ok {
```
golanglemonade: Standards are unique? Why do we need to track seen?

internal/onboarding/catalog.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return modules
+}
+
+func getFrameworkOptions(ctx context.Context, client *generated.Client) ([]models.QuestionOption, error) {
+	if client == nil {
+		return nil, nil
+	}
+
+	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
```
golanglemonade: Users have access to standards;  this shouldn't need an allowCtx

internal/onboarding/catalog.go
```diff
@@ -0,0 +1,137 @@
...
+			if question.DynamicOptions {
+				question.Options = opts
+			}
+		}
+	}
+
+	return questionnaire, nil
+}
+
+func getModules(client *generated.Client) []models.Module {
```
golanglemonade: Can you use the existing logic in `filterCatalog`?

internal/ent/schema/standard.go
```diff
@@ -123,6 +123,12 @@ func (Standard) Fields() []ent.Field {
 			).
 			Optional().
 			Comment("type of the standard - cybersecurity, healthcare , financial, etc."),
+		field.Int("priority").
```
golanglemonade: What is this for ?

golanglemonade (commented):
initial comments; I haven't looked closely at the listener/hook code yet though

internal/onboarding/genyaml/yamlgen.go
adelowo: oh this is different, this is the cli like we have in catalog ( https://github.com/theopenlane/core/blob/0a9343dca10dbee77f2e7dd397bc76e3ead07a85/cmd/catalog/main.go `

so `task onboarding:all` runs this

internal/ent/schema/standard.go
adelowo: to order the standards when listing so e.g soc2 can come ahead of others else we will need to use something generic like `created_at DESC` or similar

internal/ent/schema/task.go
adelowo: not sure why but I have metadata here already https://github.com/theopenlane/core/pull/2606/changes/BASE..0a9343dca10dbee77f2e7dd397bc76e3ead07a85#diff-1f53eb48c46cc2771d672c613934abaefa8aa5fae985d1063cf706de706bb3f0R95

internal/onboarding/genyaml/yamlgen.go
adelowo: ha seen. I had initially built the cli first before redefining models. will merge both

internal/ent/schema/standard.go
golanglemonade: hmm, I dont think we need that; we can just so it based on a map when pull the data?

adelowo:
merged into https://github.com/theopenlane/core/pull/2621. closing

PR #2621, opened 2026-07-22
feat: onboarding questions + task creation

golanglemonade:
- Adds onboarding questions -> feeds into onboarding task createion
- Task Rules for field and schema level suggseted tasks
- Fixes bug with org level notifications only showing up on backfill
- Checks for runtime before attempting to continuously create scans -> fail

Depends on: https://github.com/theopenlane/entx/pull/290 ✅ 
Replaces: https://github.com/theopenlane/core/pull/2606 (this branch was pointing at that branch but I keep getting merge conflicts)
UI Changes: https://github.com/theopenlane/openlane-ui/pull/2057

Onboarding tasks: 
image

Suggested task from onboarding: 
image


internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,214 @@
...
+
+      Generate policy drafts using Openlane AI based on your organization and compliance framework, then review and edit them so they accurately reflect how your organization operates.
+
+      [Console](https://console.theopenlane.io/policies)
+
+      ## Other Ways to Manage Policies
+
+      ### Connect Google Workspace or OneDrive
+
+      As your program grows, you can synchronize policies from Google Drive or OneDrive to keep Openlane aligned with your document library.
```
LemonAid711: I think I'd say "If you prefer" or "Already have your policies tracked elsewhere?" -just so it doesn't sound like they'll "grow out" of Openlane at some maturity point.

LemonAid711 (changes requested):
Looks good overall -one small working comment I think we should rephrase.

LemonAid711 (commented):
Looks great!

LemonAid711 (dismissed):

adelowo (dismissed):

common/models/onboarding.go
```diff
@@ -0,0 +1,62 @@
+package models
+
+type InputType string
+
+const (
```
matoszz, line 6: All these exported types need comments

common/openapi/models.go
```diff
@@ -2605,6 +2605,16 @@ type ProductCatalogResponse struct {
 	models.Catalog
 }
 
+// OnboardingQuestionsReply holds the fields sent on a response to the onboarding questions endpoint.
+type OnboardingQuestionsReply struct {
```
matoszz: Can you update to use `Response` so that the type is consistent with all our other Request/Response types

internal/ent/taskrules/notification.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+
+	"github.com/theopenlane/entx"
+
+	"github.com/theopenlane/core/common/enums"
+)
+
+// Notification task rule IDs
+const (
+	RuleReviewDomainScan = "review-domain-scan"
```
matoszz: feels a little strange to define a constant for something you use once?

internal/ent/taskrules/templates.go
```diff
@@ -0,0 +1,104 @@
...
+
+//go:embed templates/*.yaml
+var templatesFS embed.FS
+
+// DocsBaseURL is prepended to any metadata.docsLink value that starts with "/", so task
+// definitions can use relative paths (e.g. "/compliance/controls") instead of full URLs
+const DocsBaseURL = "https://docs.theopenlane.io"
+
+// Template is the rendered content for one task rule
+type Template struct {
```
matoszz, line 14: fields need comments

internal/ent/taskrules/templates.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+//go:embed templates/*.yaml
+var templatesFS embed.FS
+
+// DocsBaseURL is prepended to any metadata.docsLink value that starts with "/", so task
+// definitions can use relative paths (e.g. "/compliance/controls") instead of full URLs
+const DocsBaseURL = "https://docs.theopenlane.io"
```
matoszz: we've already got this defined in other locations in our config, would be ideal to re-use

internal/ent/taskrules/templates.go
```diff
@@ -0,0 +1,104 @@
...
+	Details      string         `yaml:"details"`
+	Priority     int            `yaml:"priority"`
+	TaskKindName string         `yaml:"taskKindName,omitempty"`
+	Source       Source         `yaml:"source,omitempty"`
+	Metadata     map[string]any `yaml:"metadata,omitempty"`
+}
+
+// validSources are the Source values allowed in a template YAML file; empty defaults to
+// SourceRecommendations at lookup time, so it's valid here too
+var validSources = map[Source]bool{
```
matoszz: why bother with this at all? Doesn't defining in this way mean you can't declaratively introduce new types?

internal/ent/taskrules/templates.go
```diff
@@ -0,0 +1,104 @@
...
+
+type templateFile struct {
+	Rules map[string]Template `yaml:"rules"`
+}
+
+var templates = mustLoadTemplates()
+
+const (
+	dirName = "templates"
+)
```
matoszz, line 40: embed FS directive already takes care of ensuring the directory that you specify for the embeds is already there, so not sure what purpose these are serving

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
+rules:
+  framework:
+    title: Set up your {{.label}} program
+    taskKindName: Operational
+    details: |
+      Kick your program off by creating a program with the framework controls
+    priority: 10
+    metadata:
+      docsLink: /docs/platform/compliance-management/programs/overview
+      link: '{{if eq .value "soc2"}}/programs/create/soc2?onboarding=true{{else}}/programs/create/framework-based?onboarding=true&framework={{.label}}{{end}}{{if .auditor_name}}&auditorName={{.auditor_name}}{{end}}{{if .auditor_email}}&auditorEmail={{.auditor_email}}{{end}}'
```
matoszz: the naming here is a little confusing

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+      Kick your program off by creating a program with the controls you need
+    priority: 10
+    metadata:
+      docsLink: /docs/platform/compliance-management/programs/overview
+      link: /programs/create?onboarding=true
+
+  import-existing-controls:
+    title: Import existing controls
+    taskKindName: Control Implementation
+    details: |
```
matoszz, line 25: this reads a little weird and is duplicating content we already have in another location

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+    priority: 50
+    metadata:
+      docsLink: /docs/platform/compliance-management/controls/import
+      references:
+        - name: "Getting started: From Gaps to Controls"
+          url: https://docs.theopenlane.io/docs/platform/grc-fundamentals/audit/gapanalysis#from-gaps-to-controls
+        - name: "Best practices: Writing Good Control Implementations"
+          url: https://docs.theopenlane.io/docs/platform/compliance-management/controls/implementations#writing-a-good-implementation
+        - name: Import Fields for Controls
+          url: https://docs.theopenlane.io/docs/platform/compliance-management/controls/import#framework-control-csv-fields
```
matoszz, line 36: why do we have a field for docs links with a concat URL from a base, and then a totally different references type that uses full URLs?

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+  #     # TODO: this is not a real link, this needs to be implemented before this is used
+  #     link: /compliance/controls/templates
+  #     docsLink: /docs/platform/compliance-management/controls/writing-controls
+
+  # Using until template control work is complete
+  import-template-controls:
+    title: Writing your controls
+    taskKindName: Control Implementation
+    details: |
+      Controls define how you actually meet requirements, grounded in the policies you wrote. Every control should trace back to a policy that requires the activity, and every operational obligation in a policy should either map to a control or be a deliberate governance-only statement.
```
matoszz, line 58: why is this stating a near identical verbiage as above?

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+          url: https://docs.theopenlane.io/docs/platform/compliance-management/controls/writing-controls
+        - name: "Best practices: Writing Control Implementations"
+          url: https://docs.theopenlane.io/docs/platform/compliance-management/controls/implementations#writing-a-good-implementation
+        - name: "Blog: Confusing Frameworks with Requirements"
+          url: https://www.theopenlane.io/blog/compliance-tools-frameworks-vs-requirements
+
+  import-existing-policies:
+    title: Import existing policies
+    taskKindName: Policy Review
+    details: |
```
matoszz, line 79: why are we using markdown in here and jumping to h3 -> h5 ?

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+
+      #### Upload from Console
+
+      Upload existing policies in Markdown or Microsoft Word format in Console(https://console.theopenlane.io/policies)
+
+      #### Connect With Integrations
+
+      Sync policies directly from our supported [integrations](https://console.theopenlane.io/automation/integrations?tags=document) such as Google Drive or OneDrive to keep Openlane aligned with your existing documentation.
+
+      #### Sync from GitHub
```
matoszz, line 92: why are you linking directly in some of the text, and not others?

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+
+      ### Create policies in Openlane
+
+      Choose the approach that best fits your organization.
+
+      #### Start with Policy Hub
+
+      Browse Openlane's library of open-source policy templates and customize them to reflect how your organization operates.
+
+      [Open Policies](https://console.theopenlane.io/policies)
```
matoszz, line 120: same question as above; why are we linking URLs inline in some instances and not others?

internal/ent/taskrules/templates/onboarding.yaml
```diff
@@ -0,0 +1,220 @@
...
+      When you're ready for your audit, invite your auditor to Openlane so they can review evidence, request additional documentation, and collaborate with your team throughout the audit.
+
+      ### Next Steps
+
+      #### Invite your auditor
+
+      Invite your auditor as a member of your Openlane organization with the Auditor role.
+
+      - [Console](https://console.theopenlane.io/settings/members)
+      - [Documentation](https://docs.theopenlane.io/docs/platform/organizations/members)
```
matoszz: same questions as above; why are we linking to docs inline with full URLs in some instances and using docs links / references in others

internal/httpserve/handlers/productcatalog.go
```diff
@@ -35,55 +26,5 @@ func (h *Handler) ProductCatalogHandler(ctx echo.Context) error {
...
-	case PublicAudience:
-		return true
-	case BetaAudience:
-		return includeBeta
-	case PrivateAudience:
-		return includePrivate
-	default:
-		return false
-	}
+	return catalog.FilterByAudience(in.IncludeBeta, in.IncludePrivate, gencatalog.GetDefaultCatalog(h.DBClient.EntConfig.Modules.UseSandbox))
```
matoszz, line 29: why is this changing to always use the sandbox catalog instead of having the conditional check

internal/httpserve/route/onboarding.go
```diff
@@ -0,0 +1,23 @@
+package route
+
+import (
+	"net/http"
+
+	"github.com/theopenlane/core/internal/httpserve/handlers"
+)
+
+func registerOnboardingQuestionsHandler(router *Router) error {
```
matoszz, line 9: we're re-reading these files from disk every time???

internal/httpserve/handlers/onboarding_questions.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	echo "github.com/theopenlane/echox"
+	"github.com/theopenlane/utils/rout"
+
+	models "github.com/theopenlane/core/common/openapi"
+	"github.com/theopenlane/core/internal/onboarding"
+	"github.com/theopenlane/core/pkg/logx"
+)
+
+func (h *Handler) ListOnboardingQuestions(ctx echo.Context) error {
```
matoszz, line 12: why is this not binding to the request struct like the other handlers we have? should also have rate limiting middleware

internal/integrations/definitions/cloudflare/domain_scan_import.go
```diff
@@ -253,150 +303,328 @@ func domainScanAssetVendorNames(ctx context.Context, client *generated.Client, s
...
+// of IDs it created
+func findOrCreateDomainScanAssets(ctx context.Context, client *generated.Client, ownerID string, assets []DomainScanImportAsset, scanIDs []string) (map[string]string, []string, error) {
 	ids := make(map[string]string, len(assets))
 
-	for _, asset := range assets {
+	var createdIDs []string
+
+	for _, a := range assets {
+		existing, err := client.Asset.Query().
+			Where(asset.NameEqualFold(a.Name)).
```
matoszz, line 334: doesn't this need to be filtered by owner?

internal/integrations/definitions/cloudflare/domain_scan_import.go
```diff
@@ -253,150 +303,328 @@ func domainScanAssetVendorNames(ctx context.Context, client *generated.Client, s
...
 
-	for _, platform := range platforms {
+	var createdIDs []string
+
+	for _, p := range platforms {
+		entityIDs := resolveDomainScanRefs(p.EntityRefs, entityIDByRef)
+		assetIDs := resolveDomainScanRefs(p.AssetRefs, assetIDByRef)
+
+		existing, err := client.Platform.Query().
+			Where(platform.NameEqualFold(p.Name)).
```
matoszz, line 398: doesn't this need to be filtered by owner?

internal/integrations/runtime/runtime.go
```diff
@@ -144,6 +144,13 @@ func (r *Runtime) Catalog() []types.DefinitionSpec {
 	return r.Registry().Catalog()
 }
 
+// RuntimeClientConfigured reports whether a pre-built runtime client exists for the definition
+func (r *Runtime) RuntimeClientConfigured(definitionID string) bool {
```
matoszz: This shouldn't be needed. It's already handled in another location and dependency injection does it twice in addition to this

internal/onboarding/catalog.go
```diff
@@ -0,0 +1,137 @@
...
+// frameworkOrder pins the display order of these frameworks ahead of the rest
+var frameworkOrder = map[string]int{
+	"SOC 2":     0, //nolint:mnd
+	"ISO 42001": 1, //nolint:mnd
+	"ISO 27001": 2, //nolint:mnd
+	"HIPAA":     3, //nolint:mnd
+	"PCI DSS":   4, //nolint:mnd
+}
+
+func Catalog(ctx context.Context, client *generated.Client) (models.Questionnaire, error) {
```
matoszz, line 25: this and multiple other functions in this file need comments

internal/onboarding/Taskfile.yaml
```diff
@@ -0,0 +1,17 @@
+version: "3"
+
+tasks:
+  default:
+    silent: false
+    cmds:
+      - task --list
+
+  genyaml:
```
matoszz, line 9: we really need a dedicated task file for this? and don't we already have one that does this call?

internal/onboarding/genyaml/yamlgen.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func getInputType(typ models.InputType) string {
+	switch typ {
+	case models.InputTypeString:
```
matoszz: we really defined the constants in the model to re-import them and use as a hand typed string value in the return? Why bother with the constants if you don't even return them as values

internal/ent/taskrules/templates.go
golanglemonade: This comment feels like its not on the right line so I'm not sure what you are asking. its showing up under the const dirName def?

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: What naming?

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: because of how this is showing up in the UI. There might be one overal docs link; but long tasks might have multiple references in addition to the main page

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: They are similar tasks, but one is for importing; the other is for writing new controls; they are not both created

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: Because of font size rendering in console

internal/ent/taskrules/templates/onboarding.yaml
matoszz: whats the difference between "link" and "docsLink"

internal/httpserve/route/onboarding.go
golanglemonade: we aren't?  the questionnaire and catalog are generated compiled into the binary, so the handler just copies already-parsed structs. YAML is only read at go generate time. The one per-request cost is the Standard DB query for the dynamic framework list

internal/httpserve/handlers/onboarding_questions.go
golanglemonade: this is an authenticated route; it would work like any other REST endpoint - why specific rate limiting here?

internal/integrations/definitions/cloudflare/domain_scan_import.go
golanglemonade: No, the context has the owner and so the org mixin handles this.

internal/integrations/definitions/cloudflare/domain_scan_import.go
golanglemonade: No, the context has the owner and so the org mixin handles this.

internal/integrations/runtime/runtime.go
golanglemonade: I was getting resubmitted scans because I didnt have the runtime client configured....

internal/onboarding/Taskfile.yaml
golanglemonade: I think you might be confusing this with the product catalog?

internal/ent/taskrules/templates/onboarding.yaml
matoszz: is there somewhere I'm missing that would indicate these share some kind of relationship? I feel like I would have a lot of difficulty extending or adding anything to this setup the way its laid out

internal/ent/taskrules/templates.go
matoszz: its in reference to the 3 lines of code above (dir name is already enforced, the files in the dir have to be loaded by the directive)

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: anything that has a link does not open the actual test; they have intentionally short descriptions and bring the user to the location in console. Ones with more paths/options have them inline so the task opens -> user can read -> go to the place

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: missed updating this one

internal/httpserve/handlers/productcatalog.go
golanglemonade: it's not? Its using the config value

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: link -> don't open the task, bring user to that location
docsLink -> shows in either case; as Docs/View Docs button in console

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: The `keys`, relate back to the CEL expressions. I'll add a readme

internal/httpserve/handlers/onboarding_questions.go
matoszz: Well... exactly.... hence why you should bind to the request struct (like every other handler does) and apply proper middleware. Do you not think that someone can sign up for a trial and DDoS our authenticated endpoints?

internal/ent/taskrules/templates.go
golanglemonade: mustLoadTemplates checks more than just the dir exists, I reduced the function to skip the err check.

internal/ent/taskrules/templates/onboarding.yaml
golanglemonade: added README

internal/httpserve/handlers/onboarding_questions.go
golanglemonade: fixing this; missed this in reviewing Lanre's original PR.

internal/integrations/runtime/runtime.go
matoszz: Why would you not put the guard in your operation, then?

matoszz: like you're adding a totally new method on Runtime that just calls `_, ok := r.Registry().RuntimeClient(definitionID)` - so you already have the available method and don't need to define a new one

internal/integrations/types/runtime_services.go
```diff
@@ -16,6 +16,8 @@ type RuntimeServices interface {
 	Gala() *gala.Gala
 	// ExecuteRuntimeOperation runs one system-initiated operation inline with no installation or run tracking
 	ExecuteRuntimeOperation(ctx context.Context, definitionID, operationName string, config json.RawMessage) (json.RawMessage, error)
+	// RuntimeClientConfigured reports whether a pre-built runtime client exists for the definition
+	RuntimeClientConfigured(definitionID string) bool
```
matoszz: Runtime is global; individual integration definitions having a runtime instantiation does not belong here

internal/integrations/runtime/runtime.go
golanglemonade: removed

internal/integrations/types/runtime_services.go
golanglemonade: removed

matoszz (approved):

PR #2627, opened 2026-07-23
Send notification to auditor when a program ready

adelowo:


internal/ent/notifications/program.go
```diff
@@ -0,0 +1,199 @@
...
+		Scan(allowCtx, &userIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	return userIDs, nil
+}
+
+func isProgramReady(payload eventqueue.MutationGalaPayload, props map[string]string) bool {
+	status, ok := eventqueue.MutationValue(payload, program.FieldStatus)
```
golanglemonade, line 154: I think you need to check `MutationFieldChanged` not just if the value is there. Otherwise if an update include the field (many times the UI will send all fields), the user could end up with a bunch of notifications.

golanglemonade (approved):

PR #2640, opened 2026-07-24
DirectorySync fixes, backfill / repair mangled data, log key name collision repairs

matoszz:
- Directory sync memberships no longer fail when GitHub's numeric IDs exceed one million - integer IDs now map to their plain decimal form instead of scientific notation
- Reinstalling an integration no longer breaks membership resolution; directory accounts carried over from a prior installation are matched by owner and directory instance rather than the current installation alone
- Existing records stored with corrupted scientific-notation keys are updated through backfill and a defensive check, so syncs already in flight adopt and repair them in place, and the startup backfill cleans up whatever no sync ever touches, so the fix can ship without ordering concerns or duplicate records
- Fixed production log name key collision where the integration operation name and the entity mutation verb no longer fight over the same structured log key, eliminating garbled values like "createorySync"
- Added integration tests demonstrating the externalID key update (given the field is immutable, needed to confirm the behavior exactly)

### DIFF
--- a/internal/httpserve/serveropts/backfill.go
+++ b/internal/httpserve/serveropts/backfill.go
@@ -1,31 +1,31 @@
 package serveropts
 
 import (
-	"bytes"
 	"context"
+	"math"
+	"strconv"
 
+	"entgo.io/ent/dialect/sql"
 	"github.com/rs/zerolog/log"
-	"github.com/stoewer/go-strcase"
 
 	"github.com/theopenlane/iam/auth"
 
-	"github.com/theopenlane/core/common/enums"
 	ent "github.com/theopenlane/core/internal/ent/generated"
-	"github.com/theopenlane/core/internal/ent/generated/file"
-	"github.com/theopenlane/core/internal/ent/generated/organization"
-	"github.com/theopenlane/core/internal/ent/generated/orgmembership"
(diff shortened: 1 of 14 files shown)


internal/httpserve/serveropts/backfill.go
```diff
@@ -1,31 +1,31 @@
...
 // backfillBypassCaps lets the backfill write organizations and memberships without a request caller while
 // skipping the org-filter, FGA, and managed-group guards the membership hooks would otherwise apply
 const backfillBypassCaps = auth.CapBypassOrgFilter | auth.CapBypassFGA | auth.CapInternalOperation | auth.CapBypassManagedGroup
 
+// maxExactExternalID is the largest float64 that can still hold every integer exactly (2^53)
+const maxExactExternalID = float64(1 << 53)
+
 // WithBackfill runs one-time, idempotent startup backfills for data introduced by recent migrations:
-// organization slug names, owner SSO exemptions, trust center update templates, and file md5 hashes.
+// organization slug names, owner SSO exemptions, trust center update templates, file md5 hashes, and
```
golanglemonade: should update the comment if you are removing all the old ones

golanglemonade (dismissed):

golanglemonade (approved):

PR #2644, opened 2026-07-28
Allow 2fa enforcement at the user level

adelowo:
there are 2 ways to enable 2fa now:

- org wide
- or specific user alone

this mimics the sso exemption flow


pkg/ssoutils/enforcement.go
```diff
@@ -53,13 +53,11 @@ type Decision struct {
 	TFARequired bool
 }
 
-// Evaluate resolves how a subject must authenticate against an organization. SSO exemption only
-// affects whether the subject is routed through the directory; multifactor enforcement applies
-// regardless of exemption
+// Evaluate resolves how a subject must authenticate against an organization. when a user is explicitly
+// exempted from SSO, we must require 2fa for that member.
```
golanglemonade: This comment isn't correct; the point of this issue was to allow 2fa enforcement when sso exempt but that would be an org level setting- not always required by default.

internal/ent/hooks/auth.go
```diff
@@ -13,34 +13,50 @@ import (
 	"github.com/theopenlane/core/internal/ent/generated/hook"
```
golanglemonade, line 13: the file should be renamed to be auth, and not sso specific anymore

internal/httpserve/handlers/userinfo.go
```diff
@@ -33,5 +33,28 @@ func (h *Handler) UserInfo(ctx echo.Context) error {
 		return h.BadRequest(ctx, err)
 	}
 
+	if user.Edges.Setting != nil {
+		orgID := caller.OrganizationID
+		allowCtx := privacy.DecisionContext(reqCtx, privacy.Allow)
```
golanglemonade: Do you know why this needs allowCtx? Shouldn't the user be able to know their own default org?

golanglemonade (approved):

PR #2654, opened 2026-08-03
README re-write

matoszz:
In comparing our README to other, larger, open-core projects there were a couple of themes I wanted to borrow from, most notably that the repositories main readme acted more as a bridge to various documentation locations rather than containing detailed technical documentation itself, and typically had either a marketing image or a screen shot of the product itself (e.g. a screen cap of the console). I'd been feeling like ours was getting a bit out of date, and also noticed that the go report card badge is officially retired so thought it would be a good juncture for a refresher.

Open to feedback on if a different image would be preferred, but stuck with something generic.


README.md
```diff
@@ -2,272 +2,80 @@
...
-   test user by running:
-
-   ```bash
-   task cli:user:all
-   ```
+</div>
 
-1. Once this command has finished ^, you can login and perform actions as user
-   `mitb@theopenlane.io` with password `mattisthebest1234`
+[Openlane](https://www.theopenlane.io) is one of the few, truly open-source compliance automation platforms - giving you a system of record for your compliance program including the people, systems, and vendors in scope; the policies and controls that govern them; and the evidence that proves it, and all with the automation to keep all of it current. This repository contains the core server and orchestration services behind the Openlane cloud service.
```
golanglemonade: > and all with the automation to keep all of it current.

`all` twice here makes it a bit hard to read

README.md
```diff
@@ -2,272 +2,80 @@
...
 
-### CSRF Tokens
+<div align="center">
+![](assets/openlane_overview.png)
```
golanglemonade, line 17: This isn't rendering; I think you either need full HTML with an `img` tag or remove the `div` for GitHub markdown

golanglemonade (commented):
🚀

golanglemonade (dismissed):

golanglemonade (approved):

PR #2657, opened 2026-08-03
Support virtual user should be able to view notifications

adelowo:
the exports were being created but there were 2 issues:

- gala handler for notification was skipping because the virtual user does not exists in the db 
https://github.com/theopenlane/core/blob/5bd385099df6f1d9f533f5f5e01d85c18029b5fc/internal/ent/notifications/exports.go#L197-L200

- notifications were not also delivering via websockets since the websocket route does not know about _a virtual user_

Also dropped the edge from `notifications.user_id` to `users` table as we need to be able to insert the virtual support user id there too

CleanShot 2026-08-04 at 00 00 28@2x


pkg/middleware/auth/authoptions.go
```diff
@@ -52,6 +52,12 @@ type Options struct {
 	RedisClient *redis.Client
 	// blacklist is consulted by the constructed validator to honor token and user revocation
 	blacklist tokens.TokenBlacklist
+	// supportSubjectID is the configured subject id of the virtual support identity.
+	supportSubjectID string
+	// supportName is the configured display name of the virtual support identity.
+	supportName string
+	// supportEmail is the configured email of the virtual support identity.
+	supportEmail string
```
golanglemonade, line 60: can you make this a defined struct, something like this:
```suggestion
	// supportUserConfig is the configuration for the  support identity
	supportUserConfig *SupportUserConfig
```

```
type SupportUserConfig struct {
	// supportSubjectID is the configured subject id of the virtual support identity
	supportSubjectID string
	// supportName is the configured display name of the virtual support identity
	supportName string
	// supportEmail is the configured email of the virtual support identity
	supportEmail string
}

golanglemonade (approved):

pkg/middleware/auth/authoptions.go
adelowo: will add that in a next pr

PR #2668, opened 2026-08-07
feat: ensure all objects, files, and history are hard deleted on org deletion

golanglemonade:
Now:
- hard deletes all objects, including the organization
- is not gated on subscriptions being enabled
- deletes files from object storage
- deletes history
- cleans up tuples in fga

Tested locally with s3 file and confirmed after org deletion file was also deleted:

```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Key>REDACTED/REDACTED/Screenshot 2026-05-14 at 13.37.46 (1).png</Key>
<RequestId>REDACTED</RequestId>
<HostId>REDACTED</HostId>
</Error>
```

Fix for tuple deletion on hard delete:
```
openfga=# select * from tuple where _user =  'organization:01KZEN1KNB9Q65QQ5D4JYMQ26E';
(0 rows)
```


internal/ent/hooks/contextx/purgehistory.go
```diff
@@ -0,0 +1,26 @@
+package contextx
+
+import "context"
+
+// PurgeHistoryKey is the context key used to opt a cascade delete into purging the
+// history rows of every record it removes
+type PurgeHistoryKey string
```
golanglemonade: I added these here because there was an existing `contextx` here for cleanup, but this doesnt' feel like the right place. Going to look at condensing this and use the `utils/contextx` pkg like we do in other places.

internal/ent/hooks/listeners_organization_cleanup.go
```diff
@@ -0,0 +1,104 @@
...
+	return contextx.WithPurgeHistory(allowCtx)
+}
+
+// newOrganizationCleanupCaller returns the caller the cascade runs as. It needs to reach every
+// record the organization owns regardless of who is deleting it, so it bypasses the organization
+// filter and FGA checks and identifies itself as an internal operation
+func newOrganizationCleanupCaller(orgID string) *auth.Caller {
+	return &auth.Caller{
+		OrganizationID: orgID,
+		Capabilities:   auth.CapBypassOrgFilter | auth.CapBypassFGA | auth.CapInternalOperation,
```
golanglemonade: Need to confirm this is actually needed `CapBypassOrgFilter`; the context should have the org that is being deleted already

internal/ent/hooks/contextx/purgehistory.go
golanglemonade: switch to using the utils package, still feel like its oddly here but will check to see if there is a better place.

internal/ent/hooks/listeners_organization_cleanup.go
golanglemonade: removed; current tests pass. will add more test coverage to confirm.

adelowo (approved):

PR #2674, opened 2026-08-07
Handle extra events for campaign targets

adelowo:
handled extra events properly from resend to track opens and clicks by splitting the paths


internal/integrations/definitions/email/webhook.go
```diff
@@ -224,29 +224,48 @@ func updateCampaignTarget(ctx context.Context, db *ent.Client, targetID, campaig
...
+		metadata["resend_delivered_at"] = event.Data.CreatedAt
+	case resend.EventEmailOpened:
+		metadata["resend_opened_at"] = event.Data.CreatedAt
+	case resend.EventEmailClicked:
+		metadata["resend_clicked_at"] = event.Data.CreatedAt
+	case resend.EventEmailBounced:
+		metadata["resend_bounced_at"] = event.Data.CreatedAt
+	case resend.EventEmailFailed:
+		metadata["resend_failed_at"] = event.Data.CreatedAt
+	}
```
golanglemonade: This wont work because the `resend_event` will be rewritten each time. I think you need to do something like:

```
func afterPeriod(s string) string {
	if _, after, found := strings.Cut(s, "."); found {
		return after
	}
	return s
}
```

```
    eventStr := afterPeriod(event.Type)
    eventData := map[string]string{
          "email_id":  event.Data.EmailID,
           "timestamp": event.Data.CreatedAt
     }
	metadata[eventStr] = eventData
	}
```

That way you end up with

```
{
  "delivered": {
    "email_id": "...",
    "timestamp": "..."
  }
  "clicked": {
    "email_id": "...",
    "timestamp": "..."
  }
}
```

golanglemonade (approved):

PR #2676, opened 2026-08-07
Enrich standard update notification

adelowo:


common/models/semver.go
```diff
@@ -90,6 +90,33 @@ func ToSemverVersion(version *string) (*SemverVersion, error) {
 	return &semver, nil
 }
 
+// DetectSemverBump returns the bump type between two semver revisions.
+func DetectSemverBump(oldRevision, newRevision string) string {
```
golanglemonade: Can you use constants instead of strings?

```
const (
	SemverBumpNone  SemverBump = ""
	SemverBumpPatch SemverBump = "patch"
	SemverBumpMinor SemverBump = "minor"
	SemverBumpMajor SemverBump = "major"
)
```

internal/ent/notifications/standard_update.go
```diff
@@ -156,6 +188,122 @@ func handleStandardMutation(ctx gala.HandlerContext, payload eventqueue.Mutation
...
+		revision := lo.FromPtrOr(c.ReferenceFrameworkRevision, "")
+
+		return revision, revision != ""
+	})
+
+	if len(revisions) == 0 {
+		return ""
+	}
+
+	slices.SortFunc(revisions, func(a, b string) int {
```
golanglemonade: Can the built in sort for semver be used? https://pkg.go.dev/golang.org/x/mod/semver#Sort If not, this should be pulled as its own sort function, should live in semver.

internal/ent/notifications/standard_update.go
```diff
@@ -156,6 +188,122 @@ func handleStandardMutation(ctx gala.HandlerContext, payload eventqueue.Mutation
...
+			return 1
+		}
+
+		return 0
+	})
+
+	return revisions[0]
+}
+
+func isMajorMinorBump(oldRevision, newRevision string) bool {
```
golanglemonade: Should also live in semver model for resuse

golanglemonade (approved):

PR #2693, opened 2026-08-10
Draft of initial OCI integration 

matoszz:
Drafts a basic OCI integration with health check and rudimentary findings sync. Additionally adds a small bash script and a few templates for scaffolding new integrations since when creating this I realized most of the basics are just boilerplate definitions to get the registration and types in place.

### DIFF
--- a/internal/integrations/cli/stub/builder.go.tmpl
+++ b/internal/integrations/cli/stub/builder.go.tmpl
@@ -1,0 +1,71 @@
+package {{ name }}
+
+import (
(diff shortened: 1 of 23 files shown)


internal/integrations/cli/stub/builder.go.tmpl
```diff
@@ -0,0 +1,71 @@
+package {{ name }}
```
golanglemonade, line 1: 😍  thanks!

golanglemonade (approved):

PR #2695, opened 2026-08-11
Dynamic campaign audience

adelowo:
### Dynamic audience

supports user, group, contact, personnel schemas.

```
mutation($input: CreateAudienceInput!) {
  createAudience(input: $input) {
    audience {
      id
      audienceType
    }
  }
}

```

```
  {
    "input": {
      "name": "All Contacts with an existing email",
      "description": "All contacts with an email address",
      "audienceType": "DYNAMIC",
      "filters": {
        "selectors": [
          {
            "schema": "contact",
            "where": "target.email != ''"
          }
        ]
      }
    }
  }
```

```
  {
    "input": {
      "name": "Openlane and Google Contacts",
      "description": "Contacts at Openlane or Google Cloud",
      "audienceType": "DYNAMIC",
      "filters": {
        "selectors": [
          {
            "schema": "contact",
            "expression": "target.company == 'Openlane'"
          },
          {
            "schema": "contact",
            "expression": "target.company == 'Google Cloud'"
          },
          {
            "schema": "user",
            "expression": "target.email != ''"
          }
        ]
      }
    }
  }
```

### Manual audience where they are defined upfront

```

mutation($input: CreateAudienceMemberInput!) {
  createAudienceMember(input: $input) {
    audienceMember {
      id
    }
  }
}

{
  "input": {
    "audienceType" : "MANUAL" or dynamc,
    "name": "Employee group 1",
    "description" : "employees of fictional company"
  }
}
```

- Also improved validation of filters + celx expressions 

CleanShot 2026-09-21 at 23 05 46@2x

CleanShot 2026-09-21 at 22 31 09@2x


fga/model/generated/crud.fga
```diff
@@ -26,6 +26,17 @@ extend type organization
     define asset_creator: [group#member]
```
golanglemonade, line 26: in roles, add `audience` to the `campaign_manager` role

fga/model/automation/assessments.fga
```diff
@@ -124,3 +124,42 @@ type campaign_target
     define parent_editor: can_edit from parent
     define parent_viewer: can_view from parent
     define parent: [campaign]
+
+# audiences are reusable campaign recipient lists, viewable/editable through org campaign permissions
```
golanglemonade, line 128: This looks like its just org, I dont see a parent or campaign
```suggestion
# audiences are reusable campaign recipient lists, viewable/editable through org permissions
```

internal/ent/schema/audience_member.go
```diff
@@ -0,0 +1,192 @@
...
+		field.String("identity_holder_id").
+			Comment("the identity holder associated with this audience member").
+			Optional(),
+		field.String("subscriber_id").
+			Comment("the subscriber associated with this audience member").
+			Optional(),
+		field.String("email").
+			Comment("the email address for this audience member").
+			NotEmpty().
+			Annotations(
```
golanglemonade, line 73: Why doe this have to be not empty? Can't it be from one of the edges?

internal/ent/hooks/audience.go
```diff
@@ -0,0 +1,85 @@
...
+	audienceFilterMissingBulkType      = "bulk audience filter updates must include audience_type"
+	audienceFilterUnsupportedOperation = "bulk dynamic audience updates must include filters"
+	audienceFilterManualTextBulk       = "bulk manual audience updates must clear filters"
+)
+
+// HookAudienceValidateFilters makes sure the filtering requested is supported and valid
+func HookAudienceValidateFilters() ent.Hook {
+	return hook.On(func(next ent.Mutator) ent.Mutator {
+		return hook.AudienceFunc(func(ctx context.Context, m *generated.AudienceMutation) (generated.Value, error) {
+			audienceTyp, err := getAudienceType(ctx, m)
```
golanglemonade, line 31: You need to make sure this isn't a soft delete first and if so, skip the hook

internal/integrations/definitions/email/audience_targets.go
```diff
@@ -0,0 +1,290 @@
...
+type audienceRecipientResolveOptions struct {
+	audienceID   string
+	audienceType enums.AudienceType
+	audience     *generated.Audience
+	db           *generated.Client
+	orgID        string
+	filters      map[string]any
+}
+
+func snapshotCampaignAudiences(ctx context.Context, db *generated.Client, camp *generated.Campaign) error {
```
golanglemonade, line 54: Would it work to combine the trust center subscribers to use this code so we can follow the same pattern?

internal/ent/schema/audience_member.go
adelowo: we can do that but I opted for this for 2 reasons:

- I'd suppose I will want to have a _view_ of those I sent the campaign to later on. if we update the email of an identity_holder/contact for some reason, we may lose the original one we have
- will get more complex where we have to track which schema has the id

internal/audiences/filters.go
```diff
@@ -0,0 +1,315 @@
...
+	errKeyMatchUnsupported           = errors.New("key_match is not supported for audience selectors yet")
+	errSourceSelectorsUnsupported    = errors.New("source selectors are not supported for audience selectors yet")
+	errSelectorSchemaNoExpressions   = errors.New("selector schema does not support expressions")
+)
+
+type filterSet struct {
+	Selectors []entityops.TargetSelector `json:"selectors,omitempty"`
+}
+
+type Recipient struct {
```
golanglemonade, line 38: Why is this redefined, entityOps has all these types? `AudienceMemberProjection` (and you are already importing that package here)

internal/integrations/definitions/email/audience_targets.go
```diff
@@ -0,0 +1,352 @@
...
+
+			return handle(recipients)
+		})
+	default:
+		return fmt.Errorf("%w: %q", errUnsupportedAudienceType, opts.audienceType)
+	}
+}
+
+func resolveTrustCenterSubscriberRecipients(ctx context.Context, db *generated.Client, camp *generated.Campaign, handle campaignRecipientHandlerFunc) error {
+	var lastID string
```
golanglemonade, line 258: I didnt mean "move the function" here; I meant can we define the filter in the same way and reuse the exact logic as a customer defined target?

internal/audiences/filters.go
```diff
@@ -0,0 +1,315 @@
...
+	errSelectorSchemaRequired        = errors.New("selector schema is required")
+	errSchemaNotRegistered           = errors.New("schema is not registered")
+	errUnsupportedRecipientSource    = errors.New("schema cannot be used as an audience recipient source")
+	errKeyMatchUnsupported           = errors.New("key_match is not supported for audience selectors yet")
+	errSourceSelectorsUnsupported    = errors.New("source selectors are not supported for audience selectors yet")
+	errSelectorSchemaNoExpressions   = errors.New("selector schema does not support expressions")
+)
+
+type filterSet struct {
+	Selectors []entityops.TargetSelector `json:"selectors,omitempty"`
```
golanglemonade, line 35: does this need to be exported for some reason?

internal/audiences/filters.go
```diff
@@ -0,0 +1,315 @@
...
+	return eval, nil
+}
+
+func doesSelectorMatch(ctx context.Context, eval *celx.NativeEntityEvaluator, expression string, entity any) (bool, error) {
+	if strings.TrimSpace(expression) == "" {
+		return true, nil
+	}
+
+	data, err := json.Marshal(entity)
+	if err != nil {
```
golanglemonade, line 305: These don't seem specific to a audiences? Should this live in a more generic internal celx / entityops eval package?

internal/audiences/filters.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	return nil
+}
+
+func resolveSelectors(ctx context.Context, db *generated.Client, orgID string, selector entityops.TargetSelector, handle func([]Recipient) error) error {
+	schema, _ := entityops.LookupSchema(selector.Schema.Name)
+
+	switch schema.Snake {
+	case entityops.SchemaContact.Snake:
```
golanglemonade, line 172: why only contact and identity holder? The schema allows for these as well, right? 
- subscribers
- users
- groups

internal/audiences/filters.go
```diff
@@ -0,0 +1,315 @@
...
+		return resolveSelector(ctx, selector, selectorResolveOptions[*generated.Contact]{
+			targetType: reflect.TypeFor[entityops.ContactProjection](),
+			fetchFn: func(lastKnownID string) ([]*generated.Contact, error) {
+				query := db.Contact.Query().
+					Where(
+						contact.OwnerIDEQ(orgID),
```
golanglemonade, line 178: This should be down under the org context, and therefore this isn't needed because its applied by the interceptor

internal/integrations/definitions/email/audience_targets.go
```diff
@@ -0,0 +1,352 @@
...
+	fullName       string
+	contactID      string
+	userID         string
+	groupID        string
+	subscriberID   string
+	source         string
+	audienceID     string
+	sourceObjectID string
+	metadata       map[string]any
+}
```
golanglemonade, line 41: Same here; why are we manually creating this struct again?

internal/integrations/definitions/email/audience_targets.go
```diff
@@ -0,0 +1,352 @@
...
+	audience     *generated.Audience
+	db           *generated.Client
+	orgID        string
+	filters      map[string]any
+}
+
+type campaignRecipientHandlerFunc func([]audienceRecipient) error
+
+func snapshotCampaignAudiences(ctx context.Context, db *generated.Client, camp *generated.Campaign) error {
+	allowCtx := privacy.DecisionContext(ctx, privacy.Allow)
```
golanglemonade, line 55: Can you confirm this is actually needed and if so add a comment about why ?

internal/integrations/definitions/email/audience_targets.go
```diff
@@ -0,0 +1,352 @@
...
+					fullName:       recipient.FullName,
+					contactID:      recipient.ContactID,
+					userID:         recipient.UserID,
+					groupID:        recipient.GroupID,
+					subscriberID:   recipient.SubscriberID,
+					source:         recipient.Source,
+					audienceID:     opts.audienceID,
+					sourceObjectID: recipient.SourceObjectID,
+					metadata:       recipient.Metadata,
+				})
```
golanglemonade, line 247: this is the exact same logic duplicated from the above section on line 206 of just mapping a field to a field

internal/ent/schema/audience_member.go
golanglemonade: We should snapshot the emails it was sent to on the `campaign_taget`, it doesn't need to be set on the audience

PR #2702, opened 2026-08-14
Integration run post-execution summaries, directorysync batching, mutationIDs error surfacing

matoszz:
- Directory sync no longer infers membership removals unless the provider batch is a fully confirmed, complete snapshot, eliminating false removals from incremental, degraded, or partially failed syncs
- Stale or superseded sync runs can no longer finalize or roll back membership state confirmed by a newer run
- Record-level ingest failures are skipped and accounted for instead of failing whole batches
- Integration runs surface richer post-execution summaries, and terminal misconfiguration failures are distinguishable from retryable errors so recurring cycles can stop instead of retrying forever
- Integration lookup during dispatch now enforces owner and definition consistency in one shared resolution path
- Mutation ID resolution failures propagate as errors instead of being silently swallowed by hooks
- Recurrence scheduling math is consolidated into a single calendar-aware implementation shared by campaigns and reviews
- Avatar discovery is now a reusable standalone capability rather than logic embedded in an event listener
- Dead computed-field plumbing and a duplicate hook registration were removed


internal/entdb/client.go
```diff
@@ -203,9 +203,6 @@ func New(ctx context.Context, c entx.Config, jobOpts []riverqueue.Option, client
 	db.Intercept(interceptors.QueryLogger())
 	db.Intercept(BlockInterceptor())
 
-	// adds default hooks for all edge permissions
-	db.Use(hooks.HookEdgePermissions())
```
golanglemonade, line 207: Why was this removed? It is used to enforce edge access

internal/favicon/favicon.go
```diff
@@ -0,0 +1,212 @@
...
+
+	_, _ = io.Copy(io.Discard, resp.Body)
+
+	return resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices
+}
+
+// normalizeURL adds HTTPS when the domain has no scheme.
+// if provided domain by the org during onboarding is theopenlane.io, this
+// converts it to https://theopenlane.io so we can reach it successfully
+func normalizeURL(domain string) string {
```
golanglemonade, line 174: I know this was copied over, so we can do this later but we have 2 functions that do this so we should move this to the `domain/normalize` package probably

internal/entdb/client.go
golanglemonade: its on line 212; this is a dupe

golanglemonade (approved):

PR #2711, opened 2026-08-18
Gala Refactor + Integrations Refactor + Global listener refactor + uplift

matoszz:
The hotspots in this PR which may result in behavioral changes to the system to pay special attention to:

- serveropts job kind and topic "migrations" - this was added to create better control mechanisms to how we group and process events via gala; before in river you would always see `gala_dispatch_v1` (not helpful) but instead of creating an arbitrary grouping, I created a globally unique inventory that `gala` controls and defines via `namespaces`. This requires us to perform a "migration" so that existing scheduled, batched, recurring jobs (e.g. via campaigns, ingest listeners, etc.) get re-scheduled into their new destination namespace and we safely process in-flight jobs until we've "drained" out the queue
- all listener definitions globally were updated to have `Caller`, `Match` (on field), `Gate` (when not to process), `ContextKeys`, `LogFields`, and more - all a result of a global audit to find the most consistently re-used function patterns and the most *abused* methods within them. The intent with these fields is to move logic surrounding when, how, and under what context, the listener is intended to perform work and so that a developer has a clear line of sight into setting them rather than re-creating helpers or copy + pasting parts of existing listeners into theirs for common things like `auth.CallerFromContext`. This resulted in several judgement calls especially when replacing ent queries with entityops `Schema.Load` and other newly introduced helpers and types, so functional verification on them is ideal.
- Integration health and connectivity now moves the integration into a specific status, sends user notifications about its status, cascade-removes all currently scheduled + processing gala / river jobs related to the integration, then on re-establishment of the integration health the jobs are re-scheduled + deployed. This same construct is leaned into for integration deletion (previously didn't remove any gala / river jobs) whether that's a direct deletion (or soft delete), or cascade as a result of integration deletion
- previously `OpUpdate` and `OpDelete` (bulk actions) did not emit events, despite many listeners subscribing to them; these events are now processed (where they were previously silent) because I made the emit event global ent hook "fan out" events so that the bulk operations end up creating individual mutation payloads. This can / will cause some behavioral quirks because prior to this PR these listeners received nothing for these event types, but after this PR they will start receiving them. I audited all of the existing listeners and tried to ensure any processing that was happening was intentional but special attention should be given to this fact
- today we "fan out" process directorysyncs because the original thought was that we should create a job for every individual directory record as to not inadvertently show membership addition or removal as a result of failure to sync the record or batch; this has turned out to be a scalability eye sore, so the processing was moved to a batch and logic was updated to account for flags / markers to use as indicators for a complete sync
- "Invocation" signals were added to recurring events as well as Cancellation criteria so that we don't end up with events + jobs that are being processed indefinitely without having cancellation criteria attached
- Effort was applied to materially consolidate all the variable "Listener" types that we had ended up with (about ~5 different types by my count) which involved collapsing various types and fields into the global `Definition`; the pkg/gala + integration API surface was narrowed to disallow ability to externally perform construction of a new registry or things like the manual creation of an integration run. This isn't fully comprehensive but the export surface narrowed as well as the available constructors
- internal/ent/notifications was materially refactored to use a more declarative listener definition and be a larger consumer of the `entityops` registry and its Schema types; functional testing should be done for notifications to ensure existing behavior is intact


internal/ent/notifications/mentions.go
```diff
@@ -64,367 +25,105 @@ type mentionNotificationInput struct {
...
-
-	if fields.evidenceID != "" {
-		evidence, err := client.Evidence.Get(allowCtx, fields.evidenceID)
-		if err != nil {
-			return "", "", "", fmt.Errorf("failed to get evidence: %w", err)
-		}
-		return generated.TypeEvidence, fields.evidenceID, evidence.Name, nil
+// noteParent resolves the parent object type, ID, and display name from the note's
+// loaded edges; a note is associated with at most one parent object
+func noteParent(noteEntity *generated.Note) (string, string, string) {
```
golanglemonade, line 103: Note for myself: I know we already have more edges here, we should make this more dynamic

golanglemonade (approved):

PR #2752, opened 2026-09-01
Extend domain scanning to include brand design profile extraction

adelowo:
~- added a new resolver to trigger domain scanning so the ui can decide to do that if there is a need to extract brand
design data or any other thing really~
- automatically updates the trustcenter settings with the brand design that was extracted from the first domain in onboarding
- allow for trustcenter settings to be updated with brand design extraction from a new domain scan
- added a new `bulkscan` config ( defaults to true so existing behaviour is kept ) but the problem here is bulk scanning is only available to _paid workers/cloudflare users_.. if set to false, will be scanned sequentially instead. defaults to true so bulk scanning is the default

> always wanted to do this and had a stashed change in `core jobs` earlier that I used to test whenever I had to work with domain scanning 

```
  mutation CreateDomainScan($input: CreateScanInput!) {
    createScan(input: $input) {
      scan {
        id
        target
        status
        scanType
      }
    }
  }

  {
    "input": {
      "target": "https://stripe.com",
      "scanType": "DOMAIN",
      "performedBy" : "openlane_domain_scan",
      "metadata": {
        "brandDesignOnly": true // filter to brand design extraction only
       "applyBrandDesign": true // auto apply to trustcenter
      }
    }
  }

```


theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/1166

### Changes Preview:
✅ Updated ConfigMap template

internal/integrations/definitions/cloudflare/domain_scan_import.go
```diff
@@ -173,9 +182,210 @@ func importDomainScanReview(ctx context.Context, client *generated.Client, envel
...
+		if err != nil {
+			return false
+		}
+
+		d, ok := domainscan.RegistrableDomain(hostname)
+		return ok && d == domain
+	})
+
+	if !doesScanMatchDomain {
+		return "", nil
```
golanglemonade: should this return an error? feels like this will silently fail ?

adelowo: not at all. this simply means none of the domains from the scan do not match the custom domain the user set in their trustcenter

common/go.sum
```diff
@@ -148,8 +148,7 @@ github.com/theopenlane/core/v2 v2.3.0 h1:C4BqyW1P/gkEGC+xtFyws6LLTB09U63lVmuBP/5
 github.com/theopenlane/core/v2 v2.3.0/go.mod h1:VkcfF7fg9Qy1KsiRYeFdXKGQd9Ov4BQZz98utBzws9U=
 github.com/theopenlane/echox v0.3.0 h1:uwOKEw+r1utGQoOR6dZQqAVuY5j8TcasqnTwO5+rMsA=
 github.com/theopenlane/echox v0.3.0/go.mod h1:yTrXnj7s3VNIg0FCvB7Dut2Elr+LqJKU/nruxx1E1cM=
-github.com/theopenlane/entx v0.33.4 h1:urRyAx/J5dLplXM73ehxOj+YlGjlS13MI00x5k6kneM=
-github.com/theopenlane/entx v0.33.4/go.mod h1:ubJf45e2Ivax9rq6tDUog6/Rthe3MT9pWB1EHnp9ylA=
+github.com/theopenlane/entx v0.33.5 h1:Ewf6ZtGzIu9jbmaev8w6njJClTJrgQDW3ODQvgtKNT8=
```
golanglemonade, line 151: you need a `go mod tidy` or `go work sync` in here; its missing the second line

internal/graphapi/schema/domainscanextended.graphql
```diff
@@ -125,6 +125,21 @@ input ImportDomainScanReviewFindingInput {
     severity: String
 }
 
+"""
+brand design data gotten from the domain scan
```
golanglemonade, line 129:
```suggestion
brand design data retrieved from the domain scan
```

internal/integrations/definitions/cloudflare/domain_scan_listeners.go
```diff
@@ -247,6 +351,22 @@ func (s domainScanSaga) persistDomainScanEnrichment(ctx context.Context, organiz
...
 
+	if !applyBrandDesign {
+		return nil
+	}
+
+	if enrichment.Branding == nil {
+		s.markDomainScanFailed(ctx, organizationID, internalScanID)
+
+		return ErrDomainScanBrandDesignMissing
+	}
```
golanglemonade, line 362: Don't fail the entire domain scan on missing branding, just add an info log.

internal/integrations/definitions/cloudflare/domain_scan_listeners.go
```diff
@@ -247,6 +351,22 @@ func (s domainScanSaga) persistDomainScanEnrichment(ctx context.Context, organiz
...
+
+	if _, err := applyBrandingToTrustCenter(systemCtx, s.services.DB(), buildBrandDesignImport(enrichment.Branding)); err != nil {
+		s.markDomainScanFailed(ctx, organizationID, internalScanID)
```
golanglemonade, line 365: Same here, don't fail the entire scan as failed based on a partial failure

internal/integrations/definitions/cloudflare/operation_domain_scan_request.go
```diff
@@ -108,6 +158,17 @@ func (d DomainScanRequest) Handle() types.OperationHandler {
 
 		saga := domainScanSaga{services: request.Services}
 
+		if cfg.BrandDesignOnly {
+			if err := saga.runBrandDesignScan(ctx, organizationID, scanRecord.ID, cfg.Domain, cfg.ApplyBrandDesign); err != nil {
```
golanglemonade, line 162: I would just mark the failure here instead of having `s.markDomainScanFailed(ctx, organizationID, scanID)` on every error line in `runBrandDesignScan`.

Also log the error here too.

internal/ent/hooks/listeners_scan_domain.go
```diff
@@ -56,27 +58,42 @@ func handleScanDomainCreated(inv entityops.Invocation, _ entityops.MutationPaylo
...
 
-	for _, domain := range setting.Domains {
+	previousDomains, _ := payload.OldValue(organizationsetting.FieldDomains)
+	domains, ok := previousDomains.([]string)
+
+	// only apply the branding to the trustcenter by default on the first run
+	// if org settings has existing domains then no need to
+	applyBrandDesign := domains == nil || (ok && len(domains) == 0)
+
+	logx.FromContext(inv.Context).Debug().Bool("applyBrandDesign", applyBrandDesign).Int("num_domains", len(domains)).Int("num_setting_domains", len(setting.Domains)).Bool("ok str", ok).Msg("HERE: applyBrandDesign setting")
```
golanglemonade: You pulled in my debugging :) 
```suggestion
```

internal/integrations/definitions/cloudflare/domain_scan_import.go
```diff
@@ -173,9 +181,171 @@ func importDomainScanReview(ctx context.Context, client *generated.Client, envel
...
+	palettes := []string{
+		brandDesign.PrimaryColor,
+		brandDesign.ForegroundColor,
+		brandDesign.BackgroundColor,
+		brandDesign.AccentColor,
+		brandDesign.SecondaryBackgroundColor,
+		brandDesign.SecondaryForegroundColor,
+	}
+
+	// check if this new palette imports have atleast one valid color code
```
golanglemonade:
```suggestion
	// check if this new palette imports have at least one valid color code
```

internal/integrations/definitions/cloudflare/domain_scan_import.go
```diff
@@ -173,9 +181,171 @@ func importDomainScanReview(ctx context.Context, client *generated.Client, envel
...
+	if !isChanged {
+		return false, nil
+	}
+
+	_, err := setting.Update().SetInput(input).Save(ctx)
+	if err != nil {
+		return false, fmt.Errorf("updating %s trust center brand design: %w", setting.Environment, err)
+	}
+
+	return true, nil
```
golanglemonade, line 347: Can you add a unit test for this function? It's pretty complex but easily testable

internal/integrations/definitions/cloudflare/domain_scan_import.go
```diff
@@ -173,9 +181,171 @@ func importDomainScanReview(ctx context.Context, client *generated.Client, envel
...
+func updateTrustcenterBrandDesignSetting(ctx context.Context, setting *generated.TrustCenterSetting, brandDesign DomainScanImportBranding) (bool, error) {
+	input := generated.UpdateTrustCenterSettingInput{}
+	isChanged := false
+	chooseColor := func(newColor, existingColor string) (string, bool) {
+		color := strings.TrimSpace(newColor)
+		if color == "" || validator.HexColorValidator(color) != nil {
+			return existingColor, false
+		}
+
+		return color, color != existingColor
```
golanglemonade: define this function outside, that will keep the function complexity down and make it testable by itself; I would also consider pulling all the "validate color pallette" to its own function as well because theres a lot of logic here, but 75% of it is color validation with other things mixed - making it a bit hard to follow

pkg/domainscan/browserbranding.go
```diff
@@ -0,0 +1,328 @@
...
+	defer cancel()
+
+	target, err := urlx.Parse(domain)
+	if err != nil {
+		return nil, err
+	}
+
+	if target.Scheme != "http" && target.Scheme != "https" {
+		return nil, errBrandingURLScheme
+	}
```
golanglemonade: This is already checked on `urlx.Parse`, shouldn't need this here as well.

pkg/domainscan/browserbranding.go
```diff
@@ -0,0 +1,328 @@
...
+	var pageInfo struct {
+		Title string `json:"title"`
+		URL   string `json:"url"`
+	}
+
+	if err := cdp.evaluatePageData(ctx, brandingPageInfoScript, &pageInfo); err != nil {
+		return nil, err
+	}
+
+	meta := cdp.loadedDocuments[navigation.FrameID]
```
golanglemonade: should check this exists: 
```
meta, ok := cdp.loadedDocuments[navigation.FrameID]
if !ok {
    return nil, errBrandingNoDocument
}
```

pkg/domainscan/browserbranding.go
```diff
@@ -0,0 +1,328 @@
...
+		if result == nil {
+			return nil
+		}
+
+		return json.Unmarshal(message.Result, result)
+	}
+}
+
+func (b *brandDesignCDP) readResponse(ctx context.Context) (*brandDesignCDPMessaging, error) {
+
```
golanglemonade: empty line
```suggestion
```

golanglemonade (approved):

PR #2760, opened 2026-09-03
Add per Integration Operation health checks

matoszz:
- Integrations now verify each operation's actual permissions rather than only credential validity, so misconfigurations like missing Azure admin consent surface at setup instead of at the first failed sync
- A new degraded status isolates a single failing operation; healthy operations keep syncing, the failure carries a user-facing reason, and the integration escalates to errored only when nothing is left working
- Health is now a typed, first-class record on the integration (reason, failing operations, last successful check) exposed through the API, replacing loose metadata keys and prior health-check-as-operation plumbing
- The console badge reflects persisted truth instead of a live credential probe that could show healthy on a stopped integration, with a dedicated assessment endpoint behind the health check button
- Health notifications now reach only organization owners and admins, and the unused deleted status is retired


internal/integrations/runtime/health.go
```diff
@@ -151,12 +471,71 @@ func (r *Runtime) verifyInstallationHealth(ctx context.Context, installation *en
...
+			return false
+		}
+
+		return true
+	})
+}
+
+// notifyIntegrationHealth sends one health notification to the owning organization's owners and admins
+func (r *Runtime) notifyIntegrationHealth(ctx context.Context, installation *ent.Integration, objectType, title, body string, data map[string]any) error {
+	ids, err := notifications.OrgUserIDsByRole(ctx, r.DB(), installation.OwnerID, enums.RoleOwner, enums.RoleAdmin)
```
golanglemonade:
```suggestion
	ids, err := notifications.OrgUserIDsByRole(ctx, r.DB(), installation.OwnerID, enums.RoleOwner, enums.RoleSuperAdmin)
```

I think you want super admin instead or in addition to admin here

golanglemonade (approved):

PR #2762, opened 2026-09-03
Fix comments not showing up on policies/procedures

adelowo:
- added migrations from https://github.com/theopenlane/core/pull/2760
- fixed a bug where 
  - comments actually exists but were not rendered on the UI tree
  - comments may be replaced when adding a new one ( go to policies/procedure , create a new comment - no refresh, then create another one - the original one gets lost )

```
{
    "errors": [
      {
        "extensions": {
          "code": "UNAUTHORIZED",
          "message": "you do not have permission to perform this action, please contact your
          organization owner"
        },
        "message": "you are not authorized to perform this action",
        "path": ["updateProcedure"]
      }
    ]
  }

```

the major problem here is it was an hidden bug all around, sometimes when being sent from the frontend
`details_json` can be empty/null so our previous comparison checks made it seem like it was the actual policy
that was being updated so it was falling down to fga checks 

So the notes were actually stored in the db still ( `InsertProcedureComment/InsertInternalPolicyComment` succeeds ) , just not attached to the policy when doing `updateProcedure` and others with the `detailsJSON` 

so this fixes the comparison and handles two cases when `details_json` is empty or not


internal/ent/privacy/rule/allow_comments.go
```diff
@@ -55,8 +55,25 @@ func CheckIfCommentOnly() privacy.MutationRuleFunc {
...
+				}
+
+				if err := m.SetField(detailsJSONFieldName, mergedComments); err != nil {
+					return privacy.Denyf("unable to merge comment markers")
+				}
+
 				return privacy.Allowf("mutation has only comments added to details_json, allowing")
 			}
+
+			if detailsJSONFieldName == "details_json" && len(oldDetailsTyped) == 0 {
```
golanglemonade: shouldn't this work for `description_json` too ?

adelowo: Actually didn't test that since the bug I was tracking down was comments on details. 

Let me verify that quickly

golanglemonade (approved):

PR #2778, opened 2026-09-10
improve Onboarding Flow for Program Creation

adelowo:
create the actual program + controls instead of just leaving tasks behind asking users to create them


internal/ent/hooks/onboarding.go
```diff
@@ -56,6 +56,11 @@ func HookOnboarding() ent.Hook {
 				}
 			}
 
+			compliance, _ := m.Compliance()
+			if err := createProgram(ctx, m.Client(), org, compliance); err != nil {
+				return nil, err
```
golanglemonade, line 61: do not error onboarding on failed creation of program, we should just log this. Erroring would prevent the rest of this from happening + onboarding would fail/never get slack message.

internal/ent/hooks/onboarding_program.go
```diff
@@ -0,0 +1,138 @@
...
+	caller, ok := auth.CallerFromContext(ctx)
+	if !ok || caller == nil {
+		return auth.ErrNoAuthUser
+	}
+
+	newCaller := *caller
+	newCaller.OrganizationID = org.ID
+
+	ctx = auth.WithCaller(ctx, &newCaller)
+	ctx = privacy.DecisionContext(ctx, privacy.Allow)
```
golanglemonade: Can you confirm why this needs privacy allow?

internal/ent/hooks/onboarding_program.go
```diff
@@ -0,0 +1,138 @@
...
+			labels = append(labels, "Other")
+			continue
+		}
+
+		std, err := client.Standard.Query().
+			Where(
+				standard.FrameworkEQ(framework),
+				standard.StatusEQ(enums.StandardActive),
+				standard.SystemOwned(true),
+				standard.IsPublic(true),
```
golanglemonade: The interceptor should do this and the active filter for you without the privacy.Allow I believe

internal/ent/hooks/onboarding_program.go
```diff
@@ -0,0 +1,138 @@
...
+				standard.FrameworkNotIn("openlane-standard", "openlane-trust-center"),
```
golanglemonade: Why is this needed if there is `standard.FrameworkEQ(framework),` ?

internal/ent/hooks/onboarding.go
```diff
@@ -56,6 +56,11 @@ func HookOnboarding() ent.Hook {
 				}
 			}
 
+			compliance, _ := m.Compliance()
+			if err := createProgram(ctx, m.Client(), org, compliance); err != nil {
```
golanglemonade: This shouldn't happen within the hook to create onboarding - it should happen in a listener that does it async - the clone controls can be slow and we are holding up the rest of this transaction on the creation.

internal/ent/hooks/onboarding_program.go
adelowo: need to always remember not to default to privacy in gala and hooks 😓

internal/ent/hooks/onboarding_program.go
adelowo: ha not needed. bad copy/pasta from catalog frameworks

internal/ent/hooks/listeners_campaign_recurring.go
```diff
@@ -17,9 +17,6 @@ import (
 	"github.com/theopenlane/core/v2/pkg/logx"
 )
 
-// init registers the campaign recurring listeners so gala setup picks them up automatically
-func init() { registerListeners(CampaignRecurringListeners) }
```
golanglemonade, line 21: Why did you delete all these? This was to prevent having to go to another file to remember to register them, and it can just be done at the source instead

internal/ent/hooks/listeners_onboarding_program.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+// OnboardingProgramListeners sets up gala to process onboarding requests such as creating programs
+// and cloning them in the background
+func OnboardingProgramListeners() []gala.Registration {
+	return []gala.Registration{
+		entityops.MutationListener{
+			Schema:     entityops.SchemaOnboarding,
+			Operations: []string{entityops.OpCreate},
+			Handle:     handleOnboardingProgram,
```
golanglemonade: This was added fairly recently, look at the `Caller` setting, e.g.:

```
			Caller: func(restored *auth.Caller, _ entityops.MutationPayload) *auth.Caller {
				return restored.WithCapabilities(auth.CapInternalOperation)
			},
```

This pattern should be used over adding things like `privacy.Allow`; Might even want to move all the setting of the caller in here with the org id; that way you shouldn't even need to bypass internal operations, because looking at the code its just to get the onboarding so something like:

1. Set a system context to get the onboarding organization
2. Set that in the caller and return the restored caller with that 
This will keep the handleOnboardingProgram just to the actual program creation and not setting up caller information

internal/ent/hooks/listeners_onboarding_program.go
```diff
@@ -0,0 +1,197 @@
...
+		},
+	}
+}
+
+func handleOnboardingProgram(inv entityops.Invocation, _ entityops.MutationPayload) error {
```
golanglemonade, line 40: isn't the payload here the onboarding? Can't we use that instead of doing a get?

internal/ent/hooks/listeners_campaign_recurring.go
adelowo: Wanted to move all to single point of registration as the inits were everywhere. I get your point now, will revert

internal/ent/hooks/listeners_onboarding_program.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil || len(labels) == 0 {
+		return err
+	}
+
+	caller, _ := auth.CallerFromContext(ctx)
+
+	newCaller := *caller
+	newCaller.OrganizationID = org.ID
+
+	ctx = auth.WithCaller(ctx, &newCaller)
```
golanglemonade: Isn't this already set earlier on the context? We shouldn't need to do it here as well? Keep the caller setup in one place.

internal/ent/hooks/listeners_onboarding_program.go
```diff
@@ -0,0 +1,197 @@
...
+
+	builder := client.Program.Create().
+		SetOwnerID(org.ID).
+		SetName(org.DisplayName + " Compliance Program").
```
golanglemonade: I'd recommend making a small function to determine the name based on:

if len(standards) == 1, return `standards[0].ShortName + Program <YYYY>`, e.g. `SOC 2 Program 2026`
if len(standards) > 1, return `Compliance Program <YYYY>` 

and then in the description include all the frameworks such as:

```
Track <frameworks> compliance activities, evidence, and audit readiness for YYYY.
```

I don't think we need the org name here because thats the org we are in; what framework it is and the year is more important.

internal/ent/hooks/onboarding.go
```diff
@@ -56,6 +56,11 @@ func HookOnboarding() ent.Hook {
 				}
 			}
 
+			compliance, _ := m.Compliance()
+			if _, _, err := resolveOnboardingStandards(ctx, m.Client(), compliance); err != nil {
+				return nil, err
+			}
```
golanglemonade: Is this still needed? Should be able to just put line 71 one back ?

golanglemonade (approved):

PR #2783, opened 2026-09-11
Drop deprecated category_type

adelowo:


internal/ent/hooks/platform.go
```diff
@@ -45,7 +46,12 @@ func HookPlatformFiles() ent.Hook {
 						}
 
 						m.AddArchitectureDiagramIDs(archFiles...)
-						if err := m.Client().File.Update().Where(file.IDIn(archFiles...)).SetCategoryType("architecture_diagram").Exec(allowCtx); err != nil {
+
+						err := m.Client().File.Update().
```
golanglemonade, line 50: have this set the category_name instead of category_type to keep the data here; check the UI side as well when we create these I think we allow them to be chosen?

golanglemonade:
Waiting on frontend removal

golanglemonade (approved):

PR #2802, opened 2026-09-15
Templated tasks should be viewable by other members

adelowo:


internal/ent/hooks/task.go
```diff
@@ -92,6 +94,75 @@ func HookTaskPermissions() ent.Hook {
 	)
 }
 
+// HookTaskTemplatePermissions manages the fga tuples for task templates
+func HookTaskTemplatePermissions() ent.Hook {
+	return hook.If(func(next ent.Mutator) ent.Mutator {
+		return hook.TaskFunc(func(ctx context.Context, m *generated.TaskMutation) (generated.Value, error) {
+
```
golanglemonade:
```suggestion
```

internal/ent/hooks/task.go
```diff
@@ -92,6 +94,75 @@ func HookTaskPermissions() ent.Hook {
...
+					return nil, err
+				}
+			}
+
+			v, err := next.Mutate(ctx, m)
+			if err != nil {
+				return nil, err
+			}
+
+			// if op.entcreate, this will be true
```
golanglemonade: what about ent.OpUpdateOne?

internal/ent/hooks/task.go
```diff
@@ -92,6 +94,75 @@ func HookTaskPermissions() ent.Hook {
...
+					ObjectType:  generated.TypeTask,
+					Relation:    "task_templates",
+				})
+
+				if isTemplate {
+					writeKeys = append(writeKeys, tuple)
+					return
+				}
+
+				deleteKeys = append(deleteKeys, tuple)
```
golanglemonade, line 151: we only need this if the old value was `isTemplate=true`?

internal/ent/hooks/task.go
adelowo: will be true for `ent.OpUpdateOne` too. should update the comment

internal/ent/hooks/task.go
adelowo: yes but fga client ignores missing tuples. I didn't want to have to do `OldXXX` and more db calls but I can add a note there

golanglemonade: but that means on any task update we are doing an update to fga that isn't required? this is okay on the bulk update but on update one, it should be cheap.

internal/ent/hooks/task.go
```diff
@@ -92,6 +94,77 @@ func HookTaskPermissions() ent.Hook {
...
+					return nil, err
+				}
+			}
+
+			v, err := next.Mutate(ctx, m)
+			if err != nil {
+				return nil, err
+			}
+
+			// if op.entcreate or ent.updateone, this will be true
```
golanglemonade, line 127:
```suggestion
			// if ent.OpCreate or ent.UpdateOne, this will be true
```

internal/ent/hooks/task.go
golanglemonade: oh, I see its only on `is_template` is on the mutation

golanglemonade (approved):

PR #2805, opened 2026-09-15
untangle preview zone target mismatch

matoszz:
- Preview domains now get their CNAME record zone and their CNAME target from two independent config values, so the record lands in theopenlane.net while pointing at the SaaS target in openlane-dns.com. The mismatch came from one value being used both as the record content and as the lookup key for the zone, which made it impossible to express the production topology
- The preview CNAME target is collapsed into the single trust center CNAME target. Preview and customer domains always point at the same SaaS host, and the separate knob had no valid value other than a copy of the main one. Its drift caused the incident / issues
- The mappable-domain backfill and its zone config are removed. With both targets sharing a name, that reconciler would have rewritten the SaaS mappable domain's zone to the preview zone on its next run and broken custom hostname creation for every vanity domain
- The stale preview domain backfill now also cleans up. It selects preview domains that Cloudflare blocked or that link to the wrong mappable domain, recreates the preview only when the trust center still points at the stale row, and schedules deletion of the stale CNAME record, custom domain, custom hostname, and verification in the zone they were written to


adelowo (approved):

golanglemonade (approved):

theopenlane-bender:

**📋 Draft PR**: https://github.com/theopenlane/openlane-infra/pull/1198

### Changes Preview:
✅ Updated ConfigMap template

- 🔄 Merged Helm values.yaml

PR #2807, opened 2026-09-15
Integration Record Provenance & Ownership

matoszz:
- Every ingested object now carries its full origin: the definition, the external system's instance id, the managing integration, and the run that last wrote it, so any row can be traced back to the exact source and sync that produced it
- Exactly one integration manages an object at a time: the first to claim it owns it, a second live integration of the same definition writes nothing and reports the row as skipped, and objects orphaned by a deleted integration are taken over by the next one that sees them
- Identity is scoped to the external system where it matters: directory accounts, groups, memberships, and check results become distinct objects per instance, while every other type stays one object per key per organization and is read-only to an integration on a different system or definition
- Ingest writes only on real change: every type is compared field by field, any data change lands as a single update with one event, unchanged objects are never touched, and run bookkeeping fields plus volatile mapped fields never count as a change on their own
- Freshness comes from runs instead of per-row timestamps: "last seen" is the finish time of the managing integration's latest successful run and "first seen" is the object's creation time, eliminating the last-seen write churn on every sync
- Mappings own data only: ownership, integration, and provenance fields are always set by ingest and overwrite anything a mapping produces, while directory name and primary source are read from the integration itself
- Concurrent and retried syncs cannot regress data: a run older than the one that last wrote an object is rejected for that object, and every retry attempt runs under its own run record linked to the attempt it retries
- Objects that fail to persist are tracked on integration health with their attempt count and last error, excluded from subsequent syncs for a bounded number of attempts, then dropped from tracking rather than blocking the sync
- Run records are the single source of sync outcomes: each run carries attempted, persisted, changed, skipped, failed, filtered, removed, and excluded counts plus a failure summary, and the separate directory sync run type is removed along with its per-row references
- Single-target links to instance-scoped objects resolve within the writer's external system, so a reference can no longer bind to a matching row that came from a different system, while to-many links still attach every match
- An integration's identity is the external system's stable identifier, resolved at install and refreshed by health checks, and re-authenticating with credentials for a different system is rejected instead of silently repointing the installation
- Directory sync processes in batches with reconcile loops purged and reseeded per operation, replacing the per-record fan-out and the duplicate loops it left behind
- Startup backfill runs once per release, migrates existing rows to the new provenance and instance-id model without re-claiming objects another integration manages, and signals completion so loop seeding runs against the migrated state
- Backfill work is pinned to the release that submitted it through a version-scoped queue, so old and new pods never process each other's backfill jobs during a rollout, and the retired mappable-domain backfill is removed


fga/model/automation/integrations.fga
```diff
@@ -54,8 +54,9 @@ type integration_run
     define editor: [service]
     define viewer: [user, service, group#member] or editor or parent_viewer
     define blocked: [user, service, group#member]
-    define audit_log_viewer: ([user, service, group#member] or audit_log_viewer from parent) and can_view
+    define audit_log_viewer: (([user, service, group#member] or audit_log_viewer from parent or audit_log_viewer from parent_context) and can_view) or can_view
 
     # parent permissions derived based on `crud` permissions or parents
-    define parent_viewer: can_view from parent
+    define parent_viewer: can_view from parent_context or can_view from parent
```
golanglemonade, line 60: Confirming that its intentional that now any org member can now view this, and not just those with access to the integration?

fga/model/automation/integrations.fga
```diff
@@ -54,8 +54,9 @@ type integration_run
...
+    define parent_context: [organization]
     define parent: [integration, assessment_response]
```
golanglemonade: Not related but saw it here - we can remove `assessment_response` now? since its not tied to an `integration_run`?

fga/model/automation/integrations.fga
matoszz: does it not inherit specifically from the parent since the parent is the integration?

internal/ent/hooks/asset.go
```diff
@@ -23,3 +27,64 @@ func HookAssetCreate() ent.Hook {
...
+				OnlyID(allowCtx)
+			if err != nil {
+				if generated.IsNotFound(err) {
+					return next.Mutate(ctx, m)
+				}
+
+				return nil, err
+			}
+
+			m.SetInternalOwnerUserID(userID)
```
golanglemonade: Should this fallback to checking for an identity holder record if the user isn't an openlane user?

internal/ent/schema/integrationrun.go
```diff
@@ -192,9 +162,9 @@ func (r IntegrationRun) Mixin() []ent.Mixin {
 		excludeAnnotations: true,
 		additionalMixins: []ent.Mixin{
 			newObjectOwnedMixin[generated.IntegrationRun](r,
-				withOrganizationOwnerServiceOnly(),
+				withOrganizationOwner(),
 				withSkipForSystemAdmin(),
-				withParents(Integration{}, AssessmentResponse{}),
```
golanglemonade, line 197: ah yes, this confirms my comment on the fga model, `assessment_response` should be removed there

fga/model/automation/integrations.fga
golanglemonade: Yes, I wrote this before I saw you changed this schema to `withOrganizationOwner` as well though.

golanglemonade (approved):
On a single GWSP integration, rerun 5 times on main:

```
  verb  | calls | rows | total_ms 
--------+-------+------+----------
 select |  2725 | 1832 |       43
 begin  |   405 |    0 |        0
 commit |   405 |    0 |        0
 update |   389 |  382 |       14
 insert |   118 |  118 |       23
(5 rows)
```

On this branch: 

```
  verb  | calls | rows | total_ms 
--------+-------+------+----------
 select |   543 | 1377 |       19
 insert |   210 |  210 |       26
 begin  |   165 |    0 |        0
 commit |   165 |    0 |        0
 update |    63 |   62 |        6
(5 rows)
```

-> 80% query reduction
-> 50% less inserts
-> 85% less update

Confirmed on integration disconnect + reconnect we maintain the same record instead of creating a duplicate

PR #2808, opened 2026-09-16
Migrate owner fields to use responsibility mixin

adelowo:
- added new responsibility fields to findings and vulnerabilities
- deprecated existing owner* fields in risks and added responsibility mixin
- added a backfill option to migrate existing values for both assets and risks
- updated risk hook to return early else we will need to add the current owner id to the ctx during the backfill

after migration, I tested backfill by checking the expected count for each schema that needed to be backfilled,
then ran it again afterwards to ensure it was back to 0

```
 SELECT 'risk' AS schema_name, COUNT(*) AS items
  FROM risks
  WHERE deleted_at IS NULL
    AND (
      (stakeholder_id <> '' AND stakeholder_group_id IS NULL)
      OR (delegate_id <> '' AND delegate_group_id IS NULL)
    )

  UNION ALL

  SELECT 'asset' AS schema_name, COUNT(*) AS items
  FROM assets
  WHERE deleted_at IS NULL
    AND internal_owner <> ''
    AND internal_owner_user_id IS NULL
    AND internal_owner_group_id IS NULL
    AND internal_owner_identity_holder_id IS NULL;

```


internal/ent/schema/program.go
```diff
@@ -175,6 +175,9 @@ func (Program) Fields() []ent.Field {
 			Optional().
 			Annotations(
```
golanglemonade, line 176: Why is this being deprecated? This wasn't on the issue and will always be a user in the system.

internal/ent/schema/remediation.go
```diff
@@ -99,7 +99,12 @@ func (Remediation) Fields() []ent.Field {
 			Optional(),
 		field.String("owner_reference").
 			Comment("reference to the owner responsible for remediation").
-			Optional(),
```
golanglemonade, line 102: Same thing here, this wasn't in the issue to migrate.

internal/ent/schema/platform.go
```diff
@@ -172,6 +172,9 @@ func (Platform) Fields() []ent.Field {
 			Optional().
 			Annotations(
 				entx.CSVRef().FromColumn("PlatformOwnerEmail").MatchOn("email"),
+				entgql.Directives(
```
golanglemonade: This wasn't on the list to migrate.

golanglemonade (approved):

PR #2817, opened 2026-09-18
Do not send notification for completed tasks and other updates except creation

adelowo:
- don't send notification to user when task is marked as completed. and only sent on creation/opening of the task.
- also don't send notification if other fields changes except assignee
- migrated the handler into a custom function because the default one is autogenerated by entityops. custom handler also much more simpler since we only care for assignee.


internal/ent/notifications/events.go
```diff
@@ -67,6 +68,13 @@ func Listeners() []gala.Registration {
 			Schema:  entityops.SchemaTask,
 			Fields:  []string{task.FieldAssigneeID},
 			Caller:  notificationCaller,
+			Match: []entityops.FieldMatch{
+				{
+					Field:  task.FieldStatus,
+					Negate: true,
+					In:     []string{string(enums.TaskStatusCompleted)},
```
golanglemonade: This should also not include `WontDo` - but the other thing is, it shouldn't send if the assignee isn't changed - which is currently what is happening today. I change something else and I get a notification a task is assigned to me.

